PR #52, opened 2017-02-10
document bootnode setup process

tylobban:


jpmsam (changes requested):
//this will generate a key and enode url inside a “tmp_file.txt” file
Please update the above comment. The command will start a bootnode with an enode URL. Only the key will be in the tmp_file which can be used to start the bootnode again with the same enode URL.

tylobban:
@jpmsam corrected, thanks.

tylobban:
@jpmsam with correct process now

jpmsam (approved):

PR #53, opened 2017-02-12
output vote block height instead of block height

bas-vk:
This PR contains 2 changes:

1. the `Vote` event emits the block height that was voted for instead of the block number. Each log already contains the block number and hash and therefore emitting the block number is redundant and thus removed.
2. the runtime code is generated with the latest stable solidity compiler (4.9)

Fixes: https://github.com/jpmorganchase/quorum/issues/26


jpmsam (commented):
Hi @bas-vk,

Can you please provide the steps to generate the bytecode from the solidity browser that would match the byte code in the PR ? Using 0.4.9+commit.364da425.Emscripten.clang (with optimizations enabled) isn't generating the same byte code. Based on our offline discussion, there might be a bug in the solidity browser or we might be missing a step.

Thanks,
Sam

bas-vk:
part of combined PR 92

PR #68, opened 2017-03-01
don't vote on CreateBlock request

### LINKED ISSUE #43: Inconsistent Voting behaviour
According to the Consensus section of the document, voter nodes will vote when they receive a newly generated block via the standard Ethereum P2P protocol (ChainHeadEvent). However they also appear to vote on the block generation event (CreateBlock).

E0120 21:44:49.886387 core/quorum/block_voting.go:288] Unable to create block: Node not configured for block creation
I0120 21:44:49.886915 core/quorum/block_voting.go:388] vote for 0xbb889fee70776551d79d7d65f6b5c3b1c932e15f63e3dba40b95b071e
495cab5 on height 8

As all nodes run the BlockMakerStrategy, looking in core/quorum/block_voting.go, in the run function for the CreateBlock event case it would seem that any voter node that is not enabled for block creation (i.e. createBlock will fail) will then always attempt to vote for the hash of the parent block of its current pending state.

### LINKED ISSUE #57: Voter nodes won't vote unless chain sync has occured
In the setup of a new chain voter nodes wont vote unless a chain sync occurs.

#### System information

Geth version: `geth version`
Geth
Version: 1.5.0-unstable
Git Commit: 85327f743b9735c32ae9231cbdde509e243ba0df
Protocol Versions: [63 62]
Network Id: 1
Go Version: go1.7.3
OS: linux
GOPATH=
GOROOT=/usr/local/go

Linux ubuntu-xenial 4.4.0-59-generic #80-Ubuntu SMP Fri Jan 6 17:47:47 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

q1.0.1

#### Expected behaviour
Voter node should vote for a block, when a ChainHeadEvent occurs in response to importing a new block provided they have synchronised with the chain (downloader.DoneEvent).

#### Actual behaviour
Voter nodes that are part of the initial chain never undertake a chain synchronisation, as by definition they are already in sync. They do still vote but only when a createBlock event occurs as a result of the fact they cannot create a block (See Issue #43) 

Code review block_voting.go 

```
case core.ChainHeadEvent: // got a new header, reset pending state
        bv.resetPendingState(e.Block)
        if bv.synced {
        number := new(big.Int)
                number.Add(e.Block.Number(), common.Big1)
                glog.Infof("ChainHeadEvent: Vote")
                if tx, err := bv.vote(number, e.Block.Hash()); err == nil {
                        if glog.V(logger.Debug) {
                                glog.Infof("Voted for %s on height %d in tx %s", e.Block.Hash().Hex(), number, tx.Hex())
                        }
                } else if glog.V(logger.Debug) {
                        glog.Errorf("Unable to vote: %v", err)
                }
        }

```

#### Backtrace

````
[backtrace]
````

### PR DESCRIPTION
**Pull request by bas-vk**
This PR addresses 3 issues:
- fixes #43 
- fixes #57
- deadlock when a vote/block request isn't handled within a 30s timeout

### DIFF
--- a/core/quorum/block_voting.go
+++ b/core/quorum/block_voting.go
@@ -2,13 +2,10 @@ package quorum
 
 import (
 	"crypto/ecdsa"
+	"fmt"
+	"gopkg.in/fatih/set.v0"
 	"math/big"
 	"sync"
-
-	"gopkg.in/fatih/set.v0"
-
-	"fmt"
-
 	"time"
 
 	"github.com/ethereum/go-ethereum/accounts"
@@ -42,14 +39,16 @@ const (
 // these transactions the parent block is selected where the next
 // block will be build on top of.
 type BlockVoting struct {
-	bc       *core.BlockChain
-	cc       *core.ChainConfig
-	txpool   *core.TxPool
-	synced   bool
-	mux      *event.TypeMux
-	db       ethdb.Database
-	am       *accounts.Manager
-	gasPrice *big.Int
+	bc               *core.BlockChain
+	cc               *core.ChainConfig
+	txpool           *core.TxPool
+	syncChain        bool
+	synced           bool
+	singleBlockMaker bool
+	mux              *event.TypeMux
+	db               ethdb.Database
+	am               *accounts.Manager
+	gasPrice         *big.Int
 
 	voteSession  *VotingContractSession
 	callContract *VotingContractCaller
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: core/quorum/block_voting.go [hunk 1]
```diff
@@ -223,14 +224,19 @@ func (bv *BlockVoting) run(strat BlockMakerStrategy) {
...
 				case downloader.DoneEvent, downloader.FailedEvent: // caught up, or got an error, start block createion and/or voting
+					bv.syncChain = false
 					bv.synced = true
 					strat.Resume()
 				case core.ChainHeadEvent: // got a new header, reset pending state
-					bv.resetPendingState(e.Block)
+					if !bv.syncChain {
+						bv.synced = true
+					}
 					if bv.synced {
```
jpmsam: @bas-vk in a multi-blockmaker configuration, synced will always be false since we aren't passing the singleblockmaker flag so the network will remain on block 0.

core/quorum/block_voting.go
```diff
@@ -244,21 +250,46 @@ func (bv *BlockVoting) run(strat BlockMakerStrategy) {
...
 					}
+
 				case CreateBlock:
+					if !bv.canCreateBlocks() {
+						if e.Err != nil {
+							e.Err <- fmt.Errorf("Node not configured/allowed to create blocks")
+						}
+						continue
+					}
+
```
jpmsam, line 285: @bas-vk in a multi-blockmaker configuration, synced will always be false since we aren't passing the singleblockmaker flag so the network will remain on block 0.

jpmsam (changes requested):
Please fix the multi blockmaker network setup.

bas-vk:
part of combined PR 92

PR #70, opened 2017-03-01
make block make request interval (pseudo) random

### LINKED ISSUE #61: Block Creation Strategy inconsistent with documentation
#### System information

Geth version: `geth version`
Geth
Version: 1.5.0-unstable
Git Commit: 85327f743b9735c32ae9231cbdde509e243ba0df
Protocol Versions: [63 62]
Network Id: 1
Go Version: go1.7.3
OS: linux
GOPATH=
GOROOT=/usr/local/go

Linux ubuntu-xenial 4.4.0-62-generic #83-Ubuntu SMP Wed Jan 18 14:10:15 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

q1.0.1

#### Expected behaviour

From the Wiki, 
> each Maker generates a random duration (a timeout) that it has to wait before it can create a block

> If a Maker reaches its timeout before any other Maker does it will create a block, after which it will generate a new random timeout for itself and wait for that to elapse before attempting to create another block.

> Note that once a Maker begins the block creation process, the other Maker Nodes will reset their current timeout, generate a new random timeout, and wait until that expires before attempting to creating a block.

#### Actual behaviour
There is no random seed to the random function, so the random number sequence is deterministic.  

The Maker timeout is actually reset twice - once when it starts creating a block and again when it inserts the newly created block onto its canonical chain.

The other Maker nodes timeout is only reset after a ChainHeadEvent, that is after the node has inserted a new block onto its canonical chain.  

N/A - Code Review

#### Backtrace

````
[backtrace]
````

bas-vk:
Fixes #61

### DIFF
--- a/core/quorum/vote_strategy.go
+++ b/core/quorum/vote_strategy.go
@@ -1,11 +1,13 @@
 package quorum
 
 import (
+	crand "crypto/rand"
+	"encoding/json"
+	"math"
+	"math/big"
 	"math/rand"
 	"sync"
 	"time"
-
-	"encoding/json"
 
 	"github.com/ethereum/go-ethereum/core"
 	"github.com/ethereum/go-ethereum/event"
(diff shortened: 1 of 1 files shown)


bas-vk:
part of combined PR 92

PR #92, opened 2017-04-11
Quorum strat update

### LINKED ISSUE #26: BlockVoting contract Vote event outputs current block.number not height voted for
In the vote function of the BlockVoting contract, the Vote event includes the block number when the vote was recorded.

Vote(msg.sender, block.number, hash);

Shouldn't the Vote event actually include the block height and hash voted for.

Vote(msg.sender, height, hash);

### LINKED ISSUE #44: BlockVoting contract getCanonHash function exception
The getCanonHash function does not guard against all input (height) scenarios . For example, when a blockmaker attempts to create its first block, it will call the function with a height of 1, this will throw an exception as no voting periods will have been completed. 

The function should check that enough voting periods have been recorded for the height being requested. 

There is also an edge-case for the input (height) zero.

bas-vk:
This PR includes:

- a bugfix for a race condition that could lead to a corrupt pending state
- change in the block maker/voting strategy that prevents deadlocks
- seed (pseudo) random deadline timers in voting strategy
- output the vote block height instead of the current block number (fixes #26)
- updates the runtime code for the vote contract to the latest solidity version (4.10) @jpmsam, this must also be changed in the quorum example genesis file).
- improves input validation for `getCanonHash`, fixes #44


eth/handler.go
```diff
@@ -112,7 +112,7 @@ func NewProtocolManager(config *core.ChainConfig, singleMiner bool, networkId in
 		quitSync:    make(chan struct{}),
 		raftMode:    raftMode,
 	}
-	if singleMiner {
+	if assumeSyncedInitially {
 		manager.synced = uint32(1)
```
ghost, line 116: There are problems propagating txs from a non-blockmaker node, even if the flag `--singleblockmaker` is on. The temporal solution is implementing the patch indicated below:

https://github.com/ethereum/go-ethereum/issues/2769#issuecomment-242141562

bas-vk: That should not be not be an issue for quorum. In quorum each blockmaker starts with the [assumption](https://github.com/jpmorganchase/quorum/pull/92/files#diff-5b7bdb45b66a5017bf89b46e6ef7a7f4R686) that it is already synced with the chain. If a node finds a node with a better chain it temporary stops accepting transactions until it has caught up.

This isn't an optimal solution and has corner cases that *might* explain why your transactions aren't propagated. Especially in small networks or networks with bad connectivity. It would help to determine what the problem is if you describe what your network setup looks like and what you are seeing.

jpmsam: @hermanjunge have you experienced the reported issue with this PR? I have multiple tests sending transactions across multiple nodes with different roles and I didn't see a TX propagation issue.

patrickmn:
Any outstanding concerns? @hermanjunge?

core/quorum/block_voting.go
```diff
@@ -223,42 +237,67 @@ func (bv *BlockVoting) run(strat BlockMakerStrategy) {
 
 				switch e := event.Data.(type) {
 				case downloader.StartEvent: // begin synchronising, stop block creation and/or voting
-					bv.synced = false
-					strat.Pause()
+					strat.PauseBlockMaking()
```
jpmsam, line 239: Please pause voting here too since it has been separated from block voting.

core/quorum/block_voting.sol
```diff
@@ -13,27 +13,27 @@ contract BlockVoting {
     // Raised when a new address is allowed to vote.
```
jpmsam, line 13: In the vote event, please change  the name of the param from "blockNumber" to "height" to match the name of the expected value.

core/quorum/vote_strategy.go
```diff
@@ -1,99 +1,156 @@
...
-		mux:    mux,
-		min:    int(min),
-		max:    int(max),
-		active: true,
+		mux:               mux,
+		minBlockTime:      int(minBlockTime),
+		maxBlockTime:      int(maxBlockTime),
+		minVoteTime:       int(minVoteTime),
+		maxVoteTime:       int(maxVoteTime),
+		blockCreateActive: true,
```
jpmsam: The `blockCreateActive` and `voteActive` flags are always set to `true`, can the default values be based on the role of the node? e.g: if the node is configured for voting, then `voteActive` will be `true`, otherwise `false`. I think the voting and block creation timers should only be activated based on the role of the node.

jpmsam (changes requested):
@bas-vk please review my feedback and update as needed.

core/quorum/vote_strategy.go
```diff
@@ -1,99 +1,168 @@
...
 // Start generating block create request events.
 func (s *randomDeadlineStrategy) Start() error {
 	if glog.V(logger.Debug) {
-		glog.Infof("Random deadline strategy configured with min=%d, max=%d", s.min, s.max)
+		glog.Infof("Random deadline strategy configured with minBlockTime=%d, maxBlockTime=%d, minVoteTime=%d, maxVoteTime=%d",
+			s.minBlockTime, s.maxBlockTime, s.minVoteTime, s.maxVoteTime)
 	}
-	s.deadlineTimer = time.NewTimer(time.Duration(s.min+rand.Intn(s.max-s.min)) * time.Second)
+
+	s.voteTimer = time.NewTimer(time.Duration(s.minBlockTime+rand.Intn(s.maxVoteTime-s.minVoteTime)) * time.Second)
```
jpmsam, line 130: @bas-vk based on my earlier suggestion, can you please create the timer for voting and block making only if the node is configured with votingActive or blockCreateActive? I don't see a reason why the timer should be active if the node can't vote or create blocks.

jpmsam (changes requested):
Hi @bas-vk, thanks for the latest commit with the fixes. I am testing the strategy update but please see my note regarding the timers.

jpmsam:
@bas-vk , closing this for now. Please submit a new one once we are done testing the latest updates.

PR #104, opened 2017-04-21
Raft updates

joelburget:


joelburget:
This fixes the stack traces in #96

patrickmn (dismissed):
Suggest either 1. Use just glog calls (like the general logging) and declare re-used messages as consts if necessary, 2. Rename logger to raftlog or something less ambiguous.

Also is DoLogRaft necessary?

joelburget:
@patrickmn the background here is that these messages are sentinels used by quorum-tools. We want them iff running in raft mode (or maybe more precisely iff quorum-tools is running quorum and listening for the sentinels). So,

* (1) It makes sense to hide this behind an abstraction since it's used five times.
* `DoLogRaft` is necessary because it allows us to tell the logger whether we're running in raft mode once, during startup.

Re (2): I think this makes sense as part of the `logger` module. I wanted it to be in `raft` but two of the checkpoints are in `internal/ethapi/api.go`.

patrickmn:
I definitely get the idea of having a "sum type" for the events, but this isn't really idiomatic Ethereum or Go. Anywhere else in the codebase they would simply be calls to glog, even if there's some redundancy in the format string.

The "logging" name adds confusion since it's not general-purpose. (Unfortunately the fact that the code is in logger/log_raft_checkpoint.go doesn't matter namespace-wise.) Might I suggest "raftlog" if you want to keep it? (i.e. raftlog.Checkpoint)

The DoLogRaft global is smelly, but it shouldn't be a problem for now. (TransactionManager/Private var is also a global.) Perhaps add a TODO comment to remove that global?

patrickmn (approved):

PR #124, opened 2017-06-01
additional 3rd party tools

tylobban:


conor10:
For Quorum-Docker, wouldn't it make more sense to simply include the Dockerfile in the 7 nodes example, rather than in a separate repo?

tylobban:
@conor10 yeah dockerizing the deployments is something we want so ultimately this makes sense. There are actually a few devs that have created docker images for Quorum (various levels of completeness and configurability) so we can link to them for now until we create one or someone wants to contribute theirs to Quorum.

expsam (approved):

jpmsam (approved):

PR #130, opened 2017-06-13
Dynamic raft membership support

bts:
This adds support for dynamic membership support (i.e. adding and removing nodes after initial cluster creation) for Raft-based consensus.

Summary of changes:

- Geth clients can now add new peers with `raft.addPeer(enodeId)` and remove existing peers with `raft.removePeer(raftId)`, where `raftId` is the number allocated/returned from `addPeer` (for newcomers to the cluster), or the 1-indexed position of an initial cluster member in the static peers list. Removal from the cluster is permanent, and if a party wishes to re-join the cluster after removal, they must be re-added. If a new node is added after the cluster has already started, they join with `--raftjoinexisting id`
- Added downloader support for a new `BoundedFullSync` mode, by which one node "fully" synchronizes with another, but only up to a specified block. Dynamically added peers, and peers that have fallen behind (e.g. after a crash), now use this functionality to catch their blockchain up with the rest of the cluster on start-up.
- All cluster information (i.e. `Address`es, to connect to other nodes, and "removed raft IDs") is now stored in raft snapshots (rather than LevelDB) so newcomers to the cluster can join without having to know all peer addresses a priori -- the rest of the cluster will connect to the newcomer.
- Added logic to recover from an etcd/raft corner case after a node crash (see coreos/etcd#7899)
- Updated documentation to reflect dynamic membership support
- Nodes can specify a custom raft port with `--raftport 12345`. If this option is omitted, we default to port 50400.

/cc @joelburget @jpmsam @patrickmn @tylobban

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -50,7 +50,6 @@ import (
 	"github.com/ethereum/go-ethereum/rpc"
 	whisper "github.com/ethereum/go-ethereum/whisper/whisperv2"
 	"gopkg.in/urfave/cli.v1"
-	"log"
 )
 
 func init() {
@@ -746,8 +745,11 @@ func RegisterEthService(ctx *cli.Context, stack *node.Node, extra []byte) {
 			if joinExistingId > 0 {
 				myId = uint16(joinExistingId)
 				joinExisting = true
+			} else if len(peers) == 0 {
+				Fatalf("Raft-based consensus requires either (1) an initial peers list (in static-nodes.json) including this enode hash (%v), or (2) the flag --raftjoinexisting RAFT_ID, where RAFT_ID has been issued by an existing cluster member calling `raft.addPeer(ENODE_ID)` with an enode ID containing this node's enode hash.", strId)
 			} else {
 				peerIds := make([]string, len(peers))
+
 				for peerIdx, peer := range peers {
 					peerId := peer.ID.String()
 					peerIds[peerIdx] = peerId
@@ -757,7 +759,7 @@ func RegisterEthService(ctx *cli.Context, stack *node.Node, extra []byte) {
 				}
 
 				if myId == 0 {
-					log.Panicf("failed to find local enode ID (%v) amongst peer IDs: %v", strId, peerIds)
+					Fatalf("failed to find local enode ID (%v) amongst peer IDs: %v", strId, peerIds)
 				}
 			}
 
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -380,7 +380,7 @@ var (
 		Name:  "raft",
 		Usage: "If enabled, uses Raft instead of Quorum Chain for consensus",
 	}
-	RaftBlockTime = cli.IntFlag{
+	RaftBlockTimeFlag = cli.IntFlag{
 		Name:  "raftblocktime",
 		Usage: "Amount of time between raft block creations in milliseconds",
 		Value: 50,
@@ -729,7 +729,7 @@ func RegisterEthService(ctx *cli.Context, stack *node.Node, extra []byte) {
 	}
 
 	if ctx.GlobalBool(RaftModeFlag.Name) {
-		blockTimeMillis := ctx.GlobalInt(RaftBlockTime.Name)
+		blockTimeMillis := ctx.GlobalInt(RaftBlockTimeFlag.Name)
 		datadir := ctx.GlobalString(DataDirFlag.Name)
 		joinExistingId := ctx.GlobalInt(RaftJoinExistingFlag.Name)
 
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -385,6 +385,11 @@ var (
 		Usage: "Amount of time between raft block creations in milliseconds",
 		Value: 50,
 	}
+	RaftJoinExistingFlag = cli.IntFlag{
+		Name:  "raftjoinexisting",
+		Usage: "The raft ID to assume when joining an pre-existing cluster",
+		Value: 0,
+	}
 )
 
 // MakeDataDir retrieves the currently requested data directory, terminating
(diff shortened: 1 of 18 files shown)


raft/constants.go
```diff
@@ -26,6 +26,8 @@ const (
 	snapshotPeriod = 250
 
 	peerUrlKeyPrefix = "peerUrl-"
+
+	chainExtensionMessage = "Successfully extended chain"
```
joelburget, line 30: This is used in just one place, no?

raft/peer.go
```diff
@@ -0,0 +1,80 @@
...
+		raftId:   raftId,
+		nodeId:   node.ID,
+		ip:       node.IP,
+		p2pPort:  node.TCP,
+		raftPort: raftPort(raftId),
+	}
+}
+
+// A peer that we're connected to via both raft's http transport, and ethereum p2p
+type Peer struct {
```
joelburget, line 34: I don't remember why we needed this new struct. Might be worth a note.

raft/peer.go
```diff
@@ -0,0 +1,80 @@
...
+	address *Address
+	p2pNode *discover.Node
+}
+
+func (addr *Address) EncodeRLP(w io.Writer) error {
+	return rlp.Encode(w, []interface{}{addr.raftId, addr.nodeId, addr.ip, addr.p2pPort, addr.raftPort})
+}
+
+func (addr *Address) DecodeRLP(s *rlp.Stream) error {
+	// These fields need to be public:
```
joelburget, line 44: most frustrating bug ever

raft/snapshot.go
```diff
@@ -1,66 +1,253 @@
...
-// See handling of EntryConfChange entries in raft/handler.go for details.
-func (pm *ProtocolManager) triggerSnapshotWithNextIndex(index uint64) {
-	pm.appliedIndex = index
-	pm.triggerSnapshot()
+func (pm *ProtocolManager) updateClusterMembership(newConfState raftpb.ConfState, addresses []Address, removedRaftIds []uint16) {
+	glog.V(logger.Info).Infof("updating cluster membership per raft snapshot")
+
+	prevConfState := pm.confState
+
+	// Update tombstones
```
joelburget: This would also be worth a note. The long discussion we had about how to remove peers / why we should tombstone has already mostly faded from memory.

joelburget (approved):
lgtm. a few minor notes but no blockers.

joelburget:
@patrickmn @jpmsam have you had the chance to take a look? any blockers?

raft/constants.go
bts: 2 places: `raft/handler.go` and `raft/snapshot.go`

raft/peer.go
bts: Done

bts:
I've tested this pretty thoroughly. We should merge it when we have a chance.

/cc @patrickmn

patrickmn (approved):

bts:
One thing to note about this branch is that it currently has new logic for the calculating the raft port. We now use `50400 + raftId` instead of `20000 + ethPort` -- it was easier/cleaner to do it this way. Now that some time has passed, I would have to dig back into the code to give more details surrounding this.

bts:
I just spoke with @patrickmn about this, and I'm going to look into adding another parameter to `addPeer` to specify the newcomer's raft port e.g. `raft.addPeer(enode, raftPort)`, and allow each node to specify their desired raft port on the command line.

bts:
Since this last commit, we now have `--raftport 12345` to allow the user specify the custom raft port `12345`. If `--raftport` is not provided, we default to port 50400.

Instead of `raft.addPeer(enode, raftPort)` as proposed above, we've settled on including a `raftport` querystring parameter in the enode ID (akin to `discport`) and continuing to use `raft.addPeer(enode)`. This is because we need to be able to specify raft ports for the initial peer set as well, and this approach is compatible with our use of `static-nodes.json` to describe the initial peer set.

bts:
@patrickmn and @joelburget, can you give this another quick look when you have a chance? Thanks.

patrickmn:
LGTM

cmd/utils/flags.go
```diff
@@ -389,6 +389,11 @@ var (
 		Usage: "The raft ID to assume when joining an pre-existing cluster",
 		Value: 0,
 	}
+	RaftPortFlag = cli.IntFlag{
```
joelburget: Maybe `UintFlag` would be better

cmd/utils/flags.go
```diff
@@ -731,6 +736,7 @@ func RegisterEthService(ctx *cli.Context, stack *node.Node, extra []byte) {
 		blockTimeMillis := ctx.GlobalInt(RaftBlockTimeFlag.Name)
 		datadir := ctx.GlobalString(DataDirFlag.Name)
 		joinExistingId := ctx.GlobalInt(RaftJoinExistingFlag.Name)
+		raftPort := uint16(ctx.GlobalInt(RaftPortFlag.Name))
```
joelburget: Similarly, this silently overflows if a value above `2^16` is provided. I probably wouldn't care except the default of `50400` is most of the way there.

cmd/utils/flags.go
```diff
@@ -751,6 +757,10 @@ func RegisterEthService(ctx *cli.Context, stack *node.Node, extra []byte) {
 				peerIds := make([]string, len(peers))
 
 				for peerIdx, peer := range peers {
+					if peer.RaftPort == 0 {
```
joelburget: An example might be worthwhile. I don't think I would know what to do if I saw this.

joelburget (commented):
A few nits but looks good otherwise

cmd/utils/flags.go
bts: I was thinking the same, but decided at the time to follow suit with other existing port flags.

antonmok:
Hello,
I am curious why you are using synchronization only up to block number stored in snapshot.

> Added downloader support for a new BoundedFullSync mode, by which one node "fully" synchronizes with another, but only up to a specified block.

What is difference between usual FullSync and BoundedFullSync mode if in both cases our aim is to sync with actual blockchain, when we starting new node or restarting it after crash?

bts:
We only want to sync up to an exact block, because in raft-based consensus
we want the state of the blockchain db to be entirely determined by the
state of the raft log (this includes what raft considers to have been
"applied" so far). Without BoundedFullSync, a node could have a newer DB
than the state of their raft log describes.

PR #135, opened 2017-06-28
Geth1.6

joelburget:
This PR still requires a bit more work, but I think the time is right for some feedback.

Design goals (in order of priority):

1. Correctness
  * I aim to stay fairly faithful to the original Quorum implementation. This diff started with naively applying Quorum changes to the Geth 1.6 codebase and only making changes to fix errors and reduce the diff size.
  * Unit tests have been incredibly useful. We should have more of them for Quorum.
  * Integration tests (quorum-tools) have similarly been incredibly useful.
  * My unscientific root cause analysis for errors that have cost me the most time so far:
    - Not-specific-enough types: It's the rule rather than the exception to have two or three `StateDB`s or `Address`es on hand when you need to know the right one to use. Would *you* know that it shouldn't be `st.to().Address()`, but rather `*msg.To()` when they're both valid `Address`es? And by the way `st.from().Address()` and `msg.From()` would also work. This is the kind of thing that typecheckers are made for but will take a human hours to figure out.
    - Getters that also set. Did you realize that the `.to()` in `st.to().Address()` in the last point modifies the state? Whoops.
    - Tricky logic. The transaction pool is complex. Execution is complex. There are way too many places we special case a gas price IMHO. Look for `params.IsQuorum` for some places we have special logic for Quorum.
    - `nil` pointers.
2. Small diff
  * Attempting to make it easy to pull in additional improvements in the future. We should be tracking `go-ethereum` master much more closely so we never have to do this again.
  * (TODO) Explore using the `Engine` interface for Raft rather than some of the special cases we currently have.

Most of the commits aren't interesting (and obviously only look at the last dozen or so that are by me). Two that are:

* https://github.com/jpmorganchase/quorum/commit/d996637fc73b191346cbe6d1879cae75d275dadd - add private state
* https://github.com/jpmorganchase/quorum/commit/a924cb629ca754af4f3f49927d15c8e52f41a9d0 -  fix some tricky errors

Misc:

* I use [EIP 155](https://github.com/ethereum/eips/issues/155) or [158](https://github.com/ethereum/eips/issues/158) depending on the block number (just like Geth)
* There are a ton of places where there's just an extra parameter or return value for the private state. We could maybe work around this by changing the type from `state.StateDB` to `state.QuorumStates` (which doesn't currently exist but would just hold but public and private state). I might play around with a couple ideas to see how well they work.
* This diff is meaningless. Don't look at it. Just look at individual commits by me. Don't try to read the commits comprehensively -- the changes are just too large. I added comments on a few of the commits where I'd like targeted feedback.


joelburget:
Other things I forgot to mention:

* I *do not* intend to merge this in. This PR is a place for conversation.
* Currently we pass all the privacy-related unit tests though there may be issues on other unit tests and quorum-tools integration tests.

gsalgado:
I just noticed that the code in core/call_helper.go is only used for tests (core/private_state_test.go), and ISTM privateTestMessage is needed only because they [pass the orig Transaction instance (ptx.Transaction) to NewEnv(), but ApplyMessage() gets passed ptx]( https://github.com/jpmorganchase/quorum/blob/856c9fe43e3eb675678c55ceda7eb4d298a7ca7d/core/call_helper.go#L62). Since here you always pass pmsg, maybe you can just call msg.SetPrivate() and get rid of privateTestMessage altogether?

gsalgado:
Actually, that test [doesn't seem to do what it's supposed](https://github.com/jpmorganchase/quorum/issues/143), and that's why it's currently passing on the master branch. In this branch it panics because it now correctly simulates private transactions

joelburget:
Significantly simplified, at least in terms of commits. This PR doesn't pass all the unit tests at the moment, though I think the failures are mostly benign. It does pass both the quorum-tools public and private state tests.

joelburget:
This diff is significantly easier to read: https://github.com/ethereum/go-ethereum/compare/master...joelburget:geth1.6

HACKING.md
```diff
@@ -0,0 +1,89 @@
+# Hacking on Quorum / various notes
+
+## Testing with Constellation
+
+### `tm.conf`
+
+Replace with appropriate absolute paths:
+
+TODO(joel): figure out how to use relative paths
```
patrickmn: Relative path should work just fine for Constellation config and PRIVATE_CONFIG env var. What's the issue?

accounts/keystore/keystore.go
```diff
@@ -0,0 +1,493 @@
...
+	// Look up the key to sign with and abort if it cannot be found
+	ks.mu.RLock()
+	defer ks.mu.RUnlock()
+
+	unlockedKey, found := ks.unlocked[a.Address]
+	if !found {
+		return nil, ErrLocked
+	}
+	// Depending on the presence of the chain ID, sign with EIP155 or homestead
+	if chainID != nil { // && !params.IsQuorum {
```
patrickmn: Add clarification of thinking?

consensus/ethash/consensus.go
```diff
@@ -0,0 +1,572 @@
...
+	errDuplicateUncle    = errors.New("duplicate uncle")
+	errUncleIsAncestor   = errors.New("uncle is ancestor")
+	errDanglingUncle     = errors.New("uncle's parent is not ancestor")
+	errNonceOutOfRange   = errors.New("nonce out of range")
+	errInvalidDifficulty = errors.New("non-positive difficulty")
+	errInvalidMixDigest  = errors.New("invalid mix digest")
+	errInvalidPoW        = errors.New("invalid proof-of-work")
+)
+
+func forceParseRfc3339(str string) time.Time {
```
patrickmn: Idiomatic naming is 'mustParseRfc3339'

core/block_validator.go
```diff
@@ -19,337 +19,129 @@ package core
...
-			// Error not unique
-			return UncleError("uncle[%d](%x) not unique", i, hash[:4])
-		}
-		uncles.Add(hash)
-
-		if ancestors[hash] != nil {
-			branch := fmt.Sprintf("  O - %x\n  |\n", block.Hash())
-			for h := range ancestors {
-				branch += fmt.Sprintf("  O - %x\n  |\n", h)
+	// TODO(joel)
```
patrickmn: What will happen with this?

core/blockchain.go
```diff
@@ -752,241 +816,278 @@ func (self *BlockChain) InsertReceiptChain(blockChain types.Blocks, receiptChain
...
-		default:
-			err = self.stateCache.Reset(chain[i-1].Root())
+		var parent *types.Block
+		if i == 0 {
+			parent = bc.GetBlock(block.ParentHash(), block.NumberU64()-1)
+		} else {
+			parent = chain[i-1]
 		}
+
+		// alias state.New because we introduce a variable named state on the next line
```
patrickmn, line 996: This shadowing smells. Maybe call state 'st'?

core/blockchain_test.go
```diff
@@ -965,34 +852,34 @@ func TestChainTxReorgs(t *testing.T) {
...
 			t.Errorf("share %d: expected tx to be found", i)
 		}
-		if GetReceipt(db, tx.Hash()) == nil {
+		if rcpt, _, _, _ := GetReceipt(db, tx.Hash()); rcpt == nil {
 			t.Errorf("share %d: expected receipt to be found", i)
 		}
 	}
 }
 
+/*
```
patrickmn: Clarify why disabled?

core/vm/evm.go
```diff
@@ -0,0 +1,426 @@
...
+	// Quorum additions:
+	publicState       PublicState
+	privateState      PrivateState
+	states            [1027]*state.StateDB // TODO(joel) we should be able to get away with 1024 or maybe 1025
+	currentStateDepth uint
+	readOnly          bool
+	readOnlyDepth     uint
+}
+
+// NewEVM returns a new EVM evmironment. The returned EVM is not thread safe
```
patrickmn: evm-ironment :)

patrickmn (commented):
Partial (mostly style) review

core/blockchain.go
joelburget: It totally smells. But I think it's more important to keep the diff small between Quorum and Geth. But I could change the name if others also want.

core/vm/evm.go
joelburget: I can't take credit for this :)

accounts/keystore/keystore.go
joelburget: I hadn't completely thought through the details. This change requires threading through a bit of state and tests were passing. I just created a [branch](https://github.com/joelburget/go-ethereum/commit/f924a27d67ab2eb4d0b2dede278fec1bd832de28) that handles this properly. I think it looks good. Thoughts?

consensus/ethash/consensus.go
joelburget: fixed

core/blockchain.go
patrickmn: Oh. I missed that. Nevermind.

core/blockchain_test.go
joelburget: Re-enabled

core/block_validator.go
joelburget: I'll re-enable this tomorrow morning.

bts:
FYI this branch is passing the [quorum-tools](https://github.com/jpmorganchase/quorum-tools) test suite as of [this commit](https://github.com/jpmorganchase/quorum-tools/commit/b95f99a78e092a0ad96fe67bd4274f44518e8ea6).

patrickmn:
awesome!

PR #137, opened 2017-07-09
fix build failure on mac

leo-xinwang:
After changing the path, the problem is gone.

### DIFF
--- a/build/ci.go
+++ b/build/ci.go
@@ -49,7 +49,7 @@ import (
 	"strings"
 	"time"
 
-	"../internal/build"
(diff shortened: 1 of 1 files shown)


patrickmn (approved):

patrickmn:
Even though this is a small change, I do have to ask you to complete the [CLA](https://github.com/jpmorganchase/cla/blob/master/cla-2016-09-01.pdf) and send it to quorum_info@jpmorgan.com.

patrickmn:
CLA received. Thank you!

PR #141, opened 2017-07-14
Quorum description update + new 3rd party tool

tylobban:
-better description of Quorum and enhancements over geth
-include blk.io's ERC20 REST service under 3rd party tools


conor10:
👍

patrickmn (approved):

PR #144, opened 2017-07-17
Fix TestPrivateTransaction to actually create private transactions

gsalgado:
It now creates private transactions, but that means it needs to run constellation in a subprocess, so do that as well

If this is something that could be accepted, we'd have to think if we want to change the build/ci scripts to install constellation to ensure constellation-node is available when we run 'make test', ~~or skip that test when running 'make test' and create a separate CI step to run it in travis~~


joelburget:
I've been starting constellation on the side before I run tests. This does seem a bit nicer.

patrickmn (approved):

Issue #112, opened 2017-05-05
Upgrade ethclient and abigen to support privateFor


stevenroose:
Currently the `privateFor` parameters is supported by the JSON-RPC API. The Go bindings in ethclient and abigen (that uses ethclient under the hood) do not yet.

Depends on: #90

stevenroose:
I'm looking into how hard it is to do.

stevenroose:
Back then, I realised it would be possible once #90 was solved, since abigen primarily relies on sendRawTransaction. Anyone else trying to solve this?

conor10:
I'm planning to take a look at https://github.com/jpmorganchase/quorum/issues/90, unless any of you guys are already @patrickmn? I saw the help wanted label was removed.

patrickmn:
@conor10 Please feel free. I was going to, but haven't started.

stevenroose:
There's two parts of the problem. One is the server, which allowing the sendRawTransaction JSON-RPC call to have a privateFor field somewhere. The other is incorporating that in the ethclient package. The latter should be more straightforward, but makes no sense if the former is not yet done. 

It's not easy to find how and where there were modifications to geth regarding private transactions. F.e. I can only see `private.P.Send` being called with the `data` field being contract code, not when making normal non-deployment transactions. 

Let me open a WIP PR with what I got.

patrickmn:
Regarding modifications, see https://github.com/jpmorganchase/quorum/pull/146#issuecomment-315831913

prd-fox:
This has been implemented in https://github.com/jpmorganchase/quorum/pull/819

Issue #90, opened 2017-04-09
Support privateFor in SendRawTransaction


patrickmn:
`SendRawTransaction` in the PublicTransactionPoolAPI (https://github.com/jpmorganchase/quorum/blob/master/internal/ethapi/api.go#L1250-L1272) should support the `privateFor` field as in `SendTransaction`: https://github.com/jpmorganchase/quorum/blob/master/internal/ethapi/api.go#L1210-L1246

stevenroose:
The `SendTransaction` method in the ethclient package also needs a way to specify privateFor.

subhodi:
@patrickmn Is there any update on this issue? 
looking for privateFor in sendrawTransaction.

stevenroose:
@subhodi There is a fundamental problem with supporting this that needs an architecture upgrade of Constellation to work.

A part of the realisation of this can be found in https://github.com/jpmorganchase/quorum/pull/146

The majority of the discussion about this was held in the Quorum Slack in the #dev channel.

Himanshu-Pandey:
Any ball park figure about when we can have this bug fixed?
Thanks

Krish1979:
PR #591 will fix this issue

PR #146, opened 2017-07-17
[WIP] Implement private tx support for ethclient and abigen

stevenroose:
It's really non-trivial to figure out where in the code the private parts of the transaction are handled.

Looking at `SendTransaction`, it does two things with the private stuff:
- call `private.P.Send()` with the tx data from the call, and afterwards use the newly returned data instead of the original one
- calling `SetPrivate()` on the transaction before handing it to `Backend.SendTx()`

The second one I can do, without a problem. The first one, though, is not easy. What I suppose I have to do is to send `tx.Data()` and then somehow replace that data with the new data. This would mean that I'd have to create a new `types.Transaction` object because you can't alter the private `txdata` field. I could in theory do that, but I guess that would invalidate the signature, no?

Involves #90 and #112.

### DIFF
--- a/accounts/abi/bind/backend.go
+++ b/accounts/abi/bind/backend.go
@@ -88,6 +88,10 @@ type ContractTransactor interface {
 	EstimateGas(ctx context.Context, call ethereum.CallMsg) (usedGas *big.Int, err error)
 	// SendTransaction injects the transaction into the pending pool for execution.
 	SendTransaction(ctx context.Context, tx *types.Transaction) error
+	// PreparePrivateTransaction sends the encoded raw transaction to Constellation,
+	// returning the encoded commitment transaction.
+	PreparePrivateTransaction(ctx context.Context, tx *types.Transaction,
+		privateFrom []byte, privateFor [][]byte) (*types.Transaction, error)
 }
 
 // ContractBackend defines the methods needed to work with contracts on a read-write basis.
--- a/accounts/abi/bind/base.go
+++ b/accounts/abi/bind/base.go
@@ -48,6 +48,9 @@ type TransactOpts struct {
 	Nonce  *big.Int       // Nonce to use for the transaction execution (nil = use pending state)
 	Signer SignerFn       // Method to use for signing the transaction (mandatory)
 
+	PrivateFrom []byte   // The public key of the Constellation identity to send this tx from.
+	PrivateFor  [][]byte // The public keys of the Constellation identities this tx is intended for.
+
 	Value    *big.Int // Funds to transfer along along the transaction (nil = 0 = no funds)
 	GasPrice *big.Int // Gas price to use for the transaction execution (nil = gas price oracle)
 	GasLimit *big.Int // Gas limit to set for the transaction execution (nil = estimate + 10%)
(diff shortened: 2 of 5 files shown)


patrickmn:
This is pretty much the full extent of the modifications (public-private state separation not included):

  - Send payload to Constellation, swap out data with identifier returned by it: https://github.com/jpmorganchase/quorum/blob/master/internal/ethapi/api.go#L280-L298

  - Identify private transaction, send identifier to Constellation, and swap out data with the decrypted payload returned by it: https://github.com/jpmorganchase/quorum/blob/master/core/state_transition.go#L225-L306

  - Private.P/PrivateTransactionManager: https://github.com/jpmorganchase/quorum/blob/master/private/private.go

  - Constellation implementation of PrivateTransactionManager: https://github.com/jpmorganchase/quorum/blob/master/private/constellation/constellation.go

The problem in this case, like you suggested, is that changing the data invalidates the signature. So it may be the "solution" is to require the user to do the data substitution and change of `v` prior to signing, and leaving SendRawTransaction pretty much as is.

stevenroose:
I left work, I'm back tomorrow.

Yeah I realized that.
Does this mean that we will need the client to communicate with Constellation directly? Or have Quorum expose some kind of API method that is a Constellation proxy (ala `preparePrivateTransaction`)? Or perhaps we can have the encryption of the payload occur in the client with possessing the Constellation key? That would break the constellation enclave concept, though..

stevenroose:
So, this is what seems like a reasonable approach; there are however some problems:

- Introduce a new API endpoint, ala `prepareRawPrivateTransaction`, that takes a raw transaction and the privateInfo data, just like the `sendRawTransaction` changes that I made yesterday. Instead of making the transaction, though, it only sends it to constellation and returns the modified transaction. To the caller.
- Leave sendRawTransaction as it is, with no private functionality. This call will then be used by the caller as a second step, which he will use to broadcast the transaction `prepareRawPrivateTransaction` returned after signing it.

There are two main issues:
- This will allow an RPC user to use Constellation directly. This means that he can have Constellation broadcast messages that will never actually be referenced by transaction. There is not a real incentive for the RPC user to do this, though. So this might be relatively safe.
- More problematically, because Constellation is a black box for the RPC user, it has to fully trust the Quorum node not to change its transaction data before sending it to Constellation. This is a big problem. A compromised Quorum node can send any arbitrary transaction data to Constellation and afterwards have the caller sign a commitment to that transaction. 

Without the second issue solved, I don't think this approach makes a lot of sense. One way to solve it, would be to have the client verify the returned transaction data. I don't know how Constellation does its magic, however. I assume the returned tx data primarily consists of a hash of something. If it's the hash of the data, that can easily be verified. If it's the hash of some encrypted data, this would require passing around public keys as well.

conor10:
So to avoid the second issue, the RPC client would need to go directly to Constellation to store and obtain the transaction hash for signing, using the [Send](https://github.com/jpmorganchase/constellation/blob/master/Constellation/Node/Api.hs#L41-L45) request that Geth currently uses (although its currently restricted to IPC connections only).

If RPC clients can only submit transactions to the store, but cannot modify/retrieve them themselves (i.e. they have to go via geth), the main risk becomes one of malicious clients sending payloads filling up Constellation's data store. Although a case could be made for removing all Geth to Constellation communication in this raw transaction scenario, as Geth is effectively being used only for logging secure transaction hashes on the blockchain  - everything else can be performed by the RPC clients performing the transaction signing directly with Constellation.

thellimist:
Any updates on this PR?

zzy96:
#633

fixanoid:
Closing. Abigen updates bundled in more recent releases.

PR #165, opened 2017-08-29
private: Distinguish between privateFor being absent vs. an empty array to allow self-sending

patrickmn:
Previously the absence of privateFor and `privateFor: []` meant the same thing: that the transaction is not private. Now, the latter will indicate that the transaction is private, but that it is applied only by the sender.

Fixes https://github.com/jpmorganchase/quorum/issues/81
Related Constellation modification: https://github.com/jpmorganchase/constellation/pull/49


joelburget (approved):
This is good 👍

Issue #139, opened 2017-07-11
Make Test Fail


chaozdeus:
#### System information

Geth version: `geth version`

Geth
Version: 1.5.0-unstable
Git Commit: 856c9fe43e3eb675678c55ceda7eb4d298a7ca7d
Protocol Versions: [63 62]
Network Id: 1
Go Version: go1.6.2
OS: linux
GOPATH=/root/go
GOROOT=/usr/lib/go-1.6

OS & Version: Ubuntu 16.04.2

Branch, Commit Hash or Release: `git status`

On branch master
Your branch is up-to-date with 'origin/master'.

#### Expected behaviour

Make Test should succeed

#### Actual behaviour

--- FAIL: TestConsoleWelcome (0.43s)
 run_test.go:259: (stderr) I0711 22:55:50.609093 ethdb/database.go:81] Allotted 128MB cache and 1024 file handles to /tmp/geth-test273396472/geth/chaindata
        run_test.go:259: (stderr) I0711 22:55:50.611859 ethdb/database.go:174] closed db:/tmp/geth-test273396472/geth/chaindata
        run_test.go:259: (stderr) I0711 22:55:50.612157 node/node.go:179] instance: Geth/v1.5.0-unstable/linux/go1.6.2
        run_test.go:259: (stderr) I0711 22:55:50.612175 ethdb/database.go:81] Allotted 128MB cache and 1024 file handles to /tmp/geth-test273396472/geth/chaindata
        run_test.go:259: (stderr) I0711 22:55:50.615649 eth/backend.go:177] Protocol Versions: [63 62], Network Id: 1
        run_test.go:259: (stderr) I0711 22:55:50.895486 eth/backend.go:195] WARNING: Wrote default ethereum genesis block
        run_test.go:259: (stderr) I0711 22:55:50.895820 core/blockchain.go:225] Last header: #0 [d4e56740…] TD=17179869184
        run_test.go:259: (stderr) I0711 22:55:50.895845 core/blockchain.go:226] Last block: #0 [d4e56740…] TD=17179869184
        run_test.go:259: (stderr) I0711 22:55:50.895852 core/blockchain.go:227] Fast block: #0 [d4e56740…] TD=17179869184
        run_test.go:259: (stderr) I0711 22:55:50.896290 p2p/server.go:328] Starting Server
        run_test.go:259: (stderr) I0711 22:55:50.896362 whisper/whisperv2/whisper.go:176] Whisper started
[...]

PR #179, opened 2017-09-13
Failing tests

bas-vk:
There are several unittests failing. This PR addresses 2 problems.

1. A problem with the banner. The raft submodule was added to the default IPC API list while it is only enabled with the `--raft` flag is specified when geth is started. This causes a difference between the list of enabled modules and the expected list of modules that is derived from this default IPC API.
2. `TestReorgLongBlocks` uses the `chm` method to create a blockchain for testing. This method didn't created the `chainEvents` channel causing it to be nil. Sending to a nil channel blocks in go, therefore this unittest fails after a timeout (default 10m).

Other failing tests:

`TestPrivateTransaction` depends on the constellation binary that I don't have installed locally.

`ExampleMakeCallHelper` fails because the test depends on the environment variable `PRIVATE_CONFIG` [here](https://github.com/jpmorganchase/quorum/blob/663710f4133bc0e80b9894a368999df0cf0339ac/private/private.go#L24). This environment variable doesn't exist and has a very generic name. Also if this test (didn't check) uses the constellation binary later on it will fail on that just like `TestPrivateTransaction`.

Go offers several methods to control which tests must be run. Since a lot of developers probably won't have the constellation binary locally installed and the environment configured would it make sense to disable these tests by default or rewrite them so they don't depend on constellation installed and configured? It is possible to only enable these test when a build tag is specified, e.g. something like `go test -tag=constellation`.

This solves #139 partially.


patrickmn:
LGTM.

Agreed re: Constellation. It should be possible to run this test suite even if it's not present. Would be happy to merge changes that make the tests simply emit a warning/not run when Constellation isn't present.

patrickmn (approved):

PR #183, opened 2017-09-15
Add the packaging metadata to build the quorum snap

come-maiz:
This package will let you publish the latest quorum in the Ubuntu store, and from there reach many users on all the supported Ubuntu versions, and more Linux distributions in progress. You just have to go to https://build.snapcraft.io and enable the automated continuous delivery.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,58 @@
+name: quorum
+version: master
+summary: A permissioned implementation of Ethereum supporting data privacy
+description: |
+  Quorum is an Ethereum-based distributed ledger protocol with
+  transaction/contract privacy and new consensus mechanisms.
+
+grade: devel # must be 'stable' to release into candidate/stable channels
+confinement: devmode # use 'strict' once you have the right plugs and slots
+
+apps:
+  abigen:
+    command: abigen
+  bootnode:
+    command: bootnode
+  constellation-node:
+    command: constellation-node
+  disasm:
+    command: disasm
+  ethtest:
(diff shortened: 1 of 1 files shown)


patrickmn:
This is great. Do you mind executing the CLA (https://github.com/jpmorganchase/cla) so we can merge?

snap/snapcraft.yaml
```diff
@@ -0,0 +1,58 @@
...
+      - curl
+      - libdb-dev
+      - libleveldb-dev
+      - libsodium-dev
+      - zlib1g-dev
+      - libtinfo-dev
+    build: |
+      curl -sSL https://get.haskellstack.org/ | sh
+      stack setup
+      stack install
```
patrickmn: This can be `stack build` since `install` copies the bins to `~/.local/bin`, but you're pulling them from the stack-work dir anyway.

come-maiz: Done

come-maiz:
Hello @patrickmn, thanks for the review!

Last night I sent my CLA signed and I've updated the build step. Please let me know if you have any questions, or need a hand. Whenever you are ready with the snap in edge, if yoou want we can start a call for testing with the ubuntu community, like this one: https://forum.snapcraft.io/t/call-for-testing-ipfs/97

pura vida.

come-maiz:
Ping, anybody around to review this?

fixanoid:
@elopio thanks for the snap addition. We may enable it at some point, but as it is we have a number of places to fetch the distro.

PR #192, opened 2017-10-06
Quorum-Geth1.6-stable with changes for PrivacyCore-1 and RDBMS connectors

kishorebhatia:
connector for RDBMS integration with Quorum: kafka connector for tx, blocks and statediff.
privacyCore-1: Hiding contract address.


stevenroose:
Could you perhaps provide a work of explanation on the technique used to achieve hidden contract addresses?

Considering this is a backwards-incompatible consensus-critical protocol change, it might benefit from some attention.

PR #196, opened 2017-10-17
added format in fatalf call.

rojotek:
Added a format directive to the fatalf call in raft/snapshot.bytesToSnapshot()

Reformatted import statements (from go fmt).


rojotek:
Cherry-picked commit and applied against latest 1.6 in #197

PR #207, opened 2017-10-31
Geth 1.7.2 rebase and addition of Istanbul BFT

patrickmn:
Some features still missing, notably QuorumChain and the Porosity integration.


stevenroose:
Grrr, trying to look into some Istanbul issues. Wonder why 1.7.2 rebase and Istanbul were put in the same commit...

PR #211, opened 2017-11-02
eth/downloader: Updated downloader tests to improve perf and reliability

rojotek:
eth/downloader: Updated downloader tests to improve perf and reliability
Updated use of Parallel and added some subtests to help isolate them.
Increased timeout in floodingTestPeer.RequestHeadersByNumber, so that it
doesn't accidently timeout and cause other tests to break.
Extra sleep for the first failure in TestFastCriticalRestarts.
(Cherry-picked commit from PR against ethereum/go-ethereum).


patrickmn:
Can you link the upstream PR please?

PR #230, opened 2017-11-15
Re-add GPO creation in eth backend

bts:
When trying to get Clique consensus working in Quorum, `EthApiBackend.SuggestPrice` in `api_backend` was panicking because we hadn't instantiated a gas price oracle in the `eth` `backend`.


bts:
FWIW, the only place in the codebase where the GPO is used is in `SuggestPrice`. This should have no effect on Raft consensus.

patrickmn (approved):

joelburget (approved):
🥇

PR #233, opened 2017-11-20
Add an memory based Private Transaction Manager

bretthenderson:
Stubbed out Constellation by creating a simple in memory based Private Transaction Manager.


trung:
Many thanks for the effort.

I don't we need this for unit testing as we can easily create mocks from `PrivateTransactionManager` interface or stubs. Therefore we don't need an actual implementation.

We would be looking at adding it as an external service instead.

PR #271, opened 2018-02-01
Fix sometimes printing wrong address.

joelburget:
Prior to this fix, a series of submitted transactions in the javascript
console would each show a different "from" field when inspected with
`eth.getTransaction`.

It turns out the transactions were all created with the correct sender,
but were just printed incorrectly. We were using the EIP155Signer when
all Quorum transactions expect the HomesteadSigner.

This fix is not exactly satisfying, since in every other place we've
been able to use an `isQuorum` boolean to tell whether a `v` of `37` or
`38` denotes a quorum private tx, but it's particularly hard to do that
in these two places.

### DIFF
--- a/core/types/transaction.go
+++ b/core/types/transaction.go
@@ -39,6 +39,7 @@ var (
 
 // deriveSigner makes a *best* guess about which signer to use.
 func deriveSigner(V *big.Int) Signer {
+	// joel: this is one of the two places we used a wrong signer to print txes
 	if V.Sign() != 0 && isProtectedV(V) {
 		return NewEIP155Signer(deriveChainId(V))
 	} else {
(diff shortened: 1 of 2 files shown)


patrickmn (approved):

patrickmn:
> This fix is not exactly satisfying, since in every other place we've
been able to use an isQuorum boolean to tell whether a v of 37 or
38 denotes a quorum private tx, but it's particularly hard to do that
in these two places.

And it's worth noting that 37 and 38 are now reserved for other purposes upstream: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md

joelburget:
Yes -- I'd like to revisit the 37/38 decision

Issue #256, opened 2018-01-10
Events emitted in Private Transactions are not present in the transaction receipt logs


macro88:
#### System information

Geth version: ` 1.7.2-stable-0905eda4`  (`Geth/v1.7.2-stable-0905eda4/linux-amd64/go1.9.2`)

OS & Version: Linux `Ubuntu 16.04.3 LTS`

Branch, Commit Hash or Release: ` HEAD detached at v2.0.0 nothing to commit, working directory clean`

#### Expected behaviour
Events logged/emitted from a Private transaction are retrieved when recalling the transaction receipt. E.g. `eth.getTransactionReceipt(tHash)`
In relation to the test contract below the expected behaviour is
* Event is present in the logs when recalling the transaction receipt
* Callback function called when watching the event
* `invocation` is incremented.

#### Actual behaviour
The logs are empty when recalling the transaction receipt for a Private Contract and Private transaction. The transaction succeeds and state is updated in the contract.
In relation to the test contract below the expected behaviour is:
* [Fail] Event is present in the logs when recalling the transaction receipt
* [Pass] Callback function called when watching the event
* [Pass] `invocation` is incremented.

#### Summary of testing performed
A test contract `EventEmitter.sol` was created. This contract is a simple contract that emits and event and increments a counter.
```
contract EventEmitter {
    event ContractEvent(string payload); 
    uint public invocations; 

    function EventEmitter() { 
    }

    function RaiseEvent(string payload) {
        ContractEvent(payload);
        invocations++;
    } 
}
```
Tested under `Raft` and `Istanbul` consensus and for each `solc` version 0.3.6, 0.4.8, 0.4.18 (the behaviour below was consistent)
1. Deploy the contract without `privateFor` and send transaction without `privateFor` - this works as expected. 
* [Pass] Event is present in the logs when recalling the transaction receipt 
* [Pass] Callback was called when watching the event in geth
* [Pass] `invocation` was incremented.
2. Deploy the contract with `privateFor` and send transaction without `privateFor` - Failed. Occasionally this killed the geth/constellation process of the sender and recipient – this was random and I can’t reliably repro it.
* [Fail] Event is present in the logs when recalling the transaction receipt
* [Fail] Callback was not called when watching the event in geth
* [Fail] `invocation` was not incremented.
3. Deploy the contract with `privateFor` and send transaction with `privateFor` - Partial success.
* [Fail] Event is present in the logs when recalling the transaction receipt
[...]

PR #282, opened 2018-02-12
Receipt fix 1

joelburget:
Write public and private block receipts separately, by default retrieve public receipts from the eth api.


joelburget:
This is the less hacky fix, but slightly more complicated and with a larger diff from upstream.

core/database_util.go
```diff
@@ -239,8 +239,15 @@ func GetBlock(db DatabaseReader, hash common.Hash, number uint64) *types.Block {
 
 // GetBlockReceipts retrieves the receipts generated by the transactions included
 // in a block given by its hash.
-func GetBlockReceipts(db DatabaseReader, hash common.Hash, number uint64) types.Receipts {
-	data, _ := db.Get(append(append(blockReceiptsPrefix, encodeBlockNumber(number)...), hash[:]...))
+func GetBlockReceipts(db DatabaseReader, hash common.Hash, number uint64, isPrivate bool) types.Receipts {
+	var prefix []byte
+	if isPrivate {
+		prefix = privateblockReceiptsPrefix
```
joelburget, line 245: `privateblockReceiptsPrefix` previously existed but was unused.

joelburget:
fixes #256

PR #283, opened 2018-02-12
Receipt fix 2

joelburget:
Overwrite public block receipts with private receipts if they exist.


joelburget:
fixes #256

patrickmn:
If there isn't a reason one would need to retrieve public receipts for private transactions, and given that this diff is so much smaller, this seems like the best choice compared to https://github.com/jpmorganchase/quorum/pull/282.

joelburget:
You're right. Let's go with this one.

patrickmn (approved):

PR #288, opened 2018-02-16
Re-enable gas price.

joelburget:
This change removes all the places we had intentionally disabled non-0
gas price.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -28,7 +28,6 @@ import (
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
-	"github.com/ethereum/go-ethereum/common/math"
 	"github.com/ethereum/go-ethereum/common/mclock"
 	"github.com/ethereum/go-ethereum/consensus"
 	"github.com/ethereum/go-ethereum/core/state"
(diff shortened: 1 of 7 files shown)


joelburget:
I want to do a bit more testing on this, but I wanted to make sure to submit the PR today.

patrickmn:
Suggestion (not necessarily for this PR, but it would immediately be needed): Private transactions should have a non-zero cost when gas price is non-zero, and that cost should probably be closer to something like ecrecover than 0. What do you think?

kegsay:
Will this mean that clients will need to manage ether (e.g. via a faucet)? If so, what's the rationale behind this change?

patrickmn:
Only if you want to, i.e. have set gas price to something greater than zero in the genesis block.

The rationale is to reinstitute a mechanism for ratelimiting that can be used when Quorum is run in environments where it's *not* okay to unconditionally allow anyone to send any number of transactions.

aroommen:
Could we have this available in Quorum ASAP? This would be a great feature to have for network metering

fixanoid:
@aroommen this is a relatively low priority for us, but we'd welcome your testing and validation of this pull as we are planning to move into master at some point in the future.

jpmsam:
This will be reworked based on latest Quorum release.

PR #297, opened 2018-02-27
Consensus failed

### LINKED ISSUE #289: Istanbul consensus failure: inconsistent environment variables
This issue continues #285.

So, I managed to find a small transaction that triggers Istanbul validators to create wrong blocks (i.e. one that the other validators don't accept).

Situation:
- observed both with 3 and 7 nodes
- **not** observed with `--istanbul.blockperiod 1 --istanbul.blockpausetime 2` (default)
- observed with `--istanbul.blockperiod 7 --istanbul.blockpausetime 8`
- every validator creates invalid blocks rejected by the others, so the block created process halts entirely
- no issues when using Raft

The interesting thing about the triggering tx is that it combines all environment variables into a hash and emits it in an event. This means that somehow the validators don't agree on the environment set for tx execution for block validation. This might be because the execution environment the validator has for building the block is wrong.

```solidity
pragma solidity ^0.4.6;

contract Test {
    event TestEvent(bytes32 hash);
    function test() public payable {
        var h = sha256(this);
        h = sha256(h, block.blockhash(block.number-1));
        h = sha256(h, block.coinbase);
        h = sha256(h, block.difficulty);
        h = sha256(h, block.gaslimit);
        h = sha256(h, block.number);
        h = sha256(h, block.timestamp);
        h = sha256(h, msg.data);
        h = sha256(h, msg.gas);
        h = sha256(h, msg.sender);
        h = sha256(h, msg.sig);
        h = sha256(h, msg.value);
        h = sha256(h, now);
        h = sha256(h, tx.gasprice);
        h = sha256(h, tx.origin);
        TestEvent(h);
    }
}
```

I build the contract with solc and generated Go bindings with abigen. This is the code (including generated code) that I used to deploy and execute the transaction:
https://gist.github.com/stevenroose/a722f360e11d23ca6032a0fa5a8477b0

markya0616:
In this PR, we fixed https://github.com/jpmorganchase/quorum/issues/289.
The root cause is we change the block header timestamp in Seal(). It will cause different root hash in proposer and validators' side.

We did the following updates:
1. remove `BlockPauseTime`
2. set block header timestamp in Prepare()

### DIFF
--- a/cmd/geth/main.go
+++ b/cmd/geth/main.go
@@ -122,7 +122,6 @@ var (
 		utils.EmitCheckpointsFlag,
 		utils.IstanbulRequestTimeoutFlag,
 		utils.IstanbulBlockPeriodFlag,
-		utils.IstanbulBlockPauseTimeFlag,
 	}
 
 	rpcFlags = []cli.Flag{
--- a/cmd/geth/usage.go
+++ b/cmd/geth/usage.go
@@ -22,6 +22,8 @@ import (
 	"io"
 	"sort"
 
+	"strings"
+
 	"github.com/ethereum/go-ethereum/cmd/utils"
 	"github.com/ethereum/go-ethereum/internal/debug"
 	"gopkg.in/urfave/cli.v1"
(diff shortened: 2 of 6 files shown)


yutelin:
@patrickmn @tylobban Can you help take a look? Thanks!

patrickmn (approved):

PR #350, opened 2018-04-23
Use EIP155 signer when needed

tsuzukit:
Fix to https://github.com/jpmorganchase/quorum/issues/344.
Tested with `Web3 1.0.0-beta.34` and `networkid` and `chainid` of `10`.

Private transaction uses `HomesteadSigner` as before, but locally signed transaction with Web3 should use `EIP155Signer` depending on the block number.

`networkid` and `chainid` must not be 1, because it will produce `V` value of `37` which is used to indicate transaction being private with Quorum. (So I think 7nodes example should be fixed too if this is correct fix to the issue)

### DIFF
--- a/core/types/transaction_signing.go
+++ b/core/types/transaction_signing.go
@@ -40,9 +40,6 @@ type sigCache struct {
 
 // MakeSigner returns a Signer based on the given chain config and block number.
 func MakeSigner(config *params.ChainConfig, blockNumber *big.Int) Signer {
-	if config.IsQuorum {
-		return HomesteadSigner{}
-	}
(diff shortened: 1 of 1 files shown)


jpmsam (approved):

peterbroadhurst:
Thanks @jpmsam - we were just looking at this issue, as we came across it when testing MetaMask with Kaleido (per https://kaleido.io/kaleido-meets-metamask/ ). We will look to pull in this fix.

jpmsam:
@peterbroadhurst thanks to @tsuzukit for his work and contribution.

PR #370, opened 2018-05-09
Quorum 2 rc

guojian1234:
Can private transactions withstand double attack?


jpmsam:
Please use slack for questions.

PR #354, opened 2018-04-24
Fix TestUpdatedKeyfileContents, TestEIP155SigningVitalik and TestChainId

tsuzukit:
For fixing `TestUpdatedKeyfileContents`, I have cherry-picked https://github.com/jpmorganchase/quorum/commit/c86c12696db61511e7983d922505f59e7ddac076 and https://github.com/jpmorganchase/quorum/commit/c8b5da70df05f7ce7f05180f87f8429a2cff4d23 which are already merged to geth 1.7.3.

For fixing `TestEIP155SigningVitalik`, I needed to change `chainId` to something other than `1` as it results in `V` of `37` and Tx is treated as protected transaction in Quorum (https://github.com/jpmorganchase/quorum/blob/master/core/types/transaction.go#L138).
Since http://vitalik.ca/files/eip155_testvec.txt is not currently accessible, I just created some tx and address pair. (tx creation rules are written in comments)

For same reason, I needed to change `chainId` for `TestChainId`.

Please review.

PR #375, opened 2018-05-15
Use eip155 signer for public transactions

zekun-shi:
Use eip155 for public transactions. Private transactions will continue to use V={37,38}, hence you need to set chain id to some value other than 1 to avoid conflict. This PR should be merged along with #350 (which enable eip155 for locally signed transactions), and #354 (which fixes `TestEIP155SigningVitalik` by using a new set of test vectors with `chain_id=10`)

Tested on public value transfer, private contract deployment, public contract deployment on 'Ubuntu 16.04.4 LTS` with `Web3 1.0.0-beta.34` and `chain_id=10`.

### DIFF
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -1169,10 +1169,9 @@ func (s *PublicTransactionPoolAPI) SendTransaction(ctx context.Context, args Sen
 	tx := args.toTransaction()
 
 	var chainID *big.Int
-	isQuorum := false
+	isQuorum := tx.IsPrivate()
 	if config := s.b.ChainConfig(); config.IsEIP155(s.b.CurrentBlock().Number()) && !isPrivate {
 		chainID = config.ChainId
-		isQuorum = true
 	}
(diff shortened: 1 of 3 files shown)


internal/ethapi/api.go
```diff
@@ -1052,7 +1053,7 @@ func (s *PublicTransactionPoolAPI) sign(addr common.Address, tx *types.Transacti
 	// Request the wallet to sign the transaction
 	var chainID *big.Int
 	isQuorum := false
-	if config := s.b.ChainConfig(); config.IsEIP155(s.b.CurrentBlock().Number()) {
+	if config := s.b.ChainConfig(); config.IsEIP155(s.b.CurrentBlock().Number()) && !tx.IsPrivate() {
 		chainID = config.ChainId
 		isQuorum = true
```
jpmsam: @Szkered isQuorum will stay false here for private(Quorum) transactions.

jpmsam (approved):

PR #378, opened 2018-05-15
Use permissioned-nodes.json for discovery

### LINKED ISSUE #287: Use the permissioned nodes for discovery
The usage of a permissioned network with `--permissioned` and the `permissioned-nodes.json` file, should make it unnecessary to have bootnodes. Nodes should use this list as the initial source of new peers instead of bootnodes.

jmcnevin:
If discovery is enabled, use the permissioned nodes list to pre-populate the node database.

Discovery needs to be enabled for this to work, but I think, in theory, you could disable the UDP port at the firewall and still get the benefits of this.  Ideally, it would be nice to be able to decouple the mechanism for finding new nodes to talk to from the UDP gossip implementations.

Fixes #287

### DIFF
--- a/cmd/bootnode/main.go
+++ b/cmd/bootnode/main.go
@@ -96,12 +96,14 @@ func main() {
 		}
 	}
 
+	var knownNodes []*discover.Node
+
 	if *runv5 {
 		if _, err := discv5.ListenUDP(nodeKey, *listenAddr, natm, "", restrictList); err != nil {
 			utils.Fatalf("%v", err)
 		}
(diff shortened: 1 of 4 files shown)


stevenroose:
> @jpmsam self-requested a review 21 days ago 

You're making your procrastination public :D

jpmsam:
@stevenroose the PRs that solve open bugs and fixing issues are my priority. Have you tested the update and does it implement your enhancement request?

As for the backlog, we are hiring.

PR #387, opened 2018-05-29
Feature/fix_some_tests

tsuzukit:
Fixed test with error `insufficient balance for transfer`.
It happened because chain id was set to be 1 in config.go.
Since `v` becomes `38` and tx is regarded as private, recovered address is not the one specified in genesis.

Cherry-picked below commits because `make test` did not run with Go 1.10.
`go vet` failed for minor issues.

- https://github.com/ethereum/go-ethereum/pull/15783/commits/eaf59f5e5f0b47a363e1971a798886a781c935b7
- https://github.com/jpmorganchase/quorum/commit/bae67281d67b3e1f57cff993a32e325ea45ccfa9

### DIFF
--- a/cmd/geth/misccmd.go
+++ b/cmd/geth/misccmd.go
@@ -134,7 +134,6 @@ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
 GNU General Public License for more details.
 
 You should have received a copy of the GNU General Public License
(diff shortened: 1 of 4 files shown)


jpmsam (approved):

tsuzukit:
@jpmsam thank you very much for approval! 
Is this going to be merged? I would also want to fix other tests currently broken if I can find a way to fix it.

PR #399, opened 2018-06-07
validate input before calling evm

jpmsam:
This fixes the issue with a node crashing when it is a participant in the private smart contract creation but it is not part of transactions updating the smart contract.


fixanoid (approved):
Tested in Raft mode. Node no longer crashes.

fixanoid:
Fixes #404

PR #405, opened 2018-06-09
Allow to use http client

atoulme:
This PR changes the way Quorum establishes connections to Constellation.

By indicating that the socket configuration parameter is empty or absent, the code uses a HTTP client instead of the Unix domain socket client.

The HTTP client is configured with higher threshold for maximum concurrent connections towards the same host to allow more than the default 2 connections against the same host.

The PR also ensures all responses are fully consumed when sent back so that HTTP connections can be reused. See https://stackoverflow.com/questions/17948827/reusing-http-connections-in-golang#17953506

### DIFF
--- a/private/constellation/constellation.go
+++ b/private/constellation/constellation.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"github.com/patrickmn/go-cache"
 	"os"
-	"path/filepath"
 	"time"
 )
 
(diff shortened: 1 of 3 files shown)


private/constellation/constellation.go
```diff
@@ -47,19 +46,23 @@ func New(path string) (*Constellation, error) {
 	}
 	// We accept either the socket or a configuration file that points to
 	// a socket.
-	isSocket := info.Mode() & os.ModeSocket != 0
+	isSocket := info.Mode()&os.ModeSocket != 0
```
Puneetha17: Please do ensure to have the spaces in place `isSocket := info.Mode() & os.ModeSocket != 0`
@conor10

conor10: Thanks @Puneetha17 - @atoulme that's the only feedback from us, otherwise looks good 👍 (build appears to be failing however...)

atoulme: Thanks folks. I am going to take a look at that failing build.

PR #426, opened 2018-06-29
expose leader enode id to console

trung:
* capture leader raft id from `Ready.SoftState` inside `eventLoop()`
* exposed API to return leader enode id 

PS: plz don't hesitate to criticize my changes.

### DIFF
--- a/raft/handler.go
+++ b/raft/handler.go
@@ -1,6 +1,7 @@
 package raft
 
 import (
+	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -894,13 +895,16 @@ func (pm *ProtocolManager) updateLeader(leader uint64) {
 	pm.leader = uint16(leader)
 }
 
-func (pm *ProtocolManager) LeaderAddress() *Address {
+// The Address for the current leader, or an error if no leader is elected.
+func (pm *ProtocolManager) LeaderAddress() (*Address, error) {
 	pm.mu.RLock()
 	defer pm.mu.RUnlock()
 
-	leaderAddress := pm.address
-	if l, ok := pm.peers[pm.leader]; ok {
-		leaderAddress = l.address
+	if minterRole == pm.role {
+		return pm.address, nil
+	} else if l, ok := pm.peers[pm.leader]; ok {
+		return l.address, nil
 	}
-	return leaderAddress
+	// We expect to reach this if pm.leader is 0, which is how etcd denotes the lack of a leader.
+	return nil, errors.New("no leader is currently elected")
 }
--- a/raft/handler.go
+++ b/raft/handler.go
@@ -50,6 +50,7 @@ type ProtocolManager struct {
 	snapshotIndex uint64 // The index of the latest snapshot.
 
 	// Remote peer state (protected by mu vs concurrent access via JS)
+	leader       uint16
 	peers        map[uint16]*Peer
 	removedPeers *set.Set // *Permanently removed* peers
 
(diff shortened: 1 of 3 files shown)


trung:
hmm, tests are failing ...

nadeemajl (commented):
The key change in the PR is in raft/handler.go where  the eventLoop is blocked until it gets a ready object from the raft ready channel.
As per the  usage doc of raft, ready state is required to follow with a advance, the control flow will eventually call advance() .

Overall this is a useful commit.
I was able to verify the changes by incorporating it in my local environment.

nadeemajl:
The commit is functionally valid, below are the logs with the changes incorporated in local environment, also note that value of the nodeid changes when leader/minter role changes.

> raft
{
  leader: "ac6b1096ca56b9f6d004b779ae3728bf83f8e22453404cc3cef16a3d9b96608bc67c4b30db88e0a5a6c6390213f7acbe1153ff6d23ce57380104288ae19373ef",
  role: "minter",
  addPeer: function(),
  getLeader: function(callback),
  getRole: function(callback),
  removePeer: function()
}
> raft
{
  leader: "eacaa74c4b0e7a9e12d2fe5fee6595eda841d6d992c35dbbcc50fcee4aa86dfbbdeff7dc7e72c2305d5a62257f82737a8cffc80474c15c611c037f52db1a3a7b",
  role: "verifier",
  addPeer: function(),
  getLeader: function(callback),
  getRole: function(callback),
  removePeer: function()
}

bts:
With these changes, I think it's now good to go.

raft/handler.go
```diff
@@ -881,3 +886,21 @@ func (pm *ProtocolManager) advanceAppliedIndex(index uint64) {
...
+	defer pm.mu.Unlock()
+
+	pm.leader = uint16(leader)
+}
+
+func (pm *ProtocolManager) LeaderAddress() *Address {
+	pm.mu.RLock()
+	defer pm.mu.RUnlock()
+
+	leaderAddress := pm.address
```
jpmsam: ` leaderAddress := pm.address`
@trung @bts Is it possible because of this line for the local address to be reported as the leader ?

trung: as the `pm.peers` doesn't contain the current node so that above statement makes an assumption the leader is the current node being evaluated.

Then the following `if` condition would override if the leader is one of the peers

Let me know if i understand your question wrongly

nadeemajl: would it be better to keep track of the leader explicitly? leader/follower state change is notified in handleRoleChange function in handler.go

bts: This implementation currently implicitly assumes there there is always a leader, which is not the case.

bts (commented):
I'm going to make a change shortly to address the discussion about `LeaderAddress`

raft/handler.go
trung: that's true. Simple test with 2 nodes showing that leader would never be selected.

trung: @nadeemajl role change event doesn't tell you who the current leader is

bts:
screenshot 2018-07-11 23 46 31

With this latest change we handle the lack of a leader and, as a starting point, we raise an error in web3. If anyone has a strong opinion to return e.g. null instead, we can do that as well.

trung:
thanks @bts , i just added a initialization of `leader` to be explicit with etcd convention of node id when there's no leader

bts:
That's great -- thanks for doing that! At this point I think we'll leave the PR to @jpmsam

jpmsam (approved):
@trung thanks for the contribution @bts thanks for the updates.

bts (approved):

Issue #428, opened 2018-07-02
stress testing on the 7node example causes unexpected behaviour


guojian1234:
#### System information

Geth version: 1.7.2-stable
Quorum Version 2.0.2
Go Version : go1.9.2

OS & Version: Centos7

Branch, Commit Hash or Release: Release Tag:v2.0.2

#### Expected behaviour
I did a test with around 150 transactions per second,  It should work all the time

#### Actual behaviour
a few hours later, Transactions cannot bt executed, and there is always an error :"nonce too high"。at this time , txpool.status is {pending:4096, queue :64}
I repeated the test several times,  Sometimes after a few hours, sometimes a few days, this error will occur

#### Steps to reproduce the behaviour
1、start quorum-example whit raft-start.sh
2、deploy contracts ENS.sol
3、use web3js call ens.setOwner()   ten times per second
ENS.sol is a very simple contracts as:
pragma solidity ^0.4.15;
contract ENS  {

    mapping(string =>string) records;
    struct Timer {
        uint256 time;
    }
    Timer timer;
    function ENS() public {
        timer.time = 0;
    }

    function setOwner(string node,string owner) public {
        records[node]= owner;
        addNumber(timer);
    }
}

#### Backtrace
6748557 TRACE[06-19|11:36:42|core/tx_pool.go:606]               add tx                                   nonce is :=303058 tx hash=5d5860…9e8f10
6748558 INFO [06-19|11:36:42|raft/minter.go:379]                TX success                               nonce=303056 txCount=2
6748559 TRACE[06-19|11:36:42|core/state_transition.go:207]      account nonce                            is :=303057 tx nonce is :=303057
6748560 DEBUG[06-19|11:36:42|core/state_transition.go:277]      set nonce                                LOG15_ERROR=                                                   LOG15_ERROR="Normalized odd number of arguments by adding nil"
6748561 INFO [06-19|11:36:42|raft/minter.go:379]                TX success                               nonce=303057 txCount=3
6748562 TRACE[06-19|11:36:42|core/tx_pool.go:666]               Pooled new future transaction            hash=5d5860…9e8f10 from=0xed9d02e382b34818e88B88a309c7fe71E65f419d to=0x9d13C6D3aFE1721BEef56B55D303B09E021E27ab nonce=303058
6748563 TRACE[06-19|11:36:42|core/tx_pool.go:919]               Promoting queued transaction             hash=5d5860…9e8f10
[...]

PR #433, opened 2018-07-04
fix issue #428

guojian1234:
fix issue #428 stress testing on the 7node example causes unexpected behaviour


jackOthera (commented):
It'd be good if we could get this approved, this seems to fix the issue we're having too.

fixanoid:
@guojian1234, @joelburget thanks! We're going in favor of the fix proposed at https://github.com/jpmorganchase/quorum/pull/437. Again, thanks for all the help.

PR #495, opened 2018-08-20
Fix sporadic crash due to insufficient gas

SatpalSandhu61:
Resolve issue where nodes crash when gas is on the limit for private contracts (https://github.com/QuorumEngineering/Quorum-Engineering-Team-Info/issues/40).
Root cause is that gas calculation was different depending on whether the node was a participant in the contract or not. Fix is to ensure that all nodes only use the public part of the transaction for calculating gas usage.
Fixes https://github.com/QuorumEngineering/Quorum-Engineering-Team-Info/issues/40

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Quorum
 
-<a href="https://quorumslack.azurewebsites.net" target="_blank" rel="noopener"><img title="Quorum Slack" src="https://quorumslack.azurewebsites.net/badge.svg" alt="Quorum Slack" /></a>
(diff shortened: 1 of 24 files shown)


fixanoid:
Tested with this provided document from @SatpalSandhu61 
[Quorum Testing - gas crashes.xlsx](https://github.com/jpmorganchase/quorum/files/2415814/Quorum.Testing.-.gas.crashes.xlsx)

fixanoid (approved):

SatpalSandhu61:
fixes github/QuorumEngineering/Quorum-Engineering-Team-Info#40

fixanoid (approved):

SatpalSandhu61:
Closing pull request to try a different approach which is less intrusive to the EVM.

PR #501, opened 2018-08-22
Feature/410/expose raft cluster api

apratt3377:
Give users the ability to see raft cluster information. raft.cluster exposes info of all the peers in the cluster (in pm.peers) including the raftId, nodeId, ip address, raft port, p2p port.  This can be used to check that all raft nodes came up that were expected.  Also can be used along with admin.peers to make sure the network was configured properly

### DIFF
--- a/internal/web3ext/web3ext.go
+++ b/internal/web3ext/web3ext.go
@@ -628,7 +628,11 @@ web3._extend({
                new web3._extend.Property({
                        name: 'leader',
                        getter: 'raft_leader'
-               })
+               }),
+							 new web3._extend.Property({
+							 				name: 'cluster',
+							 				getter: 'raft_cluster'
+							 })
        ]
 })
 `
--- a/internal/web3ext/web3ext.go
+++ b/internal/web3ext/web3ext.go
@@ -629,7 +629,7 @@ web3._extend({
                        name: 'leader',
                        getter: 'raft_leader'
                }),
-							 new web3._extend.Property({
+               new web3._extend.Property({
                        name: 'cluster',
                        getter: 'raft_cluster'
                }),
--- a/internal/web3ext/web3ext.go
+++ b/internal/web3ext/web3ext.go
@@ -630,9 +630,9 @@ web3._extend({
                        getter: 'raft_leader'
                }),
 							 new web3._extend.Property({
-							 				name: 'cluster',
-							 				getter: 'raft_cluster'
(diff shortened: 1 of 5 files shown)


internal/web3ext/web3ext.go
```diff
@@ -628,7 +628,11 @@ web3._extend({
                new web3._extend.Property({
                        name: 'leader',
                        getter: 'raft_leader'
-               })
+               }),
+               new web3._extend.Property({
+                       name: 'cluster',
```
bts, line 633: Because this new property does not contain information about the current node (which would then constitute a full "cluster"), in this current implementation it might be better named as "peers" instead?

E.g. currently for a cluster of size three it reports that it contains only two nodes, which could be confusing:

```
> raft.cluster.length
2
```

It seems like we should probably keep the name "cluster", but also include the current node's `RaftNodeInfo.Address` alongside the `RaftNodeInfo.PeerAddresses` that we're reporting back to the client.

bts (changes requested):
This looks good generally, but it seems that we should probably include the `Address` for the current/reporting node alongside those of its peers.

jpmsam (approved):

raft/api.go
```diff
@@ -35,5 +35,10 @@ func (s *PublicRaftAPI) Leader() (string, error) {
 	if nil != err {
 		return "", err
 	}
-	return addr.nodeId.String(), nil
+	return addr.NodeId.String(), nil
+}
+
+func (s *PublicRaftAPI) Cluster() []*Address {
+	return append(s.raftService.raftProtocolManager.NodeInfo().PeerAddresses,
+		s.raftService.raftProtocolManager.NodeInfo().Address)
```
bts: Can you extract `s.raftService.raftProtocolManager.NodeInfo()` to a local variable, and then reuse that value for accessing both `PeerAddresses` and `Address`? This is because `NodeInfo()` takes a lock and does a bunch of work:

https://github.com/jpmorganchase/quorum/blob/d71daf324590d0033d0fc0ad255cadb702cf2b3b/raft/handler.go#L184-L219

bts (changes requested):
I think we're really close -- can you make this one additional tweak? Thank you!

bts (approved):

bts:
@apratt3377 thanks!

@jpmsam ready to go

PR #502, opened 2018-08-23
EIP 155 Activation 

vsmk98:
As a part of EIP 155 activation, Quorum will not support chain id 1. The following changes are done as a part of this PR
1. If the chain id is 1 or network id is 1, geth start will fail with appropriate error messages
2. For migration scenarios where a network was running with a pre EIP155 version - the suggested approach is to bring down the network, update the chain id to a value > 1 in genesis.json file and run geth init with the new genesis file before bringing up the network 
3. Network id will be defaulted to 1337

### DIFF
--- a/params/config.go
+++ b/params/config.go
@@ -280,12 +280,12 @@ func (c *ChainConfig) checkCompatible(newcfg *ChainConfig, head *big.Int, isQuor
 	if isQuorumEIP155Activated && isForkIncompatible(c.EIP155Block, newcfg.EIP155Block, head) {
 		return newCompatError("EIP155 fork block", c.EIP155Block, newcfg.EIP155Block)
 	}
+	if isQuorumEIP155Activated && c.IsEIP155(head) && !configNumEqual(c.ChainId, newcfg.ChainId) {
+		return newCompatError("EIP155 chain ID", c.EIP155Block, newcfg.EIP155Block)
+	}
 	if isForkIncompatible(c.EIP158Block, newcfg.EIP158Block, head) {
 		return newCompatError("EIP158 fork block", c.EIP158Block, newcfg.EIP158Block)
 	}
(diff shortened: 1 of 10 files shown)


params/config.go
```diff
@@ -277,15 +277,15 @@ func (c *ChainConfig) checkCompatible(newcfg *ChainConfig, head *big.Int) *Confi
 	if isForkIncompatible(c.EIP150Block, newcfg.EIP150Block, head) {
 		return newCompatError("EIP150 fork block", c.EIP150Block, newcfg.EIP150Block)
 	}
-	if isForkIncompatible(c.EIP155Block, newcfg.EIP155Block, head) {
+	if isQuorumEIP155Activated && c.ChainId!=nil && isForkIncompatible(c.EIP155Block, newcfg.EIP155Block, head) {
 		return newCompatError("EIP155 fork block", c.EIP155Block, newcfg.EIP155Block)
 	}
+	if isQuorumEIP155Activated && c.ChainId!=nil && c.IsEIP155(head) && !configNumEqual(c.ChainId, newcfg.ChainId) {
+		return newCompatError("EIP155 chain ID", c.EIP155Block, newcfg.EIP155Block)
```
jpmsam, line 284: We should return the ChainId, not the EIP155Block.

jpmsam (approved):

coeniebeyers:
Apologies for resurrecting this, but [EIP155](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md) only mentions Chain ID and not Network ID

> 1. If the chain id is 1 or network id is 1, geth start will fail with appropriate error messages

Shouldn't this rather be just if Chain ID is 1 (and not if Network ID is 1), as per EIP155 ?

jpmsam:
Hi @coeniebeyers we also wanted to discourage using or keeping the default network ID, 1. We updated quorum's examples scripts to reflect that too.

Issue #481, opened 2018-08-14
The memory utilization is increasing in Istanbul BFT


hagishun:
#### System information
* ENV1
Version: 1.7.2-stable
Git Commit: 4a77480371683da5e56e48712d694c53d123b61a
Quorum Version: 2.0.2
Architecture: amd64
Network Id: 1
Go Version: go1.7.3
Operating System: linux
GOPATH=
GOROOT=/usr/local/go

* ENV2
Version: 1.7.2-stable
Git Commit: fd0e3b9d13f9b9563e7d5b3b1aba51f035f9cf0d
Quorum Version: 2.0.2
Architecture: amd64
Network Id: 1
Go Version: go1.7.3
Operating System: linux
GOPATH=
GOROOT=/usr/local/go

#### Expected behaviour
The memory utilization is periodically released.

#### Actual behaviour
Mamory utilization is increasing.
Only one node is periodically released.

#### Steps to reproduce the behaviour
Evnr

#### Backtrace
[log.zip](https://github.com/jpmorganchase/quorum/files/2285605/log.zip)

````
[backtrace]
````

fixanoid:
@hagishun could you give me some details of the cluster: whats the configuration like and what hardware was used? Also, are you getting same issued when building quorum client from master?

tharun-allu:
![image](https://user-images.githubusercontent.com/18491937/44354270-3d78ad00-a45e-11e8-94b6-53c8aee8e988.png)
This shows the memory utilization on the nodes I am running. Looks to me there might be some memory leak. 
The graph is 1 week utilization.

fixanoid:
@tharun-allu thanks for the metrics. Whats the load on the chain and how far did it advance blockwise?

tharun-allu:
@fixanoid the above graphs are from a network of nodes with 16G memory each and block height is 4 million.

I have restarted geth process in a different network yesterday with 7 million blocks and attached is the memory graph of the nodes (there are 4 in the network
![image](https://user-images.githubusercontent.com/18491937/44415998-a75c8980-a526-11e8-8833-f00e51115a1e.png)

What I noticed is 2 of the nodes shot up their memory and one node died and other will soon die.

namtruong:
Hi @tharun-allu I've been trying to replicate this issue but unable to. Did you get this only after it reached 4 million blocks? From the graph attached it seems quite stable for sometime before going up - was there any incident observed in the log?

tharun-allu:
[...]

PR #521, opened 2018-09-11
istanbul/core: fixed potential memory leak in `backlogs`

trung:
`backlogs` is the map with keys which are pointers to `Validator`. In a scenario when [node re-initializes](https://github.com/jpmorganchase/quorum/blob/master/consensus/istanbul/core/core.go#L230) the validator set, new set of pointers are instantiated causing new entry being added into `backlogs` when subsequent [`storeBacklogs()`](https://github.com/jpmorganchase/quorum/blob/master/consensus/istanbul/core/backlog.go#L91) is called.


fixanoid:
Fixes #481

fixanoid (approved):

jpmsam (approved):

marcosio:
Hi guys, we are introducing this fix in our nodes and: 

1. With 3 nodes, 1 validator: The fix works fine, the validator node has the minimum amount of memory, and if we stress the network with a lot of transactions, it is resilient.

2. With 48 nodes, 7 validators: We set 2 nodes with the modifications. Both of them increase her amount of memory for another reason. 

Any idea of another issue that can resolve this?

What can we test for identifying the origin of this memory leak?

Meanwhile, we will change our test configuration in the test environment with 2 validators and 2 regular nodes.

PR #541, opened 2018-09-27
Update signing issue for eth_sendTransactionAsync RPC call

prd-fox:
The eth_sendTrasactionAsync RPC call was implementing it's own logic to sign and submit a transaction to the pool, and the signing process wasn't updated for EIP-155. This meant private transactions were being incorrectly signed and got stuck in the TX pool.

It makes sense for the async version of eth_sendTransaction to do the same, but inside a goroutine so it can be executed whilst returning from the RPC call.

This PR also adds the missing required fields from the incoming message (the `restriction` field) and the to the outgoing callback message (the `id` field).


fixanoid (approved):

trung:
quick question: why was the safe guard channel that limits number of concurrent async execution removed?

fixanoid (dismissed):

fixanoid (dismissed):

fixanoid (approved):

jpmsam (approved):

PR #542, opened 2018-09-28
Geth upgrade 1.8.12

amalrajmani:
geth upgrade to 1.8.12

release notes here https://github.com/ethereum/go-ethereum/releases/tag/v1.8.12


jpmsam (approved):

params/protocol_params.go
```diff
@@ -77,6 +77,7 @@ const (
 	Bn256ScalarMulGas       uint64 = 40000  // Gas needed for an elliptic curve scalar multiplication
 	Bn256PairingBaseGas     uint64 = 100000 // Base price for an elliptic curve pairing check
 	Bn256PairingPerPointGas uint64 = 80000  // Per-point price for an elliptic curve pairing check
+	QuorumMaximumExtraDataSize uint64 = 65     // Maximum size extra data may be after Genesis.
```
eximchain137, line 80: 65 bytes is only large enough to put the signed block hash when you verify the header, why not pad 32 bytes in front similar to how geth does for [clique](https://github.com/jpmorganchase/quorum/blob/master/consensus/clique/clique.go#L58). Other wise should not allow this [flag](https://github.com/jpmorganchase/quorum/blob/master/cmd/utils/flags.go#L344) to set an extradata (this might be worth keeping to signal for forks (updates) in private consortia.

core/vm/interface.go
```diff
@@ -23,26 +23,36 @@ import (
 	"github.com/ethereum/go-ethereum/core/types"
 )
 
+// Quorum uses a cut-down StateDB, MinimalApiState. We leave the methods in StateDB commented out so they'll produce a
```
eximchain137: Is this design choice documented anywhere? is there a reason to not implement the full interface?

PR #555, opened 2018-10-17
p2p: remove spurious log for denied node permissioning

vdamle:
connection is only denied after cycling through entire list, remove
confusing and erroneous log in the loop


jpmsam (approved):

jpmsam:
@vdamle Thanks for the update.

PR #563, opened 2018-11-07
Do not reset V value if the tx is already set to private

Puneetha17:
When trying to send a private RawTransaction, the V value gets reset irrespective of what it previously was set to.

### DIFF
--- a/core/types/transaction.go
+++ b/core/types/transaction.go
@@ -486,6 +486,9 @@ func (tx *Transaction) IsPrivate() bool {
 }
 
 func (tx *Transaction) SetPrivate() {
+	if tx.IsPrivate() {
+		return
+	}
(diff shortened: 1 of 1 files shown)


jpmsam:
@Puneetha17 thanks for the update. Can you please complete the [CLA](https://github.com/jpmorganchase/cla/blob/master/cla-2016-09-01.pdf) and email it to quorum_cla@jpmorgan.com ?

vietlq:
@jpmsam you can get better collaboration experience by using a CLA bot like this: https://github.com/cla-assistant/cla-assistant

I signed a few CLAs using this one, very fast and no printing, no scanning, no hassle. The manual process wears contributors down.

jpmsam (approved):

PR #575, opened 2018-11-21
Add --txsizelimit flag to cli

michaelgu95:
## Overview
This PR introduces a transaction size limit flag `--txsizelimit` that extends the geth CLI within Quorum, effectively expanding the transaction size limit that is enforced in `tx_pool.go`

## Goal
Our enterprise customers have requested the ability to increase the size limit on transactions accepted by the transaction pool. We believe that in an enterprise setting, transactions that exceed 32KB are more common due for various reasons, such as the presence multi-party signatures. 

We aim to achieve this through the geth command line interface by adding an additional flag to set the txn size limit in KB (default=32). The limit is passed to the transaction pool's configuration and finally used in basic transaction validation.

## Files changed
* `flags.go`:

  - defined a new flag called `TxPoolSizeLimitFlag` of type `cli.Uint64Flag` with initial value 32
  - alter `setTxPool` to check if the flag is set in the global context, and if so we pass it to the transaction pool config (`TxPoolConfig` type)

* `usage.go` & `main.go`:
  - register the new flag definition

* `tx_pool.go`:
  - define a `SizeLimit` field in `DefaultTxPoolConfig` with  default value 32
  - leverage the new `SizeLimit` field in `config.sanitize()` to protect against invalid argument set by user, enforcing that  32 < SizeLimit < 128
  - finally, in `validateTx()` compare each incoming transaction's size to that set in the config ( `pool.config.SizeLimit` * 1024)

* `tx_pool_test.go`:
  - creates 2 tests for the flag, one negative and one positive. 
    - the positive test sets the flag to 128KB and tests that a transaction composed of `127KB+eth payload` will not throw `ErrOversizedData`, but a subsequent error is thrown instead (`ErrIntrinsicGas`)
    - the negative test sets flag to 128KB but tests that a transaction composed of `(128KB+1byte)+eth payload` throws `ErrOversizedData` as expected


prd-fox:
If two nodes in the same network have different sizes, then one will accept a transaction another would reject. If the node with the higher value is able to mint a block, then the other node would not be able to process that block (due to not accepting the transaction) and not be able to continue.

In an untrusted network, this has the effect that you have to assume the largest possible size for transactions, since you don't know if any of the nodes are malicious and going to create these "bad" blocks. Thus, you can't really have anything other than the highest value set (128kb).

Whilst the risk is lessened in a trusted network, the same concern applies. Also, if one is not going to send transactions larger than 64kb, then setting the upper limit to 128kb has no down side (aside from client enforcement, rather than network enforcement), and has the potential to be very error prone.
It may be non-obvious that a transaction is missing from the resulting errors that users may be presented with, and debugging such an error would be time-consuming (did we lose the transaction due to size, or other network problems? the latter is much more serious).
So the best option is to set the maximum allowed value.

I would therefore argue that being able to set the maximum size of transactions is not a good idea, at least not on the CLI (or anything that can be changed easily). Setting it to different sizes during hard-forks is one option, but hard-forks are usually hard coded.

michaelgu95:
>  If the node with the higher value is able to mint a block, then **the other node would not be able to process that block** (due to not accepting the transaction) and not be able to continue.

Our team had the same concern and went ahead and tested whether this assumption holds. We published our results here: https://github.com/kaleido-io/txsizelimit-testing

Most importantly we found that mixed transaction size limits across the blockchain network **does not result in non-determinism or corrupt the blockchain**. 

Though a transaction may be rejected during pool-level validation by a node with lower `txsizelimit`, the node is still able to recognize the block by which it was mined (so long as another node with higher `txsizelimit` mines it).

Our test shows that adding a transaction size limit does not cause network affects beyond pool-level validation; it does not halt the block mining process or cause "bad blocks", which is the basis of your concern.

Feel free to confirm our findings by running our test environment, or explore various network configurations as well if you believe your claims still hold true.

jpmsam:
> Most importantly we found that mixed transaction size limits across the blockchain network does not result in non-determinism or corrupt the blockchain.

That is correct, no concerns. 

> Though a transaction may be rejected during pool-level validation by a node with lower txsizelimit, the node is still able to recognize the block by which it was mined (so long as another node with higher txsizelimit mines it).

This is the concern since you have no guarantee that the validator nodes have updated their params to accept transactions with the updated size and the transaction will remain in the txpool.

Based on our last discussion on slack, you were exploring making this configuration in the genesis block so all nodes on the network can agree on the new value.

fixanoid:
Hey @michaelgu95, how are you? Are you planning to update the pull to have this value as part of the genesis? Otherwise, I'd be happy to take a stab at it on our end. Thanks!

PR #579, opened 2018-11-25
Notes for installation on Ubuntu

whittlem:


vietlq (changes requested):
Just add a link to the official tutorial how to install on Ubuntu. The error log is too verbose and unnecessary.

fixanoid:
There is a bit of knowledge required to work with quorum/geth. Since this is a fork of a go-ethereum project its implied that you need go compiler installed before you could build from source. I added a note to building doc, but I don't believe we need an os related instruction. Thank you.

PR #624, opened 2019-01-22
Re-enable gas price

aroommen:
The changes here allow users to specify a non-zero gas price for Quorum transactions. Previously Quorum only worked with a zero gas price. This is useful to enable network metering especially in consortium projects where the network enablers can be compensated for their costs. It's backward compatible in that the gas price can still be zero. 
@fixanoid  @joelburget I have incorporated code changes from the below pull request 
https://github.com/jpmorganchase/quorum/pull/288

### DIFF
--- a/core/tx_pool.go
+++ b/core/tx_pool.go
@@ -77,6 +77,8 @@ var (
 	// making the transaction invalid, rather a DOS protection.
 	ErrOversizedData = errors.New("oversized data")
 
+	ErrInvalidGasPrice = errors.New("Gas price not 0")
+
 	// ErrEtherValueUnsupported is returned if a transaction specifies an Ether Value
 	// for a private Quorum transaction.
 	ErrEtherValueUnsupported = errors.New("ether value is not supported for private transactions")
--- a/core/tx_pool.go
+++ b/core/tx_pool.go
@@ -77,8 +77,6 @@ var (
 	// making the transaction invalid, rather a DOS protection.
 	ErrOversizedData = errors.New("oversized data")
 
-	ErrInvalidGasPrice = errors.New("Gas price not 0")
-
 	// ErrEtherValueUnsupported is returned if a transaction specifies an Ether Value
 	// for a private Quorum transaction.
 	ErrEtherValueUnsupported = errors.New("ether value is not supported for private transactions")
--- a/core/tx_pool.go
+++ b/core/tx_pool.go
@@ -148,7 +148,7 @@ var DefaultTxPoolConfig = TxPoolConfig{
 	Journal:   "transactions.rlp",
 	Rejournal: time.Hour,
 
-	PriceLimit: 0;
+	PriceLimit: 0,
 	PriceBump:  10,
 
 	AccountSlots: 16,
@@ -722,7 +722,6 @@ func (pool *TxPool) journalTx(from common.Address, tx *types.Transaction) {
 	}
 }
 
-
 // promoteTx adds a transaction to the pending (processable) list of transactions
 // and returns whether it was inserted or an older was better.
 //
@@ -1203,7 +1202,6 @@ func newTxLookup() *txLookup {
 		all: make(map[common.Hash]*types.Transaction),
 	}
 }
-
 
 // Range calls f on each key and value present in the map.
 func (t *txLookup) Range(f func(hash common.Hash, tx *types.Transaction) bool) {
--- a/core/tx_pool.go
+++ b/core/tx_pool.go
@@ -148,7 +148,7 @@ var DefaultTxPoolConfig = TxPoolConfig{
 	Journal:   "transactions.rlp",
 	Rejournal: time.Hour,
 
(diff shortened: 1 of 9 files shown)


SatpalSandhu61 (changes requested):
I've done some initial testing and found one issue (on raft) - if I submit a transaction with zero gasPrice, it goes into pending state and is never completed, e.g.:
eth.sendTransaction({from:"0xed9d02e382b34818e88b88a309c7fe71e65f419d",to:"0xcc71c7546429a13796cf1bf9228bff213e7ae9cc",value:1,gasPrice:0,gas:21000})

Also, the changes are causing a number of acceptance tests to fail (https://github.com/jpmorganchase/quorum-acceptance-tests). I'll try and dig deeper into the reasons for these - but it may just be due to some transactions getting stuck in pending and causing following tests to fail.

aroommen:
Thanks @SatpalSandhu61 
am looking into the pending transfer issue.

prd-fox:
Further to Satpal's comment above, this is due to the transactions not being accepted on remote nodes due to a too low gas price.
The minimum accepted gas price can be used using the `--txpool.pricelimit` flag, but when set below 1, it gets sanitised back to 1; this means that a gas price of 0 is never allowed on remote nodes.

See txpool.go:172 for the condition, which should be reviewed. The check should compare against a Price Limit of 0.

aroommen:
I have adjusted the flow in tx_pool.go to reset the default Transaction price limit for Quorum to zero and all tests are passing. Do check and let me know if the changes are ok

core/tx_pool.go
```diff
@@ -173,6 +171,10 @@ func (config *TxPoolConfig) sanitize() TxPoolConfig {
 		log.Warn("Sanitizing invalid txpool price limit", "provided", conf.PriceLimit, "updated", DefaultTxPoolConfig.PriceLimit)
 		conf.PriceLimit = DefaultTxPoolConfig.PriceLimit
 	}
+	if isQuorum && (conf.PriceLimit == DefaultTxPoolConfig.PriceLimit) {
+		log.Warn("Sanitizing invalid txpool price limit for Quorum", "provided", conf.PriceLimit, "updated 0")
```
SatpalSandhu61: There's a small issue with the log statement. An even number of parameters are expected after the log message - the string "updated 0" is expected to be followed by a variable which is to be printed.
Because the variable is missing, the log statement is catching that and printing:
    LOG15_ERROR="Normalized odd number of arguments by adding nil"

core/tx_pool.go
```diff
@@ -173,6 +171,10 @@ func (config *TxPoolConfig) sanitize() TxPoolConfig {
 		log.Warn("Sanitizing invalid txpool price limit", "provided", conf.PriceLimit, "updated", DefaultTxPoolConfig.PriceLimit)
 		conf.PriceLimit = DefaultTxPoolConfig.PriceLimit
 	}
+	if isQuorum && (conf.PriceLimit == DefaultTxPoolConfig.PriceLimit) {
```
SatpalSandhu61: This will allow Quorum users to have a price of zero, or > 1. However a value of 1 will be reset to zero... I don't think this is what you want to achieve.
Is there a reason why you didn't stick with your change to just have a default price of zero?

core/tx_pool.go
aroommen: Ok. WIll make the change.

core/tx_pool.go
aroommen: Yes. Another testcase was failing and as this DefaultTxPoolConfig was used in multiple files, I didn't want to impact all that code. 
Since then I have revised the code and will be resubmitting it with a zero setting instead of 1.

aroommen: Yes. Another testcase was failing and as this DefaultTxPoolConfig was used in multiple files, I didn't want to impact all that code.
Since then I have revised the code and will be resubmitting it with a zero setting instead of 1.

SatpalSandhu61:
Note that this PR is on our backlog, pending some associated changes that are being worked on, and some further testing.

pjworrall:
> Note that this PR is on our backlog, pending some associated changes that are being worked on, and some further testing.

Great to here @SatpalSandhu61 . As soon as this pull is in we'll be able to migrate to Quorum because I need transactions to require Eth. As there is no mining I am keen to know where the transaction fee goes: to the Validator?

SatpalSandhu61:
@pjworrall the account where the fees go is actually the bit we're working on. It will be the coinbase account, or the account associated with the minting node. There are some nuances which are currently being ironed out, around the relationship of the account on the type of consensus used.

SatpalSandhu61:
Note that this PR should not be merged as it's not complete. Work to enable gas price is ongoing on a different branch.

Yur13L:
Any news on this?

antonydenyer:
Superseded by https://github.com/ConsenSys/quorum/pull/1446

PR #633, opened 2019-02-01
Feature/private abigen

zzy96:
Enabling privateFor feature for sending private transactions via native go dapp


ethclient/ethclient.go
```diff
@@ -490,6 +491,35 @@ func (ec *Client) SendTransaction(ctx context.Context, tx *types.Transaction) er
 	return ec.c.CallContext(ctx, nil, "eth_sendRawTransaction", common.ToHex(data))
 }
 
+// PreparePrivateTransaction sends the encoded raw transaction to Constellation,
+// returning the encoded commitment transaction.
+func (ec *Client) PreparePrivateTransaction(ctx context.Context, encodedTx hexutil.Bytes, privateFrom string, privateFor []string) (hexutil.Bytes, error) {
+	if len(privateFor) == 0 {
```
Krish1979: The privateFor object is NULL only for public transactions , but for transactions to itself it has a blank value. Checking length for zero will prevent a node sending private transactions to itself. So please change the check to != NULL rather..

vsmk98: changed to allow transaction to self node.

Krish1979 (dismissed):

zzy96:
Notes on all changes

##### accounts/abi/bind/base.go
- add PrivateFrom/PrivateFor in TransactOpts
- in transact function, which is called by go native dapp binding code, check if PrivateFor is nil
- preparePrivateTransaction will call the PreparePrivateTransaction function of ContractTransactor implementation
##### accounts/abi/bind/backend.go
- add PreparePrivateTransaction in ContractTransactor interface
##### accounts/abi/bind/backends/simulated.go
- add a dummy implementation of PreparePrivateTransaction
##### core/types/transaction.go
- add SetData method to change transaction payload, called by PreparePrivateTransaction
##### ethclient/ethclient.go
- go dapp will use ethclient to connect to a node, the actual implementation of PreparePrivateTransaction is here
##### internal/ethapi/api.go
- PrivateFor should allow empty array
##### docs/private-abigen.md
- documentation

core/types/transaction.go
```diff
@@ -185,7 +186,11 @@ func (tx *Transaction) UnmarshalJSON(input []byte) error {
 	return nil
 }
 
-func (tx *Transaction) Data() []byte       { return common.CopyBytes(tx.data.Payload) }
+func (tx *Transaction) Data() []byte { return common.CopyBytes(tx.data.Payload) }
+func (tx *Transaction) SetData(data []byte) {
```
trung: IMO, transaction data is supposed to be immutable after transaction is created. Exposing `SetData`  breaks this design.

ethclient/ethclient.go
```diff
@@ -506,6 +507,35 @@ func (ec *Client) SendTransaction(ctx context.Context, tx *types.Transaction) er
...
+	if private.P == nil {
+		return nil, errors.New("transaction manager not set up")
+	}
+	data, err := private.P.Send(tx.Data(), privateFrom, privateFor)
+	if err != nil {
+		return nil, err
+	}
+
+	tx.SetPrivate()
+	tx.SetData(data)
```
trung: Use `types.NewTransaction` or `types.NewContractCreation` instead

trung:
Please add/modify unit tests for new changes

ethclient/ethclient.go
zzy96: I have removed SetData and use types.NewTransaction/types.NewContractCreation to create private transaction

zzy96:
@trung I have optimized all changes based on your very helpful review comments.

The key code changes are private inside accounts/abi/bind/base.go, so it may not be suitable for unit test. An integration test sample is available [here](https://github.com/zzy96/quorum/blob/feature/privateAbigen-with-test/contracts/simple/simple_storage_test.go)

### NEW notes on all changes comparing to master
##### accounts/abi/bind/base.go
- add PrivateFrom/PrivateFor in TransactOpts
- in transact function, which is called by go native dapp binding code, check if PrivateFor is nil
- preparePrivateTransaction replace the payload of private transaction to the hash from Tessera/Constellation
- sendPrivateTx send the private transaction to Tessera/Constellation
##### internal/ethapi/api.go
- PrivateFor should allow empty array
##### docs/private-abigen.md
- documentation

accounts/abi/bind/base.go
```diff
@@ -340,6 +356,30 @@ func (c *BoundContract) UnpackLog(out interface{}, event string, log types.Log)
...
+	privateTx.SetPrivate()
+	return privateTx
+}
+
+// sendPrivateTx send the private transaction to Tessera/Constellation
+func sendPrivateTx(tx *types.Transaction, privateFrom string, privateFor []string) ([]byte, error) {
+	if private.P == nil {
+		return nil, errors.New("transaction manager not set up")
+	}
+	data, err := private.P.Send(tx.Data(), privateFrom, privateFor)
```
trung: this calls `/sendraw` which is only meant for Quorum node.
Native DApps can only call `/storeraw`

trung:
From [Native DApps: Go bindings to Ethereum contracts · ethereum/go-ethereum Wiki · GitHub](https://github.com/ethereum/go-ethereum/wiki/Native-DApps:-Go-bindings-to-Ethereum-contracts#transacting-with-an-ethereum-contract)
> Invoking a method that changes contract state (i.e. transacting) is a bit more involved, as a live transaction needs to be authorized and broadcast into the network.
> Opposed to the conventional way of storing accounts and keys in the node we attach to, 
> **Go bindings require signing transactions locally and do not delegate this to a remote node.** 
> This is done so to facilitate the general direction of the Ethereum community where accounts are kept private to DApps, and not shared (by default) between them.

It means we need to make a change so that externally-signed private transactions are supported. The correct flow for this is:
- Client (Go bindings) calls Private Transaction Manager's `/storeraw` API and receives encrypted payload hash
- Client constructs a private transaction with payload is the encrypted payload hash
- Client signs the transaction
- Client calls Quorum's `eth_sendRawPrivateTransaction` API

The logic in the PR is indeed utilizing different APIs:
- Client calls Private Transaction Manager's `/sendraw` API and receives encrypted payload hash
- Client constructs a private transaction with payload is the encrypted payload hash
- Client signs the transaction
- Client calls Ethereum's `eth_sendRawTransaction` API

The misassumptions are:
- `/sendraw` is a public API. Please refer to [Interfaces & API - Tessera](https://docs.goquorum.com/en/latest/Privacy/Tessera/Usage/Interface%20&%20API/) for more information
- Reusing `eth_sendRawTransaction` to handle externally-signed private transactions

zzy96:
@trung 

I have read through this wiki again. In this [example](https://github.com/ethereum/go-ethereum/wiki/Native-DApps:-Go-bindings-to-Ethereum-contracts#transacting-with-an-ethereum-contract), it is passing the `auth` which has a field `Signer` to the `TransactOpts`. Therefore, I think the current flow for signing is fine, it stays the same with public abigen.

About the API used, I have changed go binding code to call Tessera's `/storeraw` API and ethclient implementation to call Quorum's `eth_sendRawPrivateTransaction` API. Somehow `eth_sendRawPrivateTransaction` still doesn't work, I will continue to work on it. In the meanwhile please help me check if my changes to call these APIs make sense. Thanks!

accounts/abi/bind/base.go
```diff
@@ -49,6 +53,10 @@ type TransactOpts struct {
 	Nonce  *big.Int       // Nonce to use for the transaction execution (nil = use pending state)
 	Signer SignerFn       // Method to use for signing the transaction (mandatory)
 
+	TransactionManagerURL string   // Tessera/Constellation URL to call /storeraw API.
```
trung: `ethclient.Client` struct defines wrappers for Ethereum RPC API, it's the right place to define wrappers for Private Transaction Manager public API.

accounts/abi/bind/base.go
```diff
@@ -340,6 +368,58 @@ func (c *BoundContract) UnpackLog(out interface{}, event string, log types.Log)
...
+type StoreRawResp struct {
+	Key string `json:"key"`
+}
+
+// storeRaw calls the /storeraw API of tessera
+func storeRaw(data []byte, privateFrom string, url string) ([]byte, error) {
+	reqBodyJSON := &StoreRawArgs{
+		Payload: string(data),
+		From:    privateFrom}
+	reqBody, _ := json.Marshal(reqBodyJSON)
```
trung: nitpick: return error if not nil

accounts/abi/bind/base.go
```diff
@@ -340,6 +368,58 @@ func (c *BoundContract) UnpackLog(out interface{}, event string, log types.Log)
...
+	resp, err := http.Post(url+"/storeraw", "application/json", bytes.NewBuffer(reqBody))
+
+	if err != nil {
+		return nil, errors.New("Failed to call /storeRaw API on transaction manager.")
+	}
+	respBody, _ := ioutil.ReadAll(resp.Body)
+	defer resp.Body.Close()
+
+	respBodyJSON := StoreRawResp{}
+	json.Unmarshal(respBody, &respBodyJSON)
```
trung: nitpick: return error if not nil

accounts/abi/bind/base.go
```diff
@@ -340,6 +368,58 @@ func (c *BoundContract) UnpackLog(out interface{}, event string, log types.Log)
...
+	reqBodyJSON := &StoreRawArgs{
+		Payload: string(data),
+		From:    privateFrom}
+	reqBody, _ := json.Marshal(reqBodyJSON)
+	resp, err := http.Post(url+"/storeraw", "application/json", bytes.NewBuffer(reqBody))
+
+	if err != nil {
+		return nil, errors.New("Failed to call /storeRaw API on transaction manager.")
+	}
+	respBody, _ := ioutil.ReadAll(resp.Body)
```
trung: nitpick: return error if not nil

accounts/abi/bind/base.go
```diff
@@ -340,6 +368,58 @@ func (c *BoundContract) UnpackLog(out interface{}, event string, log types.Log)
...
+	Key string `json:"key"`
+}
+
+// storeRaw calls the /storeraw API of tessera
+func storeRaw(data []byte, privateFrom string, url string) ([]byte, error) {
+	reqBodyJSON := &StoreRawArgs{
+		Payload: string(data),
+		From:    privateFrom}
+	reqBody, _ := json.Marshal(reqBodyJSON)
+	resp, err := http.Post(url+"/storeraw", "application/json", bytes.NewBuffer(reqBody))
```
trung: Private Transaction Manager (Tessera) third-party API server supports [multiple ways of communications](http://docs.goquorum.com/en/latest/Privacy/Tessera/Configuration/Configuration%20Overview/#server). If `ethclient` only support HTTP, need to be explicit and handling check explicitly

accounts/abi/bind/base.go
zzy96: This is quite tricky. It seems that we cannot create transaction in `ethclient.Client` if we want to use abigen go binding code to create transaction.

trung: `ethclient.Client` implements `ContractBackend` interface and hence it is the implementation of `ContractTransactor` interface.

zzy96:
#112 #146

accounts/abi/bind/base.go
zzy96: Adding `TransactionManagerURL` as a field in `ethclient.Client` struct

accounts/abi/bind/base.go
zzy96: Make `ethclient` function name explicit as `StoreRawHTTP`

ethclient/ethclient.go
```diff
@@ -34,7 +39,8 @@ import (
 
 // Client defines typed wrappers for the Ethereum RPC API.
 type Client struct {
-	c *rpc.Client
+	c                         *rpc.Client
+	TransactionManagerHTTPURL string // Tessera/Constellation HTTP URL to call /storeraw API.
```
trung, line 43: As discussed, please create a `PrivateTransactionManagerClient` to handle `/storeraw` API call

ethclient/ethclient.go
```diff
@@ -498,12 +504,55 @@ func (ec *Client) EstimateGas(ctx context.Context, msg ethereum.CallMsg) (uint64
...
+	}
+}
+
+// PreparePrivateTransaction calls the /storeraw API of tessera
+func (ec *Client) PreparePrivateTransaction(data []byte, privateFrom string) ([]byte, error) {
+	if ec.TransactionManagerHTTPURL == "" {
+		return nil, errors.New("transaction manager url is nil")
+	}
+
+	reqBodyJSON := map[string]string{
```
trung, line 525: Model `xxRequest` and `xxResponse` instead of generic map

trung (changes requested):

zzy96:
Create a new PR in #819

PR #636, opened 2019-02-12
Adding support for transaction size limit via genesis

fixanoid:
Builds on top of #575 by @michaelgu95 but moves transaction size limit into genesis config

This PR introduces a transaction size limit genesis config within Quorum, effectively expanding the transaction size limit that is enforced in `tx_pool.go`. Quorum default transaction size limit is set to 64kb.


params/config.go
```diff
@@ -207,6 +209,14 @@ func (c *ChainConfig) String() string {
 	)
 }
 
+func (c *ChainConfig) IsValid() error {
+	if c.SizeLimit < 32 || c.SizeLimit > 128 {
```
fixanoid: TODO: Determine sane sizes

fixanoid: Quorum block gas limit is set to be 3,758,096,384 and Quorum is gas-less, so the txn gas price is not accounted for. This means, that about 6MB of data fits into the block, which in turns means that our maximum transaction size limit could theoretically be up to 6000KB

core/tx_pool.go
```diff
@@ -133,6 +133,8 @@ type TxPoolConfig struct {
 	Journal   string        // Journal of local transactions to survive node restarts
 	Rejournal time.Duration // Time interval to regenerate the local transaction journal
 
+	SizeLimit uint64 // Maximum size allowed for valid transaction (in KB)
```
SatpalSandhu61: Would it be worthwhile renaming this to 'TransactionSizeLimit' to make it more descriptive?

fixanoid: Agreed. Originally, SizeLimit was just part of tx_pool, so it was clearer in that context. Renaming everywhere applicable.

SatpalSandhu61 (approved):

params/config.go
fixanoid: Leaving at 128 until a reasonable request for a higher limit comes up.

jimthematrix (approved):
Looks good to me!

jpmsam (approved):

vdamle:
@fixanoid - Does the RPC server payload size limit require an adjustment to go along with this change? It is currently `64KB` - https://github.com/jpmorganchase/quorum/blob/master/rpc/server.go#L130

PR #639, opened 2019-02-14
geth upgrade 1.8.18

amalrajmani:


p2p/enode/node.go
```diff
@@ -82,6 +82,20 @@ func (n *Node) UDP() int {
 	return int(port)
 }
 
+// RAFTPORT returns the RAFT PORT of the node
+func (n *Node) RAFTPORT() int {
```
jbhurat: The method RAFTPORT() shouldn't be in CAPS

p2p/enode/node.go
```diff
@@ -205,6 +234,20 @@ func HexID(in string) ID {
 	return id
 }
 
+//TODO(Amal): to review
+func RaftHexID(in string) (EnodeID, error) {
+	var id EnodeID
+	b, err := hex.DecodeString(strings.TrimPrefix(in, "0x"))
+	if err != nil {
+		return id, err
```
jbhurat: Return `nil, err` instead

p2p/enode/node.go
amalrajmani: The method name for other attributes like UPD, TCP, ID and IP in p2p/enode/node.go are all in CAPS. They are following this convention for KEY names basically. I have followed the same style to keep it consistent.

jbhurat: I think the reason why `UDP`, `TCP`, `IP` are in CAPS because they are all acronyms and `RAFTPORT` is not, hence it need not be in CAPS

amalrajmani: alright, changed it to RaftPort

SatpalSandhu61 (dismissed):

jbhurat (dismissed):

SatpalSandhu61 (approved):

jbhurat (approved):

jpmsam (approved):

PR #673, opened 2019-04-04
ETHAPI: Private tx bug fixes for Sign and Send APIs

### LINKED ISSUE #677: ETHAPI: Private transactions sent using personal_sendTransaction are not processed correctly
#### System information
Geth
Version: 1.8.18-stable
Git Commit: 17f558dae978f37441940dcf63a577032945ce51
Quorum Version: 2.2.2
Architecture: amd64
Protocol Versions: [63 62]
Network Id: 1337
Go Version: go1.11.5
Operating System: darwin

#### Expected behaviour
A private tx sent using the `personal_sendTransaction` endpoint should store the tx `data` in the transaction manager (TM) and replace the `tx.data` field with the encrypted tx hash returned from the TM.

#### Actual behaviour
The encrypted tx hash returned from the TM is not being used to replace the `tx.data` field.  This means that when executing the tx, the TM is unable to provide the bytecode as it is doing a DB lookup for the bytecode itself and not the encrypted tx hash.

With a fresh Quorum/Tessera 7 nodes network running, and while attached to node 1:
```bash
personal.lockAccount(web3.eth.accounts[0])
var bytecode = "0x6060604052341561000f57600080fd5b604051602080610149833981016040528080519060200190919050505b806000819055505b505b610104806100456000396000f30060606040526000357c0100000000000000000000000000000000000000000000000000000000900463ffffffff1680632a1afcd914605157806360fe47b11460775780636d4ce63c146097575b600080fd5b3415605b57600080fd5b606160bd565b6040518082815260200191505060405180910390f35b3415608157600080fd5b6095600480803590602001909190505060c3565b005b341560a157600080fd5b60a760ce565b6040518082815260200191505060405180910390f35b60005481565b806000819055505b50565b6000805490505b905600a165627a7a72305820d5851baab720bba574474de3d09dbeaabc674a15f4dd93b974908476542c23f00029"
var params = "000000000000000000000000000000000000000000000000000000000000002a"
var tx = {from: web3.eth.accounts[0], data: bytecode+params, gas: 0x47b760, privateFor: ["ROAZBWtSacxXQrOe3FGAqJDyJjFePR5ce4TSIzmJ0Bc="]}
personal.sendTransaction(tx, "")

var address="0x1932c48b2bf8102ba33b4a6b545c32236e342f34"
var abi = [{"constant":true,"inputs":[],"name":"storedData","outputs":[{"name":"","type":"uint256"}],"payable":false,"type":"function"},{"constant":false,"inputs":[{"name":"x","type":"uint256"}],"name":"set","outputs":[],"payable":false,"type":"function"},{"constant":true,"inputs":[],"name":"get","outputs":[{"name":"retVal","type":"uint256"}],"payable":false,"type":"function"},{"inputs":[{"name":"initVal","type":"uint256"}],"type":"constructor"}];
var private = eth.contract(abi).at(address)
private.get()
```
This will deploy the simple-contract provided in the 7nodes examples and retrieve the state for the contract.  The expected value is `42` but the actual returned value is `0`.  Looking in `tessera1.log` confirms that Tessera is unable to retrieve the tx bytecode for Quorum.

chris-j-h:
This PR addresses bugs when using 3 of the sign and send API endpoints with private transactions:

1. `personal_signTransaction`
   **Bug**: Private txs signed using this endpoint are not minted into a block when submitted.
   Currently the tx is not checked to be private and so is always signed using the EIP155 signer (i.e. the signature includes the `chainID`).  When submitting the tx (e.g. with `eth_sendRawPrivateTransaction`) if the tx is private the signature is validated using the Homestead signer (i.e. assumes the signature does not include the `chainID`).  
   
    **Fix**: Check the tx is private before setting the `chainID`.  Set the `v` value to `37/38` so that the resulting signed tx has the correct `v` value.  This ensures it is signed with the Homestead signer.

2. `personal_sendTransaction`
   Fixes #677 

3. `eth_signTransaction`
   **Bug**: Private txs signed using this endpoint are not minted into a block when submitted.
   Similar to the `personal_signTransaction` bug, the signing component of this method is unable to determine whether the tx is private and so always uses the EIP155 signer.  

   **Fix**: Set the `v` of the private tx to `37/38` before the signing to ensure it is signed using the Homestead signer

### DIFF
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -398,7 +398,7 @@ func (s *PrivateAccountAPI) SendTransaction(ctx context.Context, args SendTxArgs
 		data := []byte(*args.Data)
 		if len(data) > 0 {
 			log.Info("sending private tx", "data", fmt.Sprintf("%x", data), "privatefrom", args.PrivateFrom, "privatefor", args.PrivateFor)
-			data, err := private.P.Send(data, args.PrivateFrom, args.PrivateFor)
+			data, err = private.P.Send(data, args.PrivateFrom, args.PrivateFor)
 			log.Info("sent private tx", "data", fmt.Sprintf("%x", data), "privatefrom", args.PrivateFrom, "privatefor", args.PrivateFor)
 			if err != nil {
 				return common.Hash{}, err
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -366,10 +366,12 @@ func (s *PrivateAccountAPI) signTransaction(ctx context.Context, args *SendTxArg
 	// Assemble the transaction and sign with the wallet
 	tx := args.toTransaction()
 
-	isPrivate := args.PrivateFor != nil
+	if len(args.PrivateFor) > 0 {
+		tx.SetPrivate()
+	}
 
 	var chainID *big.Int
-	if config := s.b.ChainConfig(); config.IsEIP155(s.b.CurrentBlock().Number()) && !isPrivate {
+	if config := s.b.ChainConfig(); config.IsEIP155(s.b.CurrentBlock().Number()) && !tx.IsPrivate() {
 		chainID = config.ChainID
 	}
 	return wallet.SignTxWithPassphrase(account, passwd, tx, chainID)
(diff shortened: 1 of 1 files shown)


chris-j-h:
Need to do some further investigation into the `personal_signTransaction` and `eth_signTransaction` bugs so will close this PR and raise a separate PR for the `personal_sendTransaction` bug for the time being.

PR #680, opened 2019-04-08
raft: pass gas floor and gas ceil to calcGasLimit when minting block

amalrajmani:
raft: pass gas floor and gas ceil to calcGasLimit when minting block
currently raft is using parent gas limit for both floor and ceil and  this change is a fix for it


fixanoid (dismissed):

raft/backend.go
```diff
@@ -36,7 +36,7 @@ type RaftService struct {
 	nodeKey  *ecdsa.PrivateKey
 }
 
-func New(ctx *node.ServiceContext, chainConfig *params.ChainConfig, raftId, raftPort uint16, joinExisting bool, blockTime time.Duration, e *eth.Ethereum, startPeers []*enode.Node, datadir string) (*RaftService, error) {
+func New(ctx *node.ServiceContext, chainConfig *params.ChainConfig, raftId, raftPort uint16, joinExisting bool, blockTime time.Duration, e *eth.Ethereum, startPeers []*enode.Node, datadir string, ethCfg *eth.Config) (*RaftService, error) {
```
trung: `e.Config` gives you `*eth.Config`, don't need to pass `ethCfg`

trung: changing function signature and passing config around are not the best and in this case it seems the logic could be encapsulated

`eth.Ethereum` has the `eth.Config`. I think it'd be better to create a function in `eth.Ethereum` to do `CalcGasLimit` and inject into the `RaftService` so `raft/minter.go` can utilize that function when `createWork()`

my 2 cents

jpmsam (approved):

robinbryce:
On 2.2.4 We originaly set a very high transaction gasLimit - for our scenario we don't care about per tx gasLimit and we don't care about the block level gasLimit. After running for a while, the limit produced by CalcGasLimit was less than our 'fixed' per transaction gasLimit. With the result that all transactions were then rejected by the preCheck with 'exceeds block gas limit" error. Its time consuming to track this down if not aware of the mechanics of this dyamic block gasLimit. We now ensure that 'floor' is very high and transaction gasLimit is < 'floor'. We don't care about the ceiling but considered setting genesis.gasLimit < floor, and floor to be some modestly high value, so we would later have the option of lowering the value for --miner.gastarget and re-enabling the dynamic block gas limit. What does the dynamic limit produced by CalcGasLimit actually give us for a private network ? Is there a reason not to have a fixed limit ?

PR #693, opened 2019-04-29
Add https rpc server

sc0Vu:
Add HTTPS-RPC server.

Should I add https/tls to http.Client (like code in closed pull request from go-ethereum)?


jpmsam:
We have implemented a plugin based approach via pr #1019

PR #699, opened 2019-04-30
Bugfix/geth console displays incorrect block timestamp

### LINKED ISSUE #698: Raft consensus: geth console displays incorrect block timestamp
#### System information

Version: 1.8.18-stable
Git Commit: 2d22fd0044a9370d0e3f97227354b2fb644586df
Quorum Version: 2.2.3
Architecture: amd64
Protocol Versions: [63 62]
Network Id: 1337
Go Version: go1.10.4
Operating System: darwin

#### Description

When 'geth console' is executed, the console displays the timestamp of the latest block, e.g.:
```
Welcome to the Geth JavaScript console!

instance: Geth/v1.8.18-stable-2d22fd00(quorum-v2.2.3)/darwin-amd64/go1.10.4
coinbase: 0xed9d02e382b34818e88b88a309c7fe71e65f419d
at block: 5 (Fri, 28 Nov 49325133006 11:29:03 GMT)
```

As you can see, the timestamp is incorrect in the year.
This is because the console expects the block timestamp to be in seconds, but for Raft it is held in nanoseconds.
The relevant code is in console.console.go::Welcome(), where it has the code:
```
console.log("at block: " + eth.blockNumber + " (" + new Date(1000 * eth.getBlock(eth.blockNumber).timestamp) + ")");
```
Under Raft, this code should be:
```
console.log("at block: " + eth.blockNumber + " (" + new Date(eth.getBlock(eth.blockNumber).timestamp / 1000000) + ")");
```
However, I don't know how we could tell in the console that Raft consensus is being used.

SatpalSandhu61:
Fixes #698 
Fix issue where geth console displays incorrect timestamp under Raft, due to Raft holding block timestamp in nanoseconds, instead of seconds.


jbhurat:
Hi @SatpalSandhu61, can you please run `goimport` on the modified go source file

SatpalSandhu61:
@jbhurat - ok done. I didn't realise Intellij was using it's own rules for import formatting, rather than the gofmt rules. I've installed the goformats plugin now.

jbhurat (dismissed):

console/console.go
```diff
@@ -294,6 +306,33 @@ func (c *Console) Welcome() {
...
+	if err != nil {
+		c.jsre.Run(`
+			console.log("Warning: unable to retrieve block information, so displayed timestamp may be incorrect");
+		`)
+		return false
+	}
+
+	//If timestamp is greater than max possible value (in seconds) for Unix time, then assume it's nanoseconds.
+	//If we fail to parse the timestamp then just assume it is in seconds.
+	timestamp, err := strconv.ParseInt(lastBlock.Timestamp, 0, 64)
```
jpmsam: @SatpalSandhu61 is it possible to check here if Raft is configured ?

fixanoid (dismissed):

console/console.go
SatpalSandhu61: See updated PR - I now use a new API method to determine whether Raft is configured.

trung:
There's an existing `admin_nodeInfo` API. Can we put additional info in here rather than creating a new one?

SatpalSandhu61:
> There's an existing `admin_nodeInfo` API. Can we put additional info in here rather than creating a new one?

That's a good point. I'll take a look.

SatpalSandhu61:
Closing PR, will raise another one with changes.

PR #714, opened 2019-05-13
Smart contract based permission model

namtruong:
Enabling smart contract based permission model for managing organisations, nodes, roles and accounts in the network


vsmk98:
PR 715 covers this and hence closing the PR

PR #715, opened 2019-05-13
Smart contract based permission model

namtruong:
Smart contract based permission model which caters to enterprise level needs to manage organisations, sub-organisations, roles, account level access controls and nodes in the network. This feature allows group of network admin accounts to propose and approve new organisations into the network with majority approval. Once the new organisation is added, the org admin account of the organisation can then manage sub organisations, roles, account accesses and nodes for the organisation.


core/quorum/api.go
```diff
@@ -0,0 +1,804 @@
...
+	ExecSuccess = ExecStatus{true, "Action completed successfully"}
+)
+
+// NewQuorumControlsAPI creates a new QuorumControlsAPI to access quorum services
+func NewQuorumControlsAPI(tp *core.TxPool, am *accounts.Manager) *QuorumControlsAPI {
+	return &QuorumControlsAPI{tp, nil, am, nil, nil, false, nil, nil}
+}
+
+//Init initializes QuorumControlsAPI with eth client, permission contract and org key management control
+func (p *QuorumControlsAPI) Init(ethClnt *ethclient.Client, key *ecdsa.PrivateKey, apiName string, pconfig *types.PermissionConfig, pc *pbind.PermInterface) error {
```
trung: please use a consistent and meaningful receiver.
`func (p *QuorumControlsAPI)` vs `func (s *QuorumControlsAPI)`

Suggestion: `function (q *QuorumControlsAPI)`

core/quorum/api.go
```diff
@@ -0,0 +1,804 @@
...
+		// check if node already there
+		if s.checkNodeExists(url, enodeDet.EnodeID()) {
+			return ErrNodePresent, errors.New("duplicate node")
+		}
+	}
+	return ExecSuccess, nil
+}
+
+// executePermAction helps to execute an action in permission contract
+func (s *QuorumControlsAPI) executePermAction(action PermAction, args txArgs) ExecStatus {
```
trung: This function is super long. Maybe command pattern would help here

core/tx_pool.go
```diff
@@ -220,11 +223,12 @@ type TxPool struct {
 	wg sync.WaitGroup // for shutdown sync
 
 	homestead bool
+	EnodeId   string
 }
 
 // NewTxPool creates a new transaction pool to gather, sort and filter inbound
 // transactions from the network.
-func NewTxPool(config TxPoolConfig, chainconfig *params.ChainConfig, chain blockChain) *TxPool {
+func NewTxPool(config TxPoolConfig, chainconfig *params.ChainConfig, chain blockChain, id string) *TxPool {
```
trung: Is there a better way to provide the id without changing the method signature. That would bring worry-free when merging from upstream.

Suggestion: how about move it to `TxPoolConfig`?

params/protocol_params.go
```diff
@@ -79,7 +79,8 @@ const (
 	MemoryGas        uint64 = 3     // Times the address of the (highest referenced byte in memory + 1). NOTE: referencing happens on read, write and in instructions such as RETURN and CALL.
 	TxDataNonZeroGas uint64 = 68    // Per byte of data attached to a transaction that is not equal to zero. NOTE: Not payable on data of calls between transactions.
 
-	MaxCodeSize = 24576 // Maximum bytecode to permit for a contract
+	OriginalMaxCodeSize = 24576 // Maximum bytecode to permit for a contract
+	MaxCodeSize         = 32768 // Maximum bytecode to permit for a contract
```
trung: to avoid clashing with upstream, i suggest to keep the `MaxCodeSize` and have our own `QuorumMaxCodeSize`

permission/permission.go
```diff
@@ -0,0 +1,759 @@
...
+	Accounts      []string `json:"accounts"` //initial list of account that need full access
+	SubOrgBreadth string   `json:"subOrgBreadth"`
+	SubOrgDepth   string   `json:"subOrgDepth"`
+}
+
+type PermissionCtrl struct {
+	node             *node.Node
+	ethClnt          *ethclient.Client
+	eth              *eth.Ethereum
+	isRaft           bool
```
trung: Is there a better way to remove this flag? Ideally we want to be consensus agnostic

permission/permission.go
```diff
@@ -0,0 +1,759 @@
...
+		return
+	}
+	// Load the nodes from the config file
+	blob, err := ioutil.ReadFile(path)
+	if err != nil {
+		log.Error("updatePermissionedNodes: Failed to access permissioned-nodes.json", "err", err)
+		return
+	}
+
+	nodelist := []string{}
```
trung: consider using `var nodelist []string` instead of literal

permission/permission.go
```diff
@@ -0,0 +1,759 @@
...
+			p.updatePermissionedNodes(evtNodeBlacklisted.EnodeId, NodeDelete)
+			p.updateDisallowedNodes(evtNodeBlacklisted.EnodeId)
+			types.NodeInfoMap.UpsertNode(evtNodeBlacklisted.OrgId, evtNodeBlacklisted.EnodeId, types.NodeBlackListed)
+		}
+	}
+}
+
+// updates node information in the permissioned-nodes.json file based on node
+// management activities in smart contract
+func (p *PermissionCtrl) updatePermissionedNodes(enodeId string, operation NodeOperation) {
```
trung, line 501: there might be some optimization can be done in this method:
- if there's no change, no need to write to the file
- use json streaming APIs: `json.NewEncoder`, `json.NewDecoder` rather storing data in the memory

What would be the expected behaviour when node is removed from network but writing to `permissioned-nodes.json` fails?

permission/permission.go
```diff
@@ -0,0 +1,759 @@
...
+			break
+		}
+	}
+	if operation == NodeAdd {
+		if !recExists {
+			nodelist = append(nodelist, enodeId)
+		}
+	} else {
+		if recExists {
+			nodelist = append(nodelist[:index], nodelist[index+1:]...)
```
trung: potential slice index out of bound when node being removed is the last node in the list

permission/permission.go
```diff
@@ -0,0 +1,759 @@
...
+
+		case evtStatusChanged = <-chStatusChanged:
+			ac := types.AcctInfoMap.GetAccount(evtStatusChanged.Address)
+			types.AcctInfoMap.UpsertAccount(evtStatusChanged.OrgId, ac.RoleId, evtStatusChanged.Address, ac.IsOrgAdmin, types.AcctStatus(int(evtStatusChanged.Status.Uint64())))
+		}
+	}
+}
+
+// Disconnect the node from the network
+func (p *PermissionCtrl) disconnectNode(enodeId string) {
```
trung, line 585: error is swallowed
better to capture and return an error when disconnecting node

permission/permission.go
```diff
@@ -0,0 +1,759 @@
...
+// populates the account access details from contract into cache
+func (p *PermissionCtrl) populateAccountsFromContract(auth *bind.TransactOpts) {
+	//populate accounts
+	permAcctSession := &pbind.AcctManagerSession{
+		Contract: p.permAcct,
+		CallOpts: bind.CallOpts{
+			Pending: true,
+		},
+	}
+	if numberOfRoles, err := permAcctSession.GetNumberOfAccounts(); err == nil {
```
trung, line 699: error is swallowed

permission/permission.go
```diff
@@ -0,0 +1,759 @@
...
+// populates the role details from contract into cache
+func (p *PermissionCtrl) populateRolesFromContract(auth *bind.TransactOpts) {
+	//populate roles
+	permRoleSession := &pbind.RoleManagerSession{
+		Contract: p.permRole,
+		CallOpts: bind.CallOpts{
+			Pending: true,
+		},
+	}
+	if numberOfRoles, err := permRoleSession.GetNumberOfRoles(); err == nil {
```
trung, line 721: error is swallowed

permission/permission.go
```diff
@@ -0,0 +1,759 @@
...
+// populates the node details from contract into cache
+func (p *PermissionCtrl) populateNodesFromContract(auth *bind.TransactOpts) {
+	//populate nodes
+	permNodeSession := &pbind.NodeManagerSession{
+		Contract: p.permNode,
+		CallOpts: bind.CallOpts{
+			Pending: true,
+		},
+	}
+	if numberOfNodes, err := permNodeSession.GetNumberOfNodes(); err == nil {
```
trung, line 744: error is swallowed

permission/permission.go
```diff
@@ -0,0 +1,759 @@
...
+// populates the org details from contract into cache
+func (p *PermissionCtrl) populateOrgsFromContract(auth *bind.TransactOpts) {
+	//populate orgs
+	permOrgSession := &pbind.OrgManagerSession{
+		Contract: p.permOrg,
+		CallOpts: bind.CallOpts{
+			Pending: true,
+		},
+	}
+	if numberOfOrgs, err := permOrgSession.GetNumberOfOrgs(); err == nil {
```
trung, line 766: error is swallowed

trung:
Please consider to:
- [x] make sure to run `goimports` 
- [x] write tests for changes introduced
- [ ] avoid commit auto generated files, let them be generated as a part of `make` process.  Would need to add the check for version of solc being used

docs/Permissioning/Permissioning apis.md
```diff
@@ -0,0 +1,427 @@
...
+The table below indicates the numeric value for various organization statuses.
+
+| OrgStatus                 |           Value |
+| :-----------------------: | :-------------: |
+| NotInList                 |               0 |
+| Proposed                  |               1 |
+| Approved                  |               2 |
+| PendingSuspension         |               3 |
+| Suspended                 |               4 |
+| AwaitingSuspensionRevoke  |               5 |
```
trung: this list doesn't match with values defined in `core/types/permissions_cache.go`

docs/Permissioning/Permissioning apis.md
```diff
@@ -0,0 +1,427 @@
...
+
+#### Node Status types
+The table below indicates the numeric value for various node statuses.
+
+| NodeStatus                |           Value |
+| :-----------------------: | :-------------: |
+| NotInList                 |               0 |
+| PendingApproval           |               1 |
+| Deactivated               |               2 |
+| Blacklisted               |               3 |
```
trung: this list doesn't match with values defined in `core/types/permissions_cache.go`

core/types/permissions_cache.go
```diff
@@ -0,0 +1,410 @@
...
+		if ent.Url == url {
+			v, _ := n.c.Get(ent)
+			return v.(*NodeInfo)
+		}
+	}
+	return nil
+}
+
+func (o *NodeCache) GetNodeList() []NodeInfo {
+	var olist []NodeInfo
```
trung: if we know the length of `[]NodeInfo` in advanced, it'd be better to initialize the exact slice length

core/quorum/api.go
vsmk98: Changed to function (q *QuorumControlsAPI)

params/protocol_params.go
vsmk98: Agree. changed accordingly

permission/permission.go
vsmk98: This flag is needed as whenever the node is deactivated or blacklisted, we have to disconnect the node from current network. For Raft we need to disconnect from Raft cluster and P2P.

permission/permission.go
vsmk98: changes done.

permission/permission.go
vsmk98: Added a condition to check if file is changing or not. If not it will return. Assuming that the write failed for one of the nodes in the network, when the disconnected node is brought back:
- if it is in raft consensus - it will not be able to join the raft cluster and hence no issues
- in case of IBFT, the entire network other than this one node where write failed will reject the P2P connection and as such the node should not be able to transact. 
I have not done the changes NewEncoder/NewDecoder as the file size even in large networks will be relatively small.

permission/permission.go
vsmk98: I checked specifically for the boundary condition and deleted which worked fine. go is handling the scenario where the lower bound is equal to the length of the array.

permission/permission.go
vsmk98: have logged the error. This is running in a separate go routing and hence preferred logging of the error.

permission/permission.go
vsmk98: handled error

permission/permission.go
vsmk98: handled erorr

permission/permission.go
vsmk98: handled error

docs/Permissioning/Permissioning apis.md
vsmk98: the only missing one is notinlist which is the default value. In permissions_cache.go we are starting from 1.

docs/Permissioning/Permissioning apis.md
vsmk98: updated documentation.

core/types/permissions_cache.go
vsmk98: handled for all the lists

core/quorum/api.go
vsmk98: rafactored code. Though have not used command patter. Want to test bit more before I push the changes.

vsmk98: changes pushed

cmd/geth/main.go
```diff
@@ -341,6 +345,10 @@ func startNode(ctx *cli.Context, stack *node.Node) {
 			}
 		}
 	}()
+
+	//START - QUORUM permission service
+	go startQuorumPermissionService(ctx, stack)
```
trung: It'd be better to capture the error and halt the node if permission service is not enabled
```suggestion
go func() {
  if permission enabled {
     if err := startQuorumPermissionService(..); err != nil {
        utils.Fatalf(...)
     }
   }
}()
```

vsmk98: done

cmd/geth/main.go
```diff
@@ -379,46 +386,41 @@ func startNode(ctx *cli.Context, stack *node.Node) {
...
-			log.Error("loading of permission-config.json failed", "error", err)
-			return
-		}
-	} else {
-		// permissions not enabled hence none of the services will be available.
-		return
+
+	if permissionConfig, err = permission.ParsePermissionConifg(dataDir); err != nil {
+		log.Error("loading of permission-config.json failed", "error", err)
+		return nil
```
trung:
```suggestion
return err
```

vsmk98: If permissions-config.json is not present, this means that new permissions model is not implemented and hence will continue with the earlier checks of permissioned-nodes.json. Hence returning nil logging the error. If error is returned it will bring down the node.

permission/contract/PermissionsImplementation.sol
```diff
@@ -0,0 +1,392 @@
...
+    onlyProxy
+    networkBootStatus(false)
+    {
+        adminOrg = _nwAdminOrg;
+        adminRole = _nwAdminRole;
+        orgAdminRole = _oAdminRole;
+    }
+
+    // called at the time network initialization to link all the contracts and set defaults
+    function init(address _orgManager, address _rolesManager, address _acctManager, address _voterManager, address _nodeManager, uint _breadth, uint _depth) external
```
trung: After deploying new version of `PermissionsImplementation.sol`, this `init(...)` method has to be invoked to bind the storage contracts to new implementation.

Would it be better to incorporate those params in the constructor instead?

core/quorum/api.go
```diff
@@ -0,0 +1,895 @@
...
+	tmKey      string
+	txa        ethapi.SendTxArgs
+}
+
+type PendingOpInfo struct {
+	PendingKey string `json:"pendingKey"`
+	PendingOp  string `json:"pendingOp"`
+}
+
+type ExecStatus struct {
```
trung: `ExecStatus` breaks the [JSON RPC response specification](https://www.jsonrpc.org/specification#response_object) as response is always a success.

Suggestion: return the real error and let the framework handle the JSON RPC error response

core/quorum/api.go
```diff
@@ -0,0 +1,895 @@
...
+	}
+	// validation status change is with in allowed set
+	if execStatus, er := q.valAccountStatusChange(args.orgId, args.acctId, int64(args.action)); er != nil {
+		return execStatus
+	}
+	return ExecSuccess
+}
+
+// executePermAction helps to execute an action in permission contract
+func (q *QuorumControlsAPI) executePermAction(action PermAction, args txArgs) ExecStatus {
```
trung: I think the motivation here is to reuse line 737-750. However, the switch case with action constants seem to be overkilled. Responsibilities can be encapsulated into original methods.

permission/permission.go
```diff
@@ -0,0 +1,782 @@
...
+			types.AcctInfoMap.UpsertAccount(evtStatusChanged.OrgId, ac.RoleId, evtStatusChanged.Address, ac.IsOrgAdmin, types.AcctStatus(int(evtStatusChanged.Status.Uint64())))
+		}
+	}
+}
+
+// Disconnect the node from the network
+func (p *PermissionCtrl) disconnectNode(enodeId string) {
+	if p.isRaft {
+		var raftService *raft.RaftService
+		if err := p.node.Service(&raftService); err == nil {
```
trung, line 588: can this condition be used to replace `isRaft` explicit field in the `PermissionCtrl` struct?

permission/permission.go
```diff
@@ -0,0 +1,782 @@
...
+
+type PermissionCtrl struct {
+	node             *node.Node
+	ethClnt          *ethclient.Client
+	eth              *eth.Ethereum
+	isRaft           bool
+	permissionedMode bool
+	key              *ecdsa.PrivateKey
+	dataDir          string
+	permUpgr         *pbind.PermUpgr
```
trung: is this field being used anywhere?

trung (changes requested):
Please consider the inline comments and also add unit tests for the new changes.

permission/permission.go
vsmk98: All these are used. permUpgr is the starting point and checking that its deployed properly before enabling permissions

permission/permission.go
vsmk98: have remove isRaft condition from this piece of code. actually chainConfig is available `p.eth.chainConfig.Istanbul` or `p.eth.ChainConfig.Clique`. Using these two in place of explicit isRaft

core/quorum/api.go
vsmk98: Yes agree. the motivation has been to reuse those lines but feel thats adding much more lines to overall code. agree with your comments. Changed and code pushed.

jpmsam:
Please provide more details on the PR and the features it enables in Quorum.

core/quorum/api.go
vsmk98: code changes done to return (response, error) as a part of each api calls. Documentation updated accordingly.

permission/contract/PermissionsImplementation.sol
vsmk98: Agree. Have moved all storage contract addresses to constructor.

cmd/utils/cmd.go
```diff
@@ -63,7 +65,21 @@ func Fatalf(format string, args ...interface{}) {
 	os.Exit(1)
 }
 
+func IsPermissionEnabled(ctx *cli.Context) bool {
+	if ctx.GlobalBool(EnableNodePermissionFlag.Name) {
+		fileName := "permission-config.json"
+		fullPath := filepath.Join(ctx.GlobalString(DataDirFlag.Name), fileName)
```
trung: CLI flag `--datadir` is indeed optional and has a default value. Further, `datadir` can be also provided via TOML config file using `--config` flag. 

Please pass `*node.Config` to this method and obtain `datadir` from it.

core/types/permissions_cache.go
```diff
@@ -0,0 +1,406 @@
...
+		return true
+	}
+	ac := AcctInfoMap.GetAccount(from)
+	if ac == nil {
+		return true
+	}
+	ultimateParent := OrgInfoMap.GetOrg(ac.OrgId).UltimateParent
+	// scan through the node list and validate
+	for _, n := range NodeInfoMap.GetNodeList() {
+		if OrgInfoMap.GetOrg(n.OrgId).UltimateParent == ultimateParent && strings.Contains(n.Url, enodeId) {
```
trung: Defensively, I would use `enode.ParseV4(n.Url)` then compare the enodeId

Note: there's a type `enode.ID` to represent enodeId through out the code. Even though `string` doesn't create any issue, as far as consistency is concerned, `enode.ID` is preferred.

core/tx_pool.go
trung: Based on the usage of `EnodeId` inside `TxPool`, only  `api_backend.go:func (b *EthAPIBackend) SendTx(..)` uses it to perform validation. 

IMO, `enodeId` should be referenced rather than constructed (https://github.com/jpmorganchase/quorum/blob/e9a05f63e7aee9cb82b7f34042b9fd171bd05c92/eth/backend.go#L200) and passed around 

`p2p.Server` is the right struct to reference the `enodeId` via its `LocalNode().ID()`. However, there is no reference to `p2p.Server` from `EthAPIBackend`. I did a quick check upstream and found that 
https://github.com/ethereum/go-ethereum/blob/v1.9.0/eth/backend.go#L66-L72

Adding `p2p.Server` to `Ethereum` which then can be referenced from `EthAPIBackend` to fulfill the validation needs

core/tx_pool.go
```diff
@@ -630,6 +646,14 @@ func (pool *TxPool) validateTx(tx *types.Transaction, local bool) error {
 	if tx.Gas() < intrGas {
 		return ErrIntrinsicGas
 	}
+
+	// Check if the sender account is authorized to perform the transaction
+	if isQuorum {
+		if err := checkAccount(from, tx.To()); err != nil {
+			return ErrUnAuthorizedAccount
```
trung: User experience: It'd be good to append the message from `err` object to the final error.

cmd/utils/cmd.go
vsmk98: `IsPermissionEnabled` is called from two places - `cmd.go` and `main.go`. Hence changed the function signature to pass the two values it is looking for, datadir and permissionEnabled flag.

core/types/permissions_cache.go
vsmk98: Changed to enode.ID validation for this function alone. For rest of the places I have left it as string only as this is cache of the data in contract and in contract we are storing as string.

core/tx_pool.go
vsmk98: changed.

core/types/permissions_cache.go
```diff
@@ -0,0 +1,406 @@
...
+type RoleKey struct {
+	OrgId  string
+	RoleId string
+}
+
+type AccountKey struct {
+	AcctId common.Address
+}
+
+type OrgCache struct {
```
trung, line 131: `lru.Cache` is already thread-safe. Why do we need `mux` here?

permission/permission.go
```diff
@@ -0,0 +1,863 @@
...
+	permConfig *types.PermissionConfig
+	orgChan    chan struct{}
+	nodeChan   chan struct{}
+	roleChan   chan struct{}
+	acctChan   chan struct{}
+	mux        sync.Mutex
+}
+
+// converts local permissions data to global permissions config
+func populateConfig(config PermissionLocalConfig) types.PermissionConfig {
```
trung: Please consider to implement `UnmarshalJSON()` to customize the transformation

permission/permission.go
```diff
@@ -0,0 +1,863 @@
...
+func waitForSync(e *eth.Ethereum) {
+	for !types.GetSyncStatus() {
+		time.Sleep(10 * time.Millisecond)
+	}
+	for e.Downloader().Synchronising() {
+		time.Sleep(10 * time.Millisecond)
+	}
+}
+
+func StartPermissionService(stack *node.Node) {
```
trung: `utils.Fatal()` is best used in the front layer not in the service layer. I would make this method return an error and do `utils.Fatal()` at `cmd/geth/main.go`

permission/permission.go
```diff
@@ -0,0 +1,863 @@
...
+}
+
+// Creates the controls structure for permissions
+func NewQuorumPermissionCtrl(stack *node.Node, pconfig *types.PermissionConfig) (*PermissionCtrl, error) {
+	// Create a new ethclient to for interfacing with the contract
+	return &PermissionCtrl{stack, nil, nil, stack.GetNodeKey(), stack.DataDir(), nil, nil, nil, nil, nil, nil, pconfig, make(chan struct{}), make(chan struct{}), make(chan struct{}), make(chan struct{}), sync.Mutex{}}, nil
+}
+
+func (p *PermissionCtrl) InitializeService() error {
+	clnt, eth, err := CreateEthClient(p.node)
```
trung: `eth` is collided with package `eth`

permission/permission.go
```diff
@@ -0,0 +1,863 @@
...
+	if err != nil {
+		log.Error("creating eth client failed")
+		return err
+	}
+	waitForSync(eth)
+	if err != nil {
```
trung, line 154: `err` is always nil here

permission/permission.go
```diff
@@ -0,0 +1,863 @@
...
+
+	// monitor org  level node management events
+	go p.manageNodePermissions()
+
+	// monitor org level role management events
+	go p.manageRolePermissions()
+
+	// monitor org level account management events
+	go p.manageAccountPermissions()
+	log.Info("permission service started")
```
trung: need to wait here until all the watchers are up and no error

permission/permission.go
```diff
@@ -0,0 +1,863 @@
...
+	opts := &bind.WatchOpts{}
+	var blockNumber uint64 = 1
+	opts.Start = &blockNumber
+
+	if _, err := p.permOrg.OrgManagerFilterer.WatchOrgPendingApproval(opts, chPendingApproval); err != nil {
+		log.Info("Failed WatchNodePendingApproval: %v", err)
+	}
+
+	if _, err := p.permOrg.OrgManagerFilterer.WatchOrgApproved(opts, chOrgApproved); err != nil {
+		log.Info("Failed WatchNodePendingApproval: %v", err)
```
trung: log message typo

see the below too

permission/permission.go
```diff
@@ -0,0 +1,863 @@
...
+	var evtAccessModified *pbind.AcctManagerAccountAccessModified
+	var evtAccessRevoked *pbind.AcctManagerAccountAccessRevoked
+	var evtStatusChanged *pbind.AcctManagerAccountStatusChanged
+
+	opts := &bind.WatchOpts{}
+	var blockNumber uint64 = 1
+	opts.Start = &blockNumber
+
+	if _, err := p.permAcct.AcctManagerFilterer.WatchAccountAccessModified(opts, chAccessModified); err != nil {
+		log.Info("Failed NewNodeProposed", "error", err)
```
trung: log message typo

please see the below too

permission/permission.go
```diff
@@ -0,0 +1,863 @@
...
+	_, err := permissionsSession.UpdateNetworkBootStatus()
+	if err != nil {
+		log.Warn("Failed to udpate network boot status ", "err", err)
+		return err
+	}
+	return nil
+}
+
+// monitors role management related events and updated cache
+func (p *PermissionCtrl) manageRolePermissions() {
```
trung: use consistent receiver `p` or `s` but not both for `PermissionCtrl`

trung (changes requested):

permission/permission.go
trung: have a look at:

https://github.com/jpmorganchase/quorum/blob/1f0da48be656997cd8b446d40bee8702df92af27/common/types.go#L256-L259

https://github.com/jpmorganchase/quorum/blob/1f0da48be656997cd8b446d40bee8702df92af27/common/types_test.go#L93

trung (dismissed):
LGTM

cbaus:
Could you kindly let us know what's the ETA for this feature?

jpmsam (approved):

kikoncuo:
Any documentation on how to use this?

SatpalSandhu61:
> Any documentation on how to use this?

Docs are here:  http://docs.goquorum.com/en/latest/Permissioning/Overview/

PR #727, opened 2019-05-29
set gcmode=archive always for Raft

amalrajmani:
With gcmode=full(feature introduced from quorum 2.2.0 onwards) Raft is not able to sync up when a node is restarted after crash or force kill.
This fix will set gcmode=archive always for Raft as per quorum.2.1.1


jpmsam (approved):

jimthematrix:
@jpmsam @prd-fox thanks for making this fix, exactly what we need which we were looking into but then saw this fix. I have a question though, if I have an existing Raft chain and all of the nodes are running in gcmode=full with Quorum release 2.2.3, and I upgrade the chain nodes to this release, will the chain continue to function? what happens if a gcmode=archive node finds that it's missing states older than the 128-block window?

PR #739, opened 2019-06-12
non-party node must refund gas for private transactions

trung:
### Summary

Gas pool subtracts transaction gas limit initially prior to processing the transaction.
For private transactions, only intrinsic gas are used. 
The remaining (transaction gas limit - intrinsic gas) is refunded to the gas pool. 
This logic is supposed to be applied for both party and non-party nodes.

This PR is to make sure non-party node following the above logic.

### Root cause

Under load, non-party node may see bad blocks due to `gas limit reached`. This happens when intrinsic gas is reasonably smaller than transaction gas limit. In this case, party node (as a miner) may include lots of transactions to fill up the block gas limit. However, when non-party node inserts the block into its chain, the gas pool gets exhausted quickly due to only subtracting transaction gas limit and not refunding the remaining gas. This causes `bad block` with `gas limit reached` error being reported.

### DIFF
--- a/core/state_transition.go
+++ b/core/state_transition.go
@@ -188,6 +188,12 @@ func (st *StateTransition) preCheck() error {
 // TransitionDb will transition the state by applying the current message and
 // returning the result including the used gas. It returns an error if failed.
 // An error indicates a consensus issue.
+//
+// Quorum:
+// 1. Intrinsic gas is calculated based on the encrypted payload hash
+//    and NOT the actual private payload
+// 2. For private transactions, we only deduct intrinsic gas from the gas pool
+//    regardless the current node is party to the transaction or not
 func (st *StateTransition) TransitionDb() (ret []byte, usedGas uint64, failed bool, err error) {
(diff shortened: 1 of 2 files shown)


zzy96:
Hi Trung, I am thinking if `st.state.AddBalance(st.evm.Coinbase, new(big.Int).Mul(new(big.Int).SetUint64(st.gasUsed()), st.gasPrice))` should be added also after the refund. It will not affect current quorum implementation since gasPrice is always zero, but if we reenable gasPrice at some time, the used gas should be recycled the same way for all transactions.

core/state_transition.go
```diff
@@ -254,6 +254,7 @@ func (st *StateTransition) TransitionDb() (ret []byte, usedGas uint64, failed bo
 		}
 		//if input is empty for the smart contract call, return
```
SatpalSandhu61, line 261: Would be useful if you could add a comment that this condition will occur when the node is not a participant to a private transaction.

jpmsam (approved):

PR #750, opened 2019-06-21
Bugfix/geth console displays incorrect block timestamp under Raft

SatpalSandhu61:
… due to Raft holding block timestamp in nanoseconds, instead of seconds.

### DIFF
--- a/console/console.go
+++ b/console/console.go
@@ -17,7 +17,7 @@
 package console
 
 import (
-	"encoding/json"
+	"context"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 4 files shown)


console/console.go
```diff
@@ -294,6 +311,45 @@ func (c *Console) Welcome() {
 	fmt.Fprintln(c.printer)
 }
 
+// Get the consensus mechanism that is in use
+func (c *Console) getConsensus() string {
+
```
trung, line 316: please consider this simpler approach:
```go
	var nodeInfo struct {
		Protocols struct {
			Eth struct { // only partial of eth/handler.go#NodeInfo
				Consensus string
			}
			Istanbul struct { // a bit different from others
				Consensus string
			}
		}
	}
	if err := c.client.CallContext(context.Background(), &nodeInfo, "admin_nodeInfo"); err != nil {
		_, _ = fmt.Fprintf(c.printer, "WARNING: call to admin.getNodeInfo() failed, unable to determine consensus mechanism\n")
		return "unknown"
	}
	if nodeInfo.Protocols.Istanbul.Consensus != "" {
		return nodeInfo.Protocols.Istanbul.Consensus
	}
	return nodeInfo.Protocols.Eth.Consensus
```

eth/handler.go
```diff
@@ -831,17 +831,32 @@ type NodeInfo struct {
...
+	if pm.blockchain.Config().Clique != nil {
+		consensus = "clique"
+	} else if pm.blockchain.Config().Istanbul != nil {
+		consensus = "istanbul"
+	} else if pm.blockchain.Config().Ethash != nil {
+		consensus = "ethash"
+	} else if pm.raftMode {
+		consensus = "raft"
+	}
+
```
trung: i think it would be more robust to assert on the engine type rather than config
```go
	if pm.raftMode { // raft does not use consensus interface
		consensusAlgo = "raft"
	} else {
		switch pm.engine.(type) {
		case consensus.Istanbul:
			consensusAlgo = "istanbul"
		case *clique.Clique:
			consensusAlgo = "clique"
		case *ethash.Ethash:
			consensusAlgo = "ethash"
		default:
			consensusAlgo = "unknown"
		}
	}
```

And if this could be extracted to a separate method, we can easily write unit tests for it

console/console.go
SatpalSandhu61: Thanks, that does look better.

PR #758, opened 2019-07-01
fix: fix builtin shadowing

0xmountaintop:
As `copy`, `len`, `int`,`panic`, `rune` and `print` are pre-declared identifiers in GoLang, we should avoid using them to name variables, in order to avoid shadowing and confusions.


fixanoid:
Hi @HAOYUatHZ, all the changes in this commit are part of upstream geth, Could you please open the pull on their repo instead? As a user of go-ethereum, we will pull the changes in as they make it through official client, otherwise, this creates a unneeded upstream merge issue for us. Thanks.

0xmountaintop:
Thx for the explanation.
Will do.

PR #760, opened 2019-07-01
p2p/simulations: Enable access to MsgEvents with execadapter

0xmountaintop:
`EnableMsgEvents` options for p2p config are currently hardcoded for `ExecAdapter`. This is change is required to be able to enable it using the `adapters.NodeConfig` directive.


fixanoid:
@HAOYUatHZ Thank you, as with other pulls, please submit directly to the upstream unless its specific to Quorum: permissions, consensus or privacy components.

PR #761, opened 2019-07-01
core/types: document RawSignatureValues

0xmountaintop:


fixanoid:
@HAOYUatHZ Thank you, as with other pulls, please submit directly to the upstream unless its specific to Quorum: permissions, consensus or privacy components.

Issue #672, opened 2019-04-02
Handle missing eip155Block configuration


jbhurat:
As a follow up to PR https://github.com/jpmorganchase/quorum/pull/667 which handles EIP 155 signed transactions when `eip155Block` configuration is defined. In case, the configuration is missing the EIP155 signed transaction and any non EIP155 signed transaction gets stuck. Instead we can add a validation and reject the EIP155 signed transaction if `eip155Block` is not configured

jbhurat:
Closing this ticket, as `eip155` signed transactions should not be submitted if `eip155Block` configuration is missing from genesis config

PR #766, opened 2019-07-02
Creating a signer based on chain config and block number

jbhurat:
Signer in tx pool is created using chain config and current block, instead of defaulting to eip155 signer. 

Solves #672


fixanoid (approved):

jbhurat:
Closing this PR, as it doesn't handle the scenario where, if existing network is upgraded before `eip155Block` is reached and eip155 signed transaction is submitted

PR #767, opened 2019-07-02
Adding QuorumPrivateTxSigner for signing and recovering private txs.

libby:
* Add QuorumPrivateTxSigner to encapsulate private signer logic.
* Use return QuorumPrivateTxSigner to recover Sender in EIP155. minimize changes vs upstream transaction_signing.go.
* Since SetPrivate( is used to mark txs as private and not to calc the correct v comment it as such).
* `internal/ethapi/api.go` `submitTransaction(` should accept a signed transaction, make sure the tx is properly signed before calling this function.
* Remove `isQuorum` from `wallet.SignTx(account, tx, chainID, isQuorum)`, when passing to `SignTx(` the tx should already be marked as private, and
  the wallet can check if the transaction is private, and make a decision on whether or not it can support private txs, public tx should be supported.
* Update signing api for `faucet.go` and `mobile/accounts.go`.
* Add test for `QuorumPrivateTxSigner`.

### DIFF
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -392,7 +392,14 @@ func (s *PrivateAccountAPI) SendTransaction(ctx context.Context, args SendTxArgs
 		defer s.nonceLock.UnlockAddr(args.From)
 	}
 
-	isPrivate := args.PrivateFor != nil
+	// Set some sanity defaults and terminate on failure
+	if err := args.setDefaults(ctx, s.b); err != nil {
+		return common.Hash{}, err
+	}
+	// Assemble the transaction and sign with the wallet
+	tx := args.toTransaction()
+
+	isPrivate := args.IsPrivate()
 
 	if isPrivate {
 		data := []byte(*args.Data)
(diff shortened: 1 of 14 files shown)


internal/ethapi/api.go
```diff
@@ -1432,7 +1442,7 @@ func (s *PublicTransactionPoolAPI) SendRawPrivateTransaction(ctx context.Context
 	}
 
 	txHash := []byte(tx.Data())
-	isPrivate := args.PrivateFor != nil
+	isPrivate := (args.PrivateFor != nil) && tx.IsPrivate()
```
jbhurat, line 1445: In addition to checks for PrivateFor args is the check for `tx.IsPrivate()` required

libby: The reason for the additional `tx.IsPrivate()` check, is for the case where a transaction is submitted with `PrivateFor` set, but where the transaction is not signed with the correct private signer. In this case we should reject the tx right away.

internal/ethapi/api.go
```diff
@@ -407,25 +414,18 @@ func (s *PrivateAccountAPI) SendTransaction(ctx context.Context, args SendTxArgs
 		// zekun: HACK
 		d := hexutil.Bytes(data)
 		args.Data = &d
+		tx = args.toTransaction()
```
chris-j-h: Is this line required given that the tx has already been created in line 400?

jbhurat (dismissed):

jbhurat (approved):
LGTM

jpmsam (approved):

PR #769, opened 2019-07-04
Istanbul snapshot display fix

zzy96:
Make the empty votes display as an empty array not null when reset by epoch.


zzy96:
Closing this PR for now

PR #771, opened 2019-07-08
Add private tx client

withzoo:
add send private tx client for developer

Signed-off-by: Zeyu Zhu <zhuzeyu0409@gmail.com>


zzy96:
Private tx client is supported as part of #633

jpmsam:
There is an active pr #633 to enable this.

PR #777, opened 2019-07-15
Restructure and reorder Getting Started docs

chris-j-h:
This makes several changes to the `Getting Started` section of the documentation, with the intention of making it clearer for new users which pages they should be working through first and improving the general flow/readability.

* Rename `Getting Started From Scratch` to `Creating a Network From Scratch` to make it clearer that in most cases this should not be the first thing a new user works through
* Make it clearer that `Quorum Examples` is the recommended place to start when first working with Quorum by placing it above the `Creating a Network From Scratch` page in the sidebar
* Create new `Getting Started Overview `page to summarise the different ways users can get a network running and start using Quorum
* Move Quorum/Tessera/Constellation installation instructions to a separate page
* Break up quorum-examples docs so an overview of 7nodes and 5nodesRTGS, setting up 7nodes, and running 7nodes are segregated


fixanoid (approved):
❤️ 🚀

jpmsam (approved):

PR #797, opened 2019-08-08
Allow to use hostname in enode (static-nodes.json)

### LINKED ISSUE #147: Feat: `static-nodes.json` doesn't support DNS/hostnames
It would be **awesome** if we could use hostnames/DNS in our `static-nodes.json` file.  At the moment the `ParseNode` function does not support this:

https://github.com/jpmorganchase/quorum/blob/436fc8d76a4871d67a61dc86c1a635e20594a0e6/p2p/discover/node.go#L133

I would have a crack myself but I'm clueless when it comes to golang :)

It would greatly simplify our runtime configuration as we would be able to ditch all the static IP requirements here:
https://github.com/CBAInnovationLab/quorum-examples/blob/master/examples/7nodes/docker-compose.yml

hayorov:
Close https://github.com/jpmorganchase/quorum/issues/147 

## Feature
Add ability to use `hostname` in enode  e.g. enode://...@hostname:30303?discport=0&raftport=50400
If host is not an IP uses `net.lookupIP` and use the first resolved IP.

This is a back port of https://github.com/jpmorganchase/quorum/commit/749acf065caea332e550bf31a918a8dab15f2908 that was introduced in https://github.com/jpmorganchase/quorum/issues/147 and lost after fcfc7b6d452fd9c42014c055dc9da39baf80e226 (merge with geth upstream).

### DIFF
--- a/p2p/discv5/node.go
+++ b/p2p/discv5/node.go
@@ -138,7 +138,7 @@ var incompleteNodeURL = regexp.MustCompile("(?i)^(?:enode://)?([0-9a-f]+)$")
 //
 // For complete nodes, the node ID is encoded in the username portion
 // of the URL, separated from the host by an @ sign. The hostname can
-// only be given as an IP address, DNS domain names are not allowed.
+// be given as an IP address or DNS domain.
 // The port in the host name section is the TCP listening port. If the
 // TCP and UDP (discovery) ports differ, the UDP port is specified as
 // query parameter "discport".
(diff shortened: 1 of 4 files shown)


hayorov:
I will be happy to discuss supporting hostname for the node (enode) in Quorum.
I did not find clear documents of that and don't know what is the quorum official opinion on this topic.
Current fix is just a W/A and restores missed functionality that merged.

@fixanoid

jpmsam:
Hi @hayorov we have an updated design that introduces better support but it requires more changes beyond the IP lookup. You'll see a PR in the next week on this from our team.

hayorov:
@jpmsam It must be a big thing with RAFT package changes. I hope it will be possible to use FQDN in RAFT (eg. raft_addPeer). Thanks, I'm looking forward.

easeev:
Hey @jpmsam!

>You'll see a PR in the next week on this from our team.

Any updates on this?

PR #800, opened 2019-08-13
Add configurable threshold for 'future block' check

### LINKED ISSUE #792: Add configurable threshold for 'block in the future' check
#### System information

Quorum version: 2.2.4

#### Expected behaviour

Nodes should have some tolerance for time differences. If a node is a few milliseconds out of sync then it should not cause "Block in the future" warning messages.

#### Actual behaviour

NTP will only keep nodes synchronised to within a few milliseconds (on a public network, it could be tens of milliseconds). However, the current code has zero tolerance for any time difference in seconds.

_Consider the following scenario:_
- We have 2 nodes A & B, where node A is (say) one millisecond ahead of node B.
- Node A mints a block with timestamp N seconds, a few microseconds after the time ticked has over from N-1.
- Node B receives & validates the block, but the time on this node has not yet ticked over, so it’s local time is still N-1.
- In this case, we’ll get a ‘block in the future’ message, with a time difference of microseconds.

In order to resolve this, we should only generate the warning message if the time delta is above a predefined threshold value.

SatpalSandhu61:
Fixes #792 
Added configurable threshold for 'future block' check, in order to avoid spurious "block in the future" messages for Istanbul and Clique.
Threshold is specified in seconds, using`--allowedfutureblocktime` command line argument.

Note that there is a similar change for ethash in upstream geth (albeit a fixed threshold, not configurable): https://github.com/ethereum/go-ethereum/blob/master/consensus/ethash/consensus.go#L248


cmd/utils/flags.go
```diff
@@ -647,6 +647,11 @@ var (
 		Usage: "Default minimum difference between two consecutive block's timestamps in seconds",
 		Value: eth.DefaultConfig.Istanbul.BlockPeriod,
 	}
+	IstanbulAllowedFutureBlockTimeFlag = cli.Uint64Flag{
+		Name:  "istanbul.allowedfutureblocktime",
+		Usage: "Time threshold allowed when detecting future blocks in seconds",
```
trung: Nitpick: i would use similar comment in upstream: 
> Max time (in seconds) from current time allowed for blocks, before they're considered future blocks

trung (commented):
Some nitpicks in code. Also please run `goimports` for the modified files

trung (dismissed):
LGTM

SatpalSandhu61:
Temporarily closing this PR as I want to make the parameter name generic, not just for Istanbul.

SatpalSandhu61:
PR has been reopened with changes to make it more generic and include Clique.

trung (approved):
LGTM

PR #806, opened 2019-08-16
WIP Hashicorp Vault account management

chris-j-h:
**Overview**
Similar to the [existing Vault integration in Tessera](https://docs.goquorum.com/en/latest/Privacy/Tessera/Tessera%20Services/Keys/Setting%20up%20a%20Hashicorp%20Vault/), adds support for storing and using Quorum account private keys in a Hashicorp Vault instead of in local files.  

New keys can be created using the `geth account new` CLI.  Accounts are configured for use in Quorum through the standard `.toml` config file.

**How Vault is used**
Vault-stored accounts can be in one of three states:

1. **Locked**: Private key component is not retained in memory.  Each time an account is used for signing an API request is made to the Vault to retrieve the corresponding private key.  The key is then zeroed after use.
2. **Timed Unlock**: Private key component is retained in memory for some pre-determined duration.  During this period, signing operations can be carried out without needing to make API requests to the Vault.  Once the duration has passed the key is zeroed and the account becomes locked.
3. **Unlocked**: Private key component is retained in memory indefinitely.  Signing operations can be carried out without needing to make API requests to the Vault.  

**New`personal` APIs**:
* `personal_lockVaultAccount(addr)`: Immediately lock an unlocked Vault-stored account
* `personal_unlockVaultAccount(addr, duration)`: Unlock a Vault-stored account for `duration` secs (default `300`).  `duration=0` indefinitely unlocks the account.

`web3.js` in the `geth` JS console has been extended to add support for these new APIs (this should be considered a temporary solution until support is added to QuorumJS): 
* `personal.lockVaultAccount(addr)`
* `personal.unlockVaultAccount(addr, duration)`

**Docs**
Doc updates are included in this PR.  Do `mkdocs serve` and navigate to `http://localhost:8000/Security/Accounts/Overview/` to view them.


accounts/vault/hashicorp_config_test.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+func TestHashicorpWalletConfig_Validate_NoVaultUrlReturnsError(t *testing.T) {
+	w := getMinimumValidConfig()
+	w.Client.Url = ""
+
+	err := w.Validate()
+
+	if err == nil {
+		t.Error("Wanted error")
```
jbhurat, line 46: It would make sense to assert the `err` that should be returned. Similarly for the other scenarios below

accounts/vault/hashicorp_service.go
```diff
@@ -0,0 +1,690 @@
+package vault
+
+import (
```
jbhurat, line 3: Can you run `goimports` on the updated/new files

chris-j-h:
This is implementation is being abstracted out to a separate plugin and will be raised as a separate PR soon

PR #811, opened 2019-08-28
Issue 802 fix raft hangup when isQuorum false

### LINKED ISSUE #802: Transaction goes to pending state. when "isQuorum" is set to false.
**Genesis File:**

{
	"alloc": {
		"0xWallet-Id-Here": {
		  "balance": "300000000000000000000000000000000000000"
		}
	},

	"config": {
		"homesteadBlock": 0,
		"byzantiumBlock": 123456,
		"constantinopleBlock":0,
		"chainId": 10,
		"eip150Block": 0,
		"eip155Block": 0,
		"eip150Hash": "0x0000000000000000000000000000000000000000000000000000000000000000",
		"eip158Block": 0,
		"isQuorum": false
	},
	"coinbase": "0x0000000000000000000000000000000000000000",
	"difficulty": "0x0",
	"extraData": "0x0000000000000000000000000000000000000000000000000000000000000000",
	"gasLimit": "0xE0000000000",
	"mixhash": "0x00000000000000000000000000000000000000647572616c65787365646c6578",
	"nonce": "0x0",
	"parentHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
	"timestamp": "0x00"
}

**Transactions Done:**

eth.sendTransaction({from:eth.coinbase, to: eth.accounts[1], value: 1, gas:21000})

Transactions go to pending state after **21 Transactions**

Can anyone please tell me what is the mistake.

SatpalSandhu61:
Fixes #802 
If isQuorum is set to false in genesis when using Raft, then after around 13-20 transactions are processed, all further transactions go into pending state and are not processed.

Issue is due to this historic [commit](https://github.com/jpmorganchase/quorum/commit/30a2c8c7ef53730b15d8e25433baabfe17a5dd4e#diff-83814d75e37ea3b4eb16ffc7202295d3) where Raft is unsubscribed from chainheadevent, but it doesn't take effect when isQuorum is false.
Note that current codebase explicitly executes this code for consensus mechanisms that require it (Istanbul & Clique).

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -212,7 +212,7 @@ func newWorker(config *params.ChainConfig, engine consensus.Engine, eth Backend,
 		resubmitIntervalCh: make(chan time.Duration),
 		resubmitAdjustCh:   make(chan *intervalAdjust, resubmitAdjustChanSize),
 	}
-	if _, ok := engine.(consensus.Istanbul); ok || !config.IsQuorum || config.Clique != nil {
+	if _, ok := engine.(consensus.Istanbul); ok || config.Clique != nil {
 		// Subscribe NewTxsEvent for tx pool
 		worker.txsSub = eth.TxPool().SubscribeNewTxsEvent(worker.txsCh)
 		// Subscribe events for blockchain


fixanoid (approved):

trung (changes requested):
Please review the condition so it would cover the scenario with `ethash` consensus.

jpmsam:
Raft is implemented to run only in Quorum mode. More specifically, we want to at least disable  [future](https://github.com/jpmorganchase/quorum/blob/4b45dd2004b11bf76ef1155db131b810667f73ed/consensus/ethash/consensus.go#L258) block checks when in Raft with Quorum mode.

PR #815, opened 2019-09-06
Raft issue 795 fix

### LINKED ISSUE #795: Raft cluster null after starting a single node cluster
#### System information

Geth version: 
Version: 1.8.18-stable
Quorum Version: 2.2.4
Architecture: amd64
Protocol Versions: [63 62]
Network Id: 1337
Go Version: go1.9.3
Operating System: linux
GOPATH=
GOROOT=/usr/local/go

OS & Version: 
Alpine 1.9.3

Branch, Commit Hash or Release:
Release 2.2.4 / 2.2.3

#### Expected behaviour
Raft cluster should not be [null]

#### Actual behaviour
Raft cluster is [null]

1. Start a single quorum node (permission-nodes and static-nodes should have a single entry)
2. Stop the quorum node
3. Start the quorum node again
4. raft.cluster of the node is [null]

Note : The setup is dockerized and node is stopped by stopping the docker container and started by starting the docker container

zzy96:
Fix https://github.com/jpmorganchase/quorum/issues/795 and in the meanwhile do not break https://github.com/jpmorganchase/quorum/issues/642

After analysis, the root cause is that on a single node scenario the condition `peer := pm.peers[raftId]; peer != nil` restricts the node from setting `forceSnapshot` to true. Thus, the raft snapshot is not written into WAL. When the node restarted, raft cluster is lost. `pm.isRaftIdUsed(raftId)` provides the fix by checking if raftId is the node itself.

### DIFF
--- a/raft/handler.go
+++ b/raft/handler.go
@@ -765,7 +765,7 @@ func (pm *ProtocolManager) eventLoop() {
 					case raftpb.ConfChangeAddNode:
 						if pm.isRaftIdRemoved(raftId) {
 							log.Info("ignoring ConfChangeAddNode for permanently-removed peer", "raft id", raftId)
-						} else if peer := pm.peers[raftId]; peer != nil && raftId <= uint16(len(pm.bootstrapNodes))  {
+						} else if pm.isRaftIdUsed(raftId) && raftId <= uint16(len(pm.bootstrapNodes))  {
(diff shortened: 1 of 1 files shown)


trung (approved):
LGTM

jbhurat (approved):
LGTM

PR #821, opened 2019-09-20
Change "block in the future" message to Info instead of Warn

SatpalSandhu61:
Treat "block in the future" message as informational, rather than warning (as future blocks are still processed, albeit after a delay).

Note that this is a simple fix. There is also a related [PR # 800](https://github.com/jpmorganchase/quorum/pull/800).


trung (approved):
LGTM

jpmsam (approved):

jpmsam (approved):

PR #823, opened 2019-09-23
Remove unused sleep in Raft

prd-fox:
The `sleep` function is not used, so it should be removed from production code.


trung (approved):
LGTM

PR #836, opened 2019-10-07
Use the "not in use" tx manager when env var not used

### LINKED ISSUE #834: Send private transaction failed with istanbul tools generate artifacts
I make an image with dockerfile:

_**FROM quorumengineering/quorum:2.2.5 as builder1
FROM quorumengineering/tessera:0.10.0
RUN "https://mirror.tuna.tsinghua.edu.cn/alpine/v3.10/main" > /etc/apk/repositories
RUN apk --update add python curl bash jq && rm -rf /var/cache/apk/*
COPY --from=builder1 /usr/local/bin/geth /usr/local/bin/geth
COPY --from=builder1 /usr/local/bin/bootnode /usr/local/bin/bootnode
ADD istanbul /usr/local/bin/istanbul
ADD entrypoint.sh /entrypoint.sh
RUN chmod u+x entrypoint.sh
EXPOSE 8545 8546 30303 30303/udp 9000 9080
ENTRYPOINT /entrypoint.sh**_

I start 3 nodes with k8s, and use geth attach command to open JavaScript console，create and unlock the account with personal api, send public transaction, everything is ok。 But when i send private transaction to another node, the current node seem to be destoryed, and err occur :

eth output before send private transaction:
p1
eth output after send private transaction:
p2
The private contract is from quorum-examples, just change the privateFor parameter with another node tm.pub. 
What's wrong with it, could you give me some ideas ? @vasa-develop @fixanoid

prd-fox:
If the `PRIVATE_CONFIG` environment variable isn't present, we should return an error when private transactions are sent, the same as if `PRIVATE_CONFIG=ignore` was specified.
This fixes a nil dereference crash that occurs.

Fixes https://github.com/jpmorganchase/quorum/issues/834

### DIFF
--- a/private/private.go
+++ b/private/private.go
@@ -13,12 +13,7 @@ type PrivateTransactionManager interface {
 }
 
 func FromEnvironment(name string) PrivateTransactionManager {
-	cfgPath := os.Getenv(name)
-	if cfgPath == "" {
-		//no privacy manager specified, start in public-only mode
-		return constellation.MustNew("ignore")
-	}
-	return constellation.MustNew(cfgPath)
+	return constellation.MustNew(os.Getenv(name))
 }
 
 var P = FromEnvironment("PRIVATE_CONFIG")
--- a/private/private.go
+++ b/private/private.go
@@ -12,12 +12,13 @@ type PrivateTransactionManager interface {
 	Receive(data []byte) ([]byte, error)
 }
 
-func FromEnvironmentOrNil(name string) PrivateTransactionManager {
+func FromEnvironment(name string) PrivateTransactionManager {
 	cfgPath := os.Getenv(name)
 	if cfgPath == "" {
-		return nil
+		//no privacy manager specified, start in public-only mode
+		return constellation.MustNew("ignore")
 	}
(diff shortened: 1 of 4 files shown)


fixanoid (dismissed):

SatpalSandhu61 (dismissed):

private/private.go
```diff
@@ -12,12 +12,13 @@ type PrivateTransactionManager interface {
 	Receive(data []byte) ([]byte, error)
 }
 
-func FromEnvironmentOrNil(name string) PrivateTransactionManager {
+func FromEnvironment(name string) PrivateTransactionManager {
 	cfgPath := os.Getenv(name)
 	if cfgPath == "" {
-		return nil
+		//no privacy manager specified, start in public-only mode
```
jpmsam: We don't want to default it to public but we want the user to specify whether to accept private transactions or not, hence we need PRIVATE_CONFIG to be configured.

prd-fox: Alright, I'll add a check to the startup procedure to exit if the variable isn't defined, much like the consensus check.

PR #853, opened 2019-10-15
Check if message payload is signed by the message sender

jbhurat:
When message payload was decoded, the message signature was validated to be one from the validator set, but it was not checked if the message payload was actually signed by the message sender. This PR adds that check

Thanks to @montyly [@trailofbits](https://github.com/trailofbits/) for discovering the issue and for @timmoreton for sharing it.


trung (dismissed):
LGTM

trung (approved):

jpmsam (approved):

PR #855, opened 2019-10-16
rpc: raise HTTP/WebSocket request size limit to 5MB

antonydenyer:
Increase HTTP/WebSocket request size limit to 5MB in line with go-ethereum implementation https://github.com/ethereum/go-ethereum/blob/master/rpc/http.go#L39

### DIFF
--- a/rpc/http.go
+++ b/rpc/http.go
@@ -37,7 +37,7 @@ import (
 
 const (
 	contentType             = "application/json"
-	maxRequestContentLength = 1024 * 512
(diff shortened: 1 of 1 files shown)


jpmsam:
fixes #855

jpmsam (approved):

zzy96:
This PR actually set `maxRequestContentLength` to 512MB. It causes a unit test timeout in geth 1.9.7. We should reset it back to 5MB to align with Ethereum upstream.

PR #856, opened 2019-10-17
Add block sealing information to IBFT API

prd-fox:
Allows retrieval from the API of:
- who sealed a block
- who created that block
- the current nodes signing address (derived from the `nodekey`)

This data can be manually found by users from decoding extra data or from deriving the address themselves, but it is a hassle, so these are convenience functions for quick use.


consensus/istanbul/backend/api.go
```diff
@@ -29,6 +29,61 @@ type API struct {
 	istanbul *backend
 }
 
+type BlockSigners struct {
```
jbhurat, line 33: Small niptick, can we add comments to all public types and methods

consensus/istanbul/backend/engine.go
```diff
@@ -271,22 +293,16 @@ func (sb *backend) verifyCommittedSeals(chain consensus.ChainReader, header *typ
...
-		addr, err := istanbul.GetSignatureAddress(proposalSeal, seal)
-		if err != nil {
-			sb.logger.Error("not a valid address", "err", err)
-			return errInvalidSignature
-		}
-		// Every validator can have only one seal. If more than one seals are signed by a
-		// validator, the validator cannot be found and errInvalidCommittedSeals is returned.
+	committers, err := sb.Signers(header)
+	if err != nil {
+		return errInvalidSignature
```
jbhurat: Instead of returning `errInvalidSignature`, can we just return `err` as `Signers()` can return other error type

prd-fox:
Thanks @jbhurat, I've addressed the comments.

jbhurat (approved):
LGTM

SatpalSandhu61 (approved):

jpmsam (approved):

Issue #865, opened 2019-10-29
Permission service APIs returning null data upon restart of the node 


vsmk98:
#### System information

Geth version: Quorum 2.3.0

OS & Version: Linux/OSX

#### Expected behaviour
When starting the node with new permissions feature, sometimes it's observed the permission apis (e.g. `quorumPermission.orgList` or `quorumPermission.nodeList` or `quorumPermission.acctList`) does not display any data. This is observed intermittently

#### Actual behaviour
The apis should display proper data as per the permissions model

#### Steps to reproduce the behaviour
Bring up the node in IBFT consensus and with new permissions feature. Once the node is up and running with new permissions feature, restart the network.

zzy96:
PR was merged.

PR #866, opened 2019-10-29
Permissions service not starting upon restart of the node

vsmk98:
Permissions service is dependent on RPC server to be up and monitors on event published by RPC service. The event subscription is in side a go func. This sometimes results in subscription happening after the event has been published  by RPC and resulting in the permission service waiting endlessly to start. Changes address the same.

#865


amalrajmani (approved):
looks good to me

trung (approved):
LGTM

PR #881, opened 2019-11-19
Fix premature abort during blocks processing for raft

### LINKED ISSUE #852: Two minters possible in a 3-node network in Raft resulting in fork
#### System information

Geth
Version: 1.8.18-stable
Git Commit: c894c2d
Quorum Version: 2.2.5
Architecture: amd64
Protocol Versions: [63 62]
Network Id: 1337
Go Version: go1.11.12
Operating System: linux

#### Expected behaviour
Forks should never be possible in Raft. If a node gets partitioned away from the other two nodes in a 3-node network then we expect it to be unable to mint blocks. 

#### Actual behaviour
When one of the nodes got partitioned, it lags behind the other nodes, as expected. However, submitting transactions to the partitioned node still results in blocks being minted. The other two nodes also continue to mint blocks. Thus, we end up with two minters. Blocks produced by the partitioned node are rejected by the other nodes with the error message ```` Someone else mined invalid block; ignoring block=f6f431…283a64 ````. 

We believe this may be a consequence of speculative minting combined with unfortunate interleaving of messages caused by temporary network partitions.

We are currently trying to reproduce this; we have observed this once in production at a fairly high block height (~36000).

#### Backtrace
We do not have detailed logs for the point of divergence but after they have diverged, the following error is printed repeatedly:
````
2098 INFO [10-14|08:56:50.606|raft/persistence.go:53]                 persisted the latest applied index       index=35988
2099 INFO [10-14|08:56:50.606|raft/minter.go:139]                     Handling InvalidRaftOrdering             invalid block=f6f431…283a64 current head=81b94e…4f5bc8
2100 INFO [10-14|08:56:50.606|raft/minter.go:146]                     Someone else mined invalid block; ignoring block=f6f431…283a64
````

### PR DESCRIPTION
**Pull request by zzy96**
Fix https://github.com/jpmorganchase/quorum/issues/852

Prevent raft from updating persisted storage when `bc.insertChain` is interrupted. This fixes the issue of chain fork and constant `InvalidRaftOrdering`.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1149,6 +1149,11 @@ func (bc *BlockChain) insertChain(chain types.Blocks) (int, []interface{}, []*ty
 		// If the chain is terminating, stop processing blocks
 		if atomic.LoadInt32(&bc.procInterrupt) == 1 {
 			log.Debug("Premature abort during blocks processing")
+			// QUORUM
+			if bc.chainConfig.Istanbul == nil && bc.chainConfig.Clique == nil {
+				return i, events, coalescedLogs, ErrInsertChainInterrupted
+			}
+			// END QUORUM
 			break
 		}
 		// If the header is a banned one, straight out abort
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1150,7 +1150,7 @@ func (bc *BlockChain) insertChain(chain types.Blocks) (int, []interface{}, []*ty
 		if atomic.LoadInt32(&bc.procInterrupt) == 1 {
 			log.Debug("Premature abort during blocks processing")
 			// QUORUM
-			if bc.chainConfig.Istanbul == nil && bc.chainConfig.Clique == nil {
+			if bc.chainConfig.IsQuorum && bc.chainConfig.Istanbul == nil && bc.chainConfig.Clique == nil {
 				// Only returns an error for raft mode
 				return i, events, coalescedLogs, ErrAbortBlocksProcessing
 			}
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: core/blockchain.go [hunk 1]
```diff
@@ -1149,6 +1149,12 @@ func (bc *BlockChain) insertChain(chain types.Blocks) (int, []interface{}, []*ty
 		// If the chain is terminating, stop processing blocks
 		if atomic.LoadInt32(&bc.procInterrupt) == 1 {
 			log.Debug("Premature abort during blocks processing")
+			// QUORUM
+			if bc.chainConfig.Istanbul == nil && bc.chainConfig.Clique == nil {
+				// Only returns an error for raft mode
```
trung, line 1154: If this is `raft` specific, can you review the `if` condition to make sure we target `raft` and `raft` only.
As the current condition may satisfy even for `ethhash` or any future consensus

zzy96: Add `IsQuorum` condition to make sure it is only satisfied for `raft`.

trung: Please use `bc.hc.Engine()` to identify consensus algorithm.
For raft, it's `Ethash` with `ModeFullFake` per:
https://github.com/jpmorganchase/quorum/blob/5cb92ceb3ccde7d30464f43df5f869aec3722499/eth/backend.go#L278-L282

zzy96: Changed to use `bc.engine.Protocol().Name` to identify consensus algorithm. To reduce redundant logic, reuse `bc.GetConsensus` function in `pm.getConsensusAlgorithm`. Rewrote the test for `pm.getConsensusAlgorithm` which will internally cover `bc.GetConsensus`.

zzy96: Reverted to `bc.chainConfig.IsQuorum && bc.chainConfig.Istanbul == nil && bc.chainConfig.Clique == nil`. We have a check here to make sure Quorum only run one of these three consensuses for now. https://github.com/jpmorganchase/quorum/blob/5cb92ceb3ccde7d30464f43df5f869aec3722499/cmd/geth/config.go#L266-L277

jbhurat (approved):
LGTM

jpmsam (approved):

PR #884, opened 2019-11-19
Restrict p2p connection for non raft peer

### LINKED ISSUE #883: No p2p level connection restriction for non raft peers
There are two issues for raft without restriction on p2p connection:
- A removed raft peer can still connect to the network and send transaction
- A node start with a nodekey different from the one provided in `raft.addPeer` but use the correct raft ID in `--raftjoinexisting` can still connect to the network, sync block and send transaction

zzy96:
Fix https://github.com/jpmorganchase/quorum/issues/883

### DIFF
--- a/raft/handler.go
+++ b/raft/handler.go
@@ -140,6 +140,8 @@ func (pm *ProtocolManager) Start(p2pServer *p2p.Server) {
 	pm.p2pServer = p2pServer
 	pm.minedBlockSub = pm.eventMux.Subscribe(core.NewMinedBlockEvent{})
 	pm.startRaft()
+	// update raft peers info to p2p server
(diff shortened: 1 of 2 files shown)


raft/handler.go
```diff
@@ -140,6 +140,8 @@ func (pm *ProtocolManager) Start(p2pServer *p2p.Server) {
 	pm.p2pServer = p2pServer
 	pm.minedBlockSub = pm.eventMux.Subscribe(core.NewMinedBlockEvent{})
 	pm.startRaft()
+	// update raft peers info to p2p server
+	pm.p2pServer.SetCheckPeerInRaft(pm.peerExist)
```
jpmsam, line 147: We shouldn't create a cross dependency in the server on this function in the handler.

jbhurat (approved):
LGTM

PR #885, opened 2019-11-19
DNS support for static nodes and discovery

prd-fox:
Adds full DNS support for Quorum.

This includes:

use of DNS hostnames in the static-nodes.json file, so that we can connect to our static nodes even when the IP address of the DNS name has changed, meaning less updating of the static node definition file. This is an improvement to previous behaviour which resolved the DNS hostname once at startup and then only tracked the IP address


Krish1979 (dismissed):

mkdocs.yml
```diff
@@ -18,6 +18,7 @@ nav:
         - Creating a Network From Scratch: Getting Started/Creating-A-Network-From-Scratch.md
         - Running Quorum: Getting Started/running.md
         - Quorum API: Getting Started/api.md
+        - CLI arguments: Getting Started/command-line-arguments.md
```
trung: I think we don't really need this page as it seems to be a duplicated of `geth --help`

mkdocs.yml
```diff
@@ -75,7 +76,9 @@ nav:
         - Overview: Cakeshop/Overview.md
         - Getting Started: Cakeshop/Getting started.md
         - Cakeshop FAQ: Cakeshop/Cakeshop FAQ.md
-    - Product Roadmap: roadmap.md         
+    - Quorum Features:
```
trung, line 78: As all features are in the 1st level, DNS is deserved in the first level too.
Otherwise we could move others to `Enterprise Features` menu. But i prefer the former.

p2p/shared_udp_connection.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	"errors"
+	"github.com/ethereum/go-ethereum/p2p/discover"
+	"net"
+)
+
+// UDPReader implements a shared connection. Write sends messages to the
+// underlying connection while read returns messages that were found
+// unprocessable and sent to the unhandled channel by the primary listener.
+type UDPReader interface {
```
trung: This interface has number of methods that deal with UDP including read/write/close. It's deserved a more descriptive name than just a `Reader`.  And it seems a dup of local interface `conn` from `p2p/discover` package.

Is there a chance to move some to `discover` package or expose from `discover` package instead?

p2p/discover/udp_quorum.go
```diff
@@ -0,0 +1,479 @@
...
+		Rest []rlp.RawValue `rlp:"tail"`
+	}
+
+	// findnode is a query for nodes close to the given target.
+	findnodeQuorum struct {
+		Target     encPubkey
+		Expiration uint64
+		// Ignore additional fields (for forward compatibility).
+
+		IsQuorum   bool
```
trung: Why do we need this field here?

Same for for `neighborsQuorum`

p2p/discover/udp_quorum.go
```diff
@@ -0,0 +1,479 @@
...
+	case <-t.closing:
+		return false
+	}
+}
+
+var (
+	// Neighbors replies are sent across multiple packets to
+	// stay below the 1280 byte limit. We compute the maximum number
+	// of entries by stuffing a packet until it grows too large.
+	maxNeighborsQuorum = 3
```
trung: this value is calculated in `udp.go#init()`, why do we hard code here?

trung (commented):
Initial quick feedbacks

p2p/enode/urlv4.go
```diff
@@ -103,6 +103,37 @@ func NewV4(pubkey *ecdsa.PublicKey, ip net.IP, tcp, udp, raftPort int) *Node {
 	return n
 }
 
+// Quorum
```
libby, line 112: We could break out the shared logic between `NewV4` and `NewV4Hostname` into a private function `newV4` this would make it easier to see there is no change in the logic except for hostname vs IP type. 
```
// NewV4 creates a node from discovery v4 node information. The record
// contained in the node has a zero-length signature.
func NewV4(pubkey *ecdsa.PublicKey, ip net.IP, tcp, udp, raftPort int) *Node {
	var r enr.Record
	if ip != nil {
		r.Set(enr.IP(ip))
	}
        if raftPort != 0 { // Quorum specific (TODO: can we remove this here? and from the function sig?)
		r.Set(enr.RaftPort(raftPort))
	}
	return newV4(pubkey, r, tcp, udp, raftPort)
}
// broken out from `func NewV4` (above) same in upstream go-ethereum, but taken out
// to avoid code duplication b/t NewV4 and NewV4Hostname
func newV4(pubkey *ecdsa.PublicKey, r enr.Record, tcp, udp, raftPort int) *Node {
     if udp != 0 {
       r.Set(enr.UDP(udp))
     }
     if tcp != 0 {
       r.Set(enr.TCP(tcp))
     }

     if raftPort != 0 { // Quorum
       r.Set(enr.RaftPort(raftPort))
     }

     signV4Compat(&r, pubkey)
     n, err := New(v4CompatID{}, &r)
     if err != nil {
       panic(err)
     }
     return n
}
// Quorum

// NewV4Hostname creates a node from discovery v4 node information. The record
// contained in the node has a zero-length signature. It sets the hostname of
// the node instead of the IP address.
func NewV4Hostname(pubkey *ecdsa.PublicKey, hostname string, tcp, udp, raftPort int) *Node {
	var r enr.Record
	if hostname != "" {
		r.Set(enr.Hostname(hostname))
	}
	if raftPort != 0 {
		r.Set(enr.RaftPort(raftPort))
	}
	return newV4(pubkey, r, tcp, udp, raftPort)
}

// End-Quorum
```
Additionally, it seems the `raftPort` is only set in `parseComplete` so the signature for `NewV4` could lose the raftPort `func NewV4(pubkey *ecdsa.PublicKey, ip net.IP, tcp, udp int) *Node` and thus match upstream (https://github.com/ethereum/go-ethereum/blob/master/p2p/enode/urlv4.go#L85), and the Quorum specific `raftport` could be set only in the Quorum specific `NewV4Hostname(pubkey *ecdsa.PublicKey, hostname string, tcp, udp, raftPort int)` before calling `func newV4(pubkey *ecdsa.PublicKey, r enr.Record, tcp, udp int) *Node ` (TODO: check me on this).

p2p/enode/urlv4.go
```diff
@@ -124,19 +155,17 @@ func parseComplete(rawurl string) (*Node, error) {
...
-	}
-	if ip = net.ParseIP(host); ip == nil {
-		return nil, errors.New("invalid IP address")
+		return nil, err
 	}
+	ip = ips[0]
 	// Ensure the IP is 4 bytes long for IPv4 addresses.
 	if ipv4 := ip.To4(); ipv4 != nil {
 		ip = ipv4
 	}
```
libby: This logic has changed slight in upstream https://github.com/ethereum/go-ethereum/blob/master/p2p/enode/urlv4.go#L130 we might want to look into syncing it up in this PR. Additionally, a quorum specific comment might be helpful. Something like: 
```
 // Quorum: is only testing the hostname resolves to  a validate the IPv4 addr, but unlike upstream geth, it does not pass the resolved IP on, but rather uses the hostname.
// Parse the IP address.
...
 ```

File: p2p/enode/urlv4_test.go [hunk 8]
```diff
@@ -55,39 +54,39 @@ var parseNodeTests = []struct {
...
 		rawurl: "enode://1dd9d65c4552b5eb43d5ad55a2ee3f56c6cbc1c64a5c8d659f51fcd51bace24351232b8d7821617d2b29b54b81cdefb9b3e9c37d7fd5f63270bcc9e1a6f6a439@127.0.0.1:52150",
-		wantResult: NewV4(
+		wantResult: NewV4Hostname(
 			hexPubkey("1dd9d65c4552b5eb43d5ad55a2ee3f56c6cbc1c64a5c8d659f51fcd51bace24351232b8d7821617d2b29b54b81cdefb9b3e9c37d7fd5f63270bcc9e1a6f6a439"),
-			net.IP{0x7f, 0x0, 0x0, 0x1},
+			"127.0.0.1",
 			52150,
 			52150,
 			0,
 		),
```
libby, line 59: Should we keep the `NewV4` test and just add `NewV4Hostname` tests? Could even be in another file like `p2p/enode/urlv4_quorum_test.go`.

libby (changes requested):
Initial feedback =)

mkdocs.yml
prd-fox: I can expand it a bit more as I agree that they are quite similar. Ideally `geth --help` should be a one liner of the flag, whereas this page can give a bit more description.
The flags will likely be explained further (or repeated) in their respective doc pages, but it is helpful IMO to have an aggregate page of all flags.

mkdocs.yml
prd-fox: It was intended to be a place to put features that don't fit the "main" categories. Adding them to the top level would create too much clutter, but happy to consider alternatives.

p2p/discover/udp_quorum.go
prd-fox: The value calculated in `udp.go#init()` (12, if I remember correctly) is so because a record in discovery V4 is a fixed size, so you can always fit 12 records in a single packet. 
Since hostnames are a variable size, this is no longer true. 3 was fairly arbitrary but unlikely to ever reach the limit.
I'll change it to check the size after each node has been added and see if it has gone over the limit.

p2p/discover/udp_quorum.go
prd-fox: I considered this a cheap way to differentiate the packet from a DiscV4 packet. A DiscV4 packet will fail decoding when checked in the Quorum DiscV4 decoder, and get passed on appropriately.

p2p/shared_udp_connection.go
prd-fox: Agree about the name.
Do you think I should export the `conn` interface from `discover` instead?

p2p/enode/urlv4.go
prd-fox: I made this change, although I originally kept them separate to avoid changing upstream code (only adding onto). As far as the raft port is concerned, you may be right, but I don't plan on messing around with that as it doesn't pertain to this change.

p2p/enode/urlv4.go
prd-fox: I'm not enthusiastic on pulling in individual upstream changes unless its to fix a critical bug. We will get this change when we upgrade, and we can keep our version (1.8.18) as close to the same version upstream.
As you note in your comment, we don't use the IP anyway, rather passing in the hostname field (which may be an IP).
Let me know if you feel strongly that we do pull this change in.

p2p/enode/urlv4_test.go
prd-fox: That sounds like a good idea.

mkdocs.yml
trung: CLI flags are expected to be explained well in their respected features doc. Creating a central page  is useful but I see some maintenance overhead and duplication of docs here.

prd-fox: We can add it later with a more comprehensive set of options.

p2p/enode/urlv4_test.go
prd-fox: Hm, then again, since the tests make use of `parseComplete` which has been modified to use `NewV4Hostname`, the original tests don't make sense. We could duplicate the entire functionality of `parseComplete` and all its callsites, but that seems like a lot of effort for little gain.

p2p/server.go
```diff
@@ -533,25 +520,50 @@ func (srv *Server) setupDiscovery() error {
...
 			Unhandled:   unhandled,
 		}
-		ntab, err := discover.ListenUDP(conn, srv.localnode, cfg)
+
+		var tab *discover.Table
+		if tabCasted, ok := srv.ntab.(*discover.Table); ok {
+			tab = tabCasted
+		}
+
+		_, err := discover.ListenUDPExisting(quorumSharedConnection, srv.localnode, cfg, tab)
```
trung: Are these lines from 551 to 566 redundant? Lines 547-550 can be moved to 529?

p2p/discover/udp.go
```diff
@@ -268,6 +268,39 @@ func newUDP(c conn, ln *enode.LocalNode, cfg Config) (*Table, *udp, error) {
 	return udp.tab, udp, nil
 }
 
+// Quorum
+
+// ListenUDPExisting uses an existing table that listens for UDP packets on the connection.
+func ListenUDPExisting(c conn, ln *enode.LocalNode, cfg Config, tab *Table) (*Table, error) {
```
jbhurat: Existing `ListenUDP()` and `newUDP()` functions are no longer referenced anywhere, is the plan to delete those?

prd-fox: I hadn't planned to, in order to keep the diff from upstream smaller. `newUDP()` is still used in the upstream tests (`udp_test.go`).

raft/handler.go
```diff
@@ -656,11 +675,24 @@ func (pm *ProtocolManager) addPeer(address *Address) {
...
 
 	// Add P2P connection:
-	p2pNode := enode.NewV4(pubKey, address.Ip, int(address.P2pPort), 0, int(address.RaftPort))
+	var ip net.IP
+	if ip = net.ParseIP(address.Hostname); ip == nil {
+		// QUORUM
+		// attempt to look up IP addresses if host is a FQDN
+		lookupIPs, err := net.LookupIP(string(address.Hostname))
+		if err != nil {
+			panic(err.Error())
```
trung: Node is not able to start. This happens when peers' hostnames and IPs have not yet registered in DNS. 

It is a challenge for environments where  DNS entry is only available when VM starts up.

jbhurat (commented):
I see that there is a lot of duplication between `udp.go` and `udp_quorum.go` is it possible to re-evaluate `udp.go` and `udp_quorum.go` to see there is no code duplication

prd-fox:
I have removed the discovery section of the PR, so discovery will only use IP addresses as before. I have resolved comments relating to those portions.

trung (dismissed):
LGTM

jbhurat (dismissed):
LGTM

nicolae-leonte-go (approved):

jpmsam (approved):

p2p/enode/urlv4.go
libby: I don't feel strongly, it was only because the change from upstream was being brought in and 
 that had fallen behind (https://github.com/ethereum/go-ethereum/commit/b90cdbaa79cfe438aab0f1389d35980f3d38ec84#diff-0532cc1dc91dc5633d79c93dac1cd87bR128). Removing the logic (as you have now done) seems fine to me and less confusing as it isn't being passed in.

jimthematrix:
thank you Quorum team for making this enhancement!

PR #886, opened 2019-11-19
Fix for Raft stability 

vsmk98:
increased the size of blockProposal and chainHeadEvent channels. Changed the call to mintNewBlock from a go routine to a function call. Changes done to improve performance of Raft in high load envs

### DIFF
--- a/raft/minter.go
+++ b/raft/minter.go
@@ -89,7 +89,7 @@ func newMinter(config *params.ChainConfig, eth *RaftService, blockTime time.Dura
 		speculativeChain: newSpeculativeChain(),
 
 		invalidRaftOrderingChan: make(chan InvalidRaftOrdering, 1),
-		chainHeadChan:           make(chan core.ChainHeadEvent, 10),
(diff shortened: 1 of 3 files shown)


raft/minter.go
```diff
@@ -89,7 +89,7 @@ func newMinter(config *params.ChainConfig, eth *RaftService, blockTime time.Dura
 		speculativeChain: newSpeculativeChain(),
 
 		invalidRaftOrderingChan: make(chan InvalidRaftOrdering, 1),
-		chainHeadChan:           make(chan core.ChainHeadEvent, 1),
+		chainHeadChan:           make(chan core.ChainHeadEvent, 10),
```
jbhurat: Instead of hardcoded value can we use `tx_pool.go::chainHeadChanSize` constant, in case if the buffer size is changed later, it is changed at all the places

vsmk98: Changed to use the chainHeadChanSize constant instead of hard coded value

jpmsam (approved):

Issue #825, opened 2019-09-24
Quorum reelection loop under heavy load


ThomasRalee:
Hi,
i am not sure if what i've encountered is related to this issue logged on [etcd -Leadership changes under heavy load](https://github.com/etcd-io/etcd/issues/8553). It seems pretty similar, these are more research done on [ETCD - heartbeat interval config](https://github.com/jpmorganchase/quorum/blob/58fc5cd59d68c4c3e658f995b179713c7cc47336/raft/handler.go#L440)

Logging this issue here hopefully to get some additional insights. 

#### System information

Geth version: `2.2.5` - RAFT
Tessera Version: `0.10.0`

#### Expected behaviour
Election should take plan smoothly

#### Actual behaviour
Reelection happens on 2 seconds interval

#### Steps to reproduce the behaviour
- this is a very rare issue that happened, I was unable to reproduce this on a step-by-step basis 

#### Backtrace - filtered logs to show how often raft leadership election was taking place.
````
"raft2019/09/19 08:18:07 INFO: 1 became candidate at term 287
"
"raft2019/09/19 08:18:07 INFO: 1 is starting a new election at term 286
"
"raft2019/09/19 08:18:05 INFO: 1 [logterm: 2, index: 4] sent MsgVote request to 2 at term 286
"
"raft2019/09/19 08:18:05 INFO: 1 received MsgVoteResp from 1 at term 286
"
"raft2019/09/19 08:18:05 INFO: 1 became candidate at term 286
"
"raft2019/09/19 08:18:05 INFO: 1 is starting a new election at term 285
"
"raft2019/09/19 08:18:04 INFO: 1 [logterm: 2, index: 4] sent MsgVote request to 2 at term 285
"
"raft2019/09/19 08:18:04 INFO: 1 received MsgVoteResp from 1 at term 285
"
"raft2019/09/19 08:18:04 INFO: 1 became candidate at term 285
"
"raft2019/09/19 08:18:04 INFO: 1 is starting a new election at term 284
"
"raft2019/09/19 08:18:02 INFO: 1 [logterm: 2, index: 4] sent MsgVote request to 2 at term 284
"
"raft2019/09/19 08:18:02 INFO: 1 received MsgVoteResp from 1 at term 284
"
"raft2019/09/19 08:18:02 INFO: 1 became candidate at term 284
"
"raft2019/09/19 08:18:02 INFO: 1 is starting a new election at term 283
"
[...]

Issue #790, opened 2019-07-31
Raft - upgrade to v3.3.13 and addition of learner node feature 


amalrajmani:
currently quorum is using an old version of raft library(v3.1.0).
We should upgrade quorum's Raft library to latest v3.3.13 ( https://github.com/etcd-io/etcd/releases/tag/v3.3.13 ) as there are numerous improvements and some new features(like learner node) in raft library. 
We should also enhance quorum to support adding learner node and promoting learner node to peer node.

shun-tak:
Excellent update! I think this can also resolve #180

fixanoid:
❤️

Issue #180, opened 2017-09-14
Allow adding non-consensus-participating nodes to a network using Raft consensus


jcgarridosanlab:
#### System information

Geth version: 1.5.0-unstable

OS & Version: Ubuntu 16.04 

Branch, Commit Hash or Release: Quorum 1.1.0 Constellation 0.1.0

#### Expected behaviour
We have a 7 nodes network and we want 5 of them ("core") to be participating in the raft consensus algorithm and creating new blocks. The two "relay" nodes left should have a copy of the blockchain updated with the transactions and should be able to send transactions to the network but they don't participate on the raft consensus.

¿Is it possible to have such a kind of network (with nodes sending transactions but not participating in the raft consensus?

#### Actual behaviour
The "core" nodes (that use raft consensus) are seeing each other and can perform transactions (the block number increases when a transaction is made). The "relay" nodes block number does not increase although the admin.peers shows the "core" nodes only if we add them as boot nodes on the geth start script.

#### Steps to reproduce the behaviour
-Create a network using raft with 5 nodes. The static-nodes.json file of these nodes contains the enodes of these 5 nodes.
-Add 2 more nodes ("relay") without the --raft flag including the former 5 nodes in the boot node parameter. They do not have static-nodes.json file
-Send a transaction to one of the 5 core nodes
-Check that the "relay" node blocknumber does not increase
-Unlock an account in one of the relay nodes
-Send a transaction to this relay node and check that it says the account is locked

patrickmn:
This is not possible in Raft at the moment, but is a desired feature. What you want is possible using the Istanbul consensus option.

fixanoid:
Implemented as of Quorum v2.4.0

PR #887, opened 2019-11-19
Quorum raft upgrade v3.3.13

vsmk98:
This PR contains the following changes to Raft consensus:

- Upgrade quorum's Raft library to version v3.3.13 from upstream.
- Raft API to add raft learner node in quorum. learner node will follow the leader but will not vote or take part in quorum for commits.
- Raft API for promoting a learner node to peer node
- Disallow learner node from adding another learner or peer 
- Disallow learner node from promoting itself or other learners to peer
- Disallow learner node to remove other peers or learners from raft cluster
- Allow learner node to remove itself from raft cluster

#825, #790, #852, #180


lookfwd:
Wondering how has this been tested?

amalrajmani:
> Wondering how has this been tested?

When a new node joins the network as a peer, it tries to sync with leader and during this time the leader is overloaded and can potentially lead to periods of cluster unavailability. Please refer to https://github.com/etcd-io/etcd/blob/master/Documentation/learning/design-learner.md  for further details on various scenarios. The advantage of a learner node is it joins the network as a non voting member and hence does not impact the Raft quorum.
For testing this, brought up a 3 node raft network and created few thousand blocks. Then added a 4th node as peer. While the 4th node is syncing with the network, brought down one of the other nodes. Now if any transaction is send, the transaction will wait till the newly joined node is fully synced with the network.
Repeated the same test adding the 4th node as learner. In this case the network will be able to process the transactions while the new learner node is still syncing with the network

jbhurat (commented):
When a node is learner and it calls `raft.removePeer()` on itself, it returns null. When checking `raft.cluster` the node is still seen with a **`verifier`** role. Running `raft.role` returns an empty string. Running `raft.cluster` from any other node returns correct results though

jbhurat (dismissed):
LGTM

amalrajmani:
> When a node is learner and it calls `raft.removePeer()` on itself, it returns null. When checking `raft.cluster` the node is still seen with a **`verifier`** role. Running `raft.role` returns an empty string. Running `raft.cluster` from any other node returns correct results though

this has been fixed now. after a learner/peer node removes itself it will return an empty array when `raft.cluster` is called

trung (approved):
LGTM. Changes in the original etcd raft library:
- Inject a channel to capture role change: `roleC`
- Populate `ConfState` to separate voters and learners, similar to [raft: raft learners should be returned after applyConfChange by absolute8511 · Pull Request #9116 · etcd-io/etcd](https://github.com/etcd-io/etcd/pull/9116/files)

jpmsam (approved):

PR #892, opened 2019-11-25
Update file descriptor limits for macOS

### LINKED ISSUE #889: go 1.13 on macOS gives file descriptor problem
I understand (from Slack history, thread from Sep 7) that for the moment only Go 1.11.x is supported. This is not clearly documented, and really difficult to figure out if you naively try to follow the getting started docs (i.e. https://github.com/jpmorganchase/quorum-examples#running-locally)

On MacOS, the failing behaviour of Go 1.13 is: `Fatal: Failed to raise file descriptor allowance: invalid argument` which is very difficult to connect to the need to downgrade Go. (See also #756.)

I think it would be helpful to:
* document the Go 1.11.x limitation better
* enforce the Go 1.11.x limitation in build/env.sh, so that if you try building with go 1.13, you get a nice loud error
* use something like the code in https://github.com/dedis/onet/commit/03a8c57f6ba53d5e430bfe5a5b032b18a1093a53#diff-f7b6c3a5b64e312c20712b3c2ae51345 in order to make file descriptor setting work in MacOS again

I don't know if upstream has already solved this, or if I should be reporting it there.

prd-fox:
This PR is cherry-picked from:
https://github.com/ethereum/go-ethereum/pull/19035
https://github.com/ethereum/go-ethereum/pull/19068
https://github.com/ethereum/go-ethereum/pull/19141
https://github.com/ethereum/go-ethereum/pull/19166 (only relevant files)

This introduces proper file handle limits for `darwin` based OSes (macOS/OSX), and a few other test fixes.

This fixes the main complaint when building with newer versions of Golang, as Quorum currently only supports up to version 1.11.x.

As far as I can tell, there are no other blockers for 1.12, although someone feel free to correct me :)

Fixes https://github.com/jpmorganchase/quorum/issues/889

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -878,10 +878,11 @@ func makeDatabaseHandles() int {
 	if err != nil {
 		Fatalf("Failed to retrieve file descriptor allowance: %v", err)
 	}
-	if err := fdlimit.Raise(uint64(limit)); err != nil {
+	raised, err := fdlimit.Raise(uint64(limit))
+	if err != nil {
 		Fatalf("Failed to raise file descriptor allowance: %v", err)
 	}
-	return limit / 2 // Leave half for networking and other stuff
+	return int(raised / 2) // Leave half for networking and other stuff
 }
 
 // MakeAddress converts an account specified directly as a hex encoded string or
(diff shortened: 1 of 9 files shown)


jbhurat (commented):
The changes look good. I believe we should also update Dockerfile and travis to use go 1.13

prd-fox:
@jbhurat I agree, although I would prefer to make that a follow up PR to keep the scopes small.

jbhurat (approved):
> @jbhurat I agree, although I would prefer to make that a follow up PR to keep the scopes small.

Sounds good. LGTM

PR #896, opened 2019-11-29
Disabling swarm tests

jbhurat:
Disabling Swarm tests as we do not use swarm and they intermittently fail the build

This would disable swarm-related tests in all branches and PRs


trung (approved):
LGTM

prd-fox (approved):

jpmsam (approved):

PR #897, opened 2019-12-03
p2p: fix connection leakage when peer is not authorized to connect

trung:
When returning nil in `p2p/server.go#setupConn()`, it often indicates that the peer connection is established.
In the below scenarios:
1. Node is not in raft cluster (#884)
1. Node is not permissioned per smart-contract based permissioning model (#715)

We want to disconnect the peer and clean up the connection. This is done by returning an appropriate `peerError` which then follows the existing catching and handling logic by the caller.

### DIFF
--- a/p2p/peer_error.go
+++ b/p2p/peer_error.go
@@ -26,9 +26,22 @@ const (
 	errInvalidMsg
 )
 
+// Quorum
+//
+// Constants for peer connection errors
+const (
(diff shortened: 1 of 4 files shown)


jpmsam (dismissed):

jbhurat (dismissed):
LGTM

jbhurat (approved):

PR #900, opened 2019-12-04
raft: correct the setup sequence for applied index related test

trung:
Make sure raft cluster is fully stopped before changing the applied index value in the local db


jbhurat (approved):
LGTM

jpmsam (approved):

PR #910, opened 2019-12-13
fix raft minter

kjx98:
using nanoseconds instead of normal seconds in Time field of block head stucks ethereumjs, change back to seconds since epoch

### DIFF
--- a/raft/minter.go
+++ b/raft/minter.go
@@ -265,7 +265,7 @@ func (minter *minter) createWork() *work {
 		GasLimit:   minter.eth.calcGasLimitFunc(parent),
 		GasUsed:    0,
 		Coinbase:   minter.coinbase,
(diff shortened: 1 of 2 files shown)


SatpalSandhu61:
With raft-based consensus we can produce far more than one block per second, therefore we cannot use seconds for the block timestamp (every block must have a timestamp greater than its parent).

kjx98:
I see. Then the Time of head should be specially interpreted, it's better adding subsecond field while converting to json.

SatpalSandhu61:
We do actually have some work ongoing in this area which would solve this issue.

jpmsam:
As explained by @SatpalSandhu61 , we are looking into different options for enhancing this but we can't use seconds.

PR #922, opened 2020-01-08
Add Quorum eth namespace RPC methods to console

prd-fox:
Methods are not automatically added to the built-in JS console, even though they may be available over the RPC API. Two Quorum specific methods (`eth_getQuorumPayload` and `eth_sendTransactionAsync`) were not updated to be accessible via the JS console. This commit adds them.
**note**: adding them under the `eth` namespace in a separate subsection didn't seem to work, so I've added Quorum comment boundary markers.

During testing of `eth_sendTransactionAsync` to see if the extra parameters outside of upstream `SendTxArgs` object are kept using the web3.js `inputTransactionFormatter` are retained (they are, so using the formatter is okay), I found two log messages not using the correct logging format, so they have been updated to use the proper number of arguments.


trung (approved):
LGTM

jpmsam (approved):

PR #925, opened 2020-01-13
Create tutorial for adding Quorum nodes

prd-fox:
Add a new subsection named "How-to Guides" that will contain tutorials on how to perform various tasks in Quorum.

Added a tutorial on how to add Quorum nodes to an existing network, both for Raft & IBFT, along with various extra options e.g. permissioning.

Depends on resources added in https://github.com/jpmorganchase/quorum-examples/pull/218


docs/How-To-Guides/add_ibft_validator.md
```diff
@@ -0,0 +1,291 @@
...
+          authorize: true,
+          block: 268,
+          validator: "0xd8dba507e85f116b1f7e231ca8525fc9008a6966"
+      }]
+    }
+    ```
+   
+   Let's break this down.
+   Firstly, we proposed the address `0xb131288f355bc27090e542ae0be213c20350b767` to be added; that is what the `true` 
+   parameter is for. If we has set it to `false`, that means we want to remove an existing validator with that address.
```
jbhurat: Typo here
> If we has set it to \`false\`

Should instead be `If we had set it to false`

fixanoid (dismissed):

jbhurat (approved):
LGTM

Issue #931, opened 2020-01-28
Issues while electing leader on a single node quorum cluster.


navdeep-kaur:
Hi there,

On upgrading to quorum 2.4, when we are starting up the cluster with a single node, the raft_cluster is responding with {"jsonrpc":"2.0","id":1,"error":{"code":-32000,"message":"**no leader is currently elected**"}}. The single node is unable to elect itself as the leader and is always chosen as a candidate instead.

#### Expected behaviour
The single node should be elected as the leader.

The idea is to spin up a single boot node and run add peer for all the other nodes that we want to add to the quorum network.

vsmk98:
Hi @navdeep-kaur - I checked in 2.4.0 version bringing up a single node cluster and it works fine for me.
```
➜  7nodes git:(master) ✗ geth attach qdata/dd1/geth.ipc
Welcome to the Geth JavaScript console!

instance: Geth/v1.8.18-stable(quorum-v2.4.0)/darwin-amd64/go1.11.13
coinbase: 0xed9d02e382b34818e88b88a309c7fe71e65f419d
at block: 0 (Thu, 01 Jan 1970 07:30:00 +0730)
 datadir: /Users/saiv/temp/quorum-examples/examples/7nodes/qdata/dd1
 modules: admin:1.0 debug:1.0 eth:1.0 ethash:1.0 miner:1.0 net:1.0 personal:1.0 raft:1.0 rpc:1.0 txpool:1.0 web3:1.0

> raft.cluster
[{
    hostname: "127.0.0.1",
    nodeId: "ac6b1096ca56b9f6d004b779ae3728bf83f8e22453404cc3cef16a3d9b96608bc67c4b30db88e0a5a6c6390213f7acbe1153ff6d23ce57380104288ae19373ef",
    p2pPort: 21000,
    raftId: 1,
    raftPort: 50401,
    role: "minter"
}]
```
Can you please check if the `static-nodes.json` is having only one entry (the node which you are bringing up)?

navdeep-kaur:
@vsmk98 we have entries for all nodes (that will be added ) in the quorum network in the static-nodes.json when we start the network with a single node. This used to work in v2.2.3. Has something changed around this?

MrHmP:
@vsmk98 this used to work in v2.2.3, has something updated in v2.4.0
@jpmsam @felixonmars

vsmk98:
[...]

PR #934, opened 2020-02-01
raft.cluster api returns error if no leader is there in the network in version 2.4.0

vsmk98:
In Quorum version 2.4.0, call to raft.cluster returns in error no leader is currently elected if the network is not having a leader node. This is different from behaviour in version where all nodes part of cluster are listed even when the leader is not elected.

This fix addresses the same. The api response has been addressed to include a new attribute nodeActive which will be true if the node is part for raft cluster else will be false. Further if the network is not having a valid leader, the role of each node will be shown as "" . Error handling added for other apis.

#931


jbhurat (dismissed):
LGTM

fixanoid (dismissed):

jpmsam (dismissed):

jpmsam (approved):

PR #939, opened 2020-02-09
Add some parameters for puppeth to support non-interactive mode when we want a genesis.json

dipingxian2:
Add some parameters for puppeth to support non-interactive mode when we want a genesis.json. 
For example :
puppeth --consensusType 2 --network test12 --blocksTime 10 --sealAccounts "25Cde39d96684E2a681Ae0289b37AF8E9859ed99, 25Cde39d96684E2a681Ae0289b37AF8E9859ed98, 25Cde39d96684E2a681Ae0289b37AF8E9859ed97" --preFundedAccounts "25Cde39d96684E2a681Ae0289b37AF8E9859ed99, 25Cde39d96684E2a681Ae0289b37AF8E9859ed98, " --networkID 231


jpmsam:
Did you submit these changes to upstream geth?

dipingxian2:
Hi jpmsam, you mean go-ethereum?  
No, so far, because I think ethereum focus on public network. quorum used by private network, maybe non-interactive puppeth is more useful in quorum.

jpmsam:
Geth is also used in private networks and hence puppeth was built. I'd prefer if this is pushed upstream as it can be used in both Geth and Quorum.

PR #944, opened 2020-02-20
Fix params number for debug.dumpBlock in web3ext

zzy96:
Quorum has modified `debug.dumpBlock` to accept two parameters but this change is not reflected in geth console. Update it in this PR.


trung (commented):
Does this API support backward compatibility? It means if `debug_blockDump` is called with single param (which is the block number), does it return the dump from public state by default?

zzy96:
Hi @trung, 

I don't think so. It requires mandatory two params since the function signature is modified two years ago. Without a proper "public" or "private" type, it will return "unknown type" error.

If we want to make sure backward compatibility. I believe the best way is to add a new `debug_privateBlockDump` API and keep the `debug_blockDump` always align with upstream. What do you think?

trung (changes requested):
We don't need to introduce new API. 
And we need to make the current one backward compatible.

The convention is that if the argument is a pointer, it's considered to be optional. Similar to `eth_storageRoot` API.

zzy96:
Updated to make it backward compatible.

### Scenario Tested

Started a network with a few public & private transaction

**use geth console**

- :white_check_mark: Use `debug.dumpBlock("latest")` to get a public state dump at latest
- :white_check_mark: Use `debug.dumpBlock("0xf")` to get a public state dump at a specific block
- :white_check_mark: Use `debug.dumpBlock("latest", "private")` to get a private state dump at latest
- :white_check_mark: Use `debug.dumpBlock("0xf", "private")` to get a private state dump at a specific block
- :white_check_mark: Use `debug.dumpBlock("latest", "random")` with a random string, default to get a public state dump at latest

**use rpc endpoint**

- :white_check_mark: Use `curl -X POST --data '{"jsonrpc":"2.0","method":"debug_dumpBlock","params":["latest"],"id":99}' -H "Content-Type: application/json" http://localhost:22000` to get a public state dump at latest
- :white_check_mark: Use `curl -X POST --data '{"jsonrpc":"2.0","method":"debug_dumpBlock","params":["0xf"],"id":99}' -H "Content-Type: application/json" http://localhost:22000` to get a public state dump at a specific block
- :white_check_mark: Use `curl -X POST --data '{"jsonrpc":"2.0","method":"debug_dumpBlock","params":["latest", "private"],"id":99}' -H "Content-Type: application/json" http://localhost:22000` to get a private state dump at latest
- :white_check_mark: Use `curl -X POST --data '{"jsonrpc":"2.0","method":"debug_dumpBlock","params":["0xf", "private"],"id":99}' -H "Content-Type: application/json" http://localhost:22000` to get a private state dump at a specific block
- :white_check_mark: Use `curl -X POST --data '{"jsonrpc":"2.0","method":"debug_dumpBlock","params":["latest", "random"],"id":99}' -H "Content-Type: application/json" http://localhost:22000` with a random string, default to get a public state dump at latest

trung (approved):
Good work. LGTM

jbhurat (approved):
LGTM

jpmsam (approved):

PR #945, opened 2020-02-20
Bloom filter fixes

zzy96:
If we try to query a private transaction log from a specific block (query filter with block hash) using `eth_getLogs` endpoint, the request will fail because the private block bloom is not available. This issue is fixed in this PR.


trung (commented):
Are there tests covering this bug?

trung (approved):
LGTM

jpmsam (approved):

Issue #942, opened 2020-02-14
Bad block error when syncing blocks on new nodes after increasing maxCodeSize


vdamle:
#### System information

Geth version: `1.8.18`
OS & Version: Linux
Branch, Commit Hash or Release: Quorum 2.3.0

#### Expected behaviour

New nodes added to a chain which is reconfigured to run with `maxCodeSize: 128` should be able to sync blocks from existing nodes, which may have had failed contract deployment txn due to size exceeding the previous `maxCodeSize`

#### Actual behaviour

New nodes reject blocks from existing nodes which have a failed contract deployment txn with `BAD BLOCK` error with reason: `Error: invalid gas used`

#### Steps to reproduce the behaviour

* Start a quorum chain with default `maxCodeSize` (no overriding genesis config)
* Deploy contract which exceeds the code size - this is rejected by nodes as expected
* Reconfigure the chain by adding `maxCodeSize` in genesis config (ex: 128kb) and `geth init` all nodes to reinitialize genesis. Now the chain runs fine and can accept contracts upto the new `maxCodeSize`
* Add a new node to the chain. This node will be configured with `maxCodeSize: 128` from the start. When this node syncs blocks from the chain, it rejects the block with the failed contract deployment. Below is an example log and transaction receipt for the failed txn:

```
ERROR[02-10|16:37:32.196]
########## BAD BLOCK #########
Chain config: {ChainID: 649600883 Homestead: 0 DAO: <nil> DAOSupport: false EIP150: 0 EIP155: 0 EIP158: 0 Byzantium: 0 IsQuorum: true Constantinople: <nil> Engine: istanbul}
Number: 12851
Hash: 0x6f6aa1ed44525cae2fa02d5732238a4f5c163ec2452517f47c7b4f34021245d7
[...]

PR #947, opened 2020-02-20
maxCodeSizeChangeBlock to genesis for tracking maxCodeSize value change

vsmk98:
Added `maxCodeSizeChangeBlock` to track the block number when the `maxCodeSize` changes from default 24K to the value given.
For long standing networks, it is possible that contracts which crossed default 24k size were deployed at some time and the network rejected the same with `max code size exceeded` error. If the network at later point changed the maxCodeSize to a higher value and post this if a new node joins the network, the new node will sync the old blocks with the higher max code size value and thus the gas computed on the historical block will be different from the network resulting in bad block error. 
`maxCodeSizeChangeBlock` has been added track the block number where the new `maxCodeSize` value was set for the network overwriting the default value of 24K.  This will ensure that any new nodes joining the network will be able to sync the historical blocks without the bad block error. 

However, it should be noted that, if `maxCodeSize` has been changed several times at network level, `maxCodeSizeChangeBlock` will only track the last change and for all other prior blocks it will take 24K as the default `maxCodeSize`. In this scenario, its possible to get the bad blocks error at the time of network sync

#942

### DIFF
--- a/core/vm/evm.go
+++ b/core/vm/evm.go
@@ -488,7 +488,7 @@ func (evm *EVM) create(caller ContractRef, codeAndHash *codeAndHash, gas uint64,
 	ret, err := run(evm, contract, nil, false)
 
 	var maxCodeSize int
-	if evm.ChainConfig().MaxCodeSize > 0 {
+	if evm.chainConfig.IsMaxCodeSizeChangeBlock(evm.BlockNumber) && evm.ChainConfig().MaxCodeSize > 0 {
 		maxCodeSize = int(evm.ChainConfig().MaxCodeSize * 1024)
 	} else {
 		maxCodeSize = params.MaxCodeSize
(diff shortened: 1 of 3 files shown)


jbhurat (approved):
LGTM

jpmsam:
>if maxCodeSize has been changed several times at network level, maxCodeSizeChangeBlock will only track the last change and for all other prior blocks it will take 24K as the default maxCodeSize.

While this PR fix's the immediate issue, we'll need to look into handling possible multiple maxCodeSizes on the network.

jimthematrix:
thanks for making this update @vsmk98, we already have a customer that was hit by the issue and can benefit from the enhanced maxCodeSize configuration.

@jpmsam definitely agree having support for multiple settings each tied to a block number would be a very useful next step.

jiakun9707:
Why not do the same tracking for the transaction size limit?

PR #951, opened 2020-03-09
hotfix for Istabul BFT consensus block header Time issue

ZJJ1991:
Because of hardwares(node) time deviation, the block header's "Time" calculated by proposer may be bigger than that of other validator nodes. Given this, here, 1 hour deviation is allowed.


ZJJ1991:
This always cause unnecessary "block in the future" error.

jpmsam:
Thanks for the contribution @ZJJ1991 . We are currently looking into making this configurable similar to PR #800 with a smaller default value.

PR #956, opened 2020-03-18
plugin: delegate JSON RPC calls to plugin

trung:
By providing `apiProviderFunc` implementation, plugin can expose their APIs
via internal plugin interfaces to geth JSON RPC by convention:

`plugin@<plugin interface name>_<api>`

E.g.: `plugin@helloworld_greeting`
`plugin@helloworld` namespace must be provided via `--rpcapi` or `--wsapi` arguments

### DIFF
--- a/plugin/service.go
+++ b/plugin/service.go
@@ -7,7 +7,6 @@ import (
 	"unsafe"
 
 	"github.com/ethereum/go-ethereum/log"
-
 	"github.com/ethereum/go-ethereum/p2p"
 	"github.com/ethereum/go-ethereum/rpc"
 )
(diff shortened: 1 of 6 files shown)


plugin/service.go
```diff
@@ -82,6 +65,12 @@ func (s *PluginManager) getPlugin(name PluginInterfaceName) (managedPlugin, bool
 	return p, ok
 }
 
+// Check if a plugin is enabled/setup
+func (s *PluginManager) IsEnabled(name PluginInterfaceName) bool {
```
jbhurat, line 71: Is this function being used. I don't see any references of this

trung: ah, this is a place holder. It will be used in the coming PRs.

jbhurat: Sounds good

jbhurat (dismissed):
LGTM

chris-j-h (approved):

jpmsam (approved):

PR #960, opened 2020-03-25
Quorum geth upgrade to 1.9.7

vsmk98:
PR for upgrading the base geth version from 1.8.18 to 1.9.7


cmd/puppeth/genesis_test.go
```diff
@@ -0,0 +1,110 @@
...
+				t.Logf("got: %v\nexp: %v\n", exp[i], got[i])
+			}
+		}
+	}
+}
+
+// Tests the go-ethereum to Parity chainspec conversion for the Stureby testnet.
+//Quorum - skip this test as MinGasLimit and GasLimitBoundDivisor has been overriden for quorum
+func testParitySturebyConverter(t *testing.T) {
+	blob, err := ioutil.ReadFile("testdata/stureby_geth.json")
```
trung, line 79: Consider to use `t.Skipf()` instead

cmd/puppeth/genesis_test.go
```diff
@@ -0,0 +1,110 @@
...
+	"strings"
+	"testing"
+
+	"github.com/davecgh/go-spew/spew"
+	"github.com/ethereum/go-ethereum/core"
+)
+
+// Tests the go-ethereum to Aleth chainspec conversion for the Stureby testnet.
+//Quorum - skip this test as MinGasLimit and GasLimitBoundDivisor has been overriden for quorum
+func testAlethSturebyConverter(t *testing.T) {
```
trung: Consider to use `t.Skipf()` instead

cmd/utils/flags_test.go
```diff
@@ -5,6 +5,8 @@ import (
 	"io/ioutil"
```
trung, line 22: License header is missing

params/network_params.go
```diff
@@ -51,4 +46,37 @@ const (
...
+	// CheckpointProcessConfirmations is the number before a checkpoint is generated
+	CheckpointProcessConfirmations = 256
+
+	// ImmutabilityThreshold is the number of blocks after which a chain segment is
+	// considered immutable (i.e. soft finality). It is used by the downloader as a
+	// hard limit against deep ancestors, by the blockchain against deep reorgs, by
+	// the freezer as the cutoff treshold and by clique as the snapshot trust limit.
+	ImmutabilityThreshold = 90000
 )
+
```
trung, line 62: Please review the below code to avoid package level state.

Suggestion:`immutabilityThresHold` could be moved to structs that need it.

core/call_helper.go
```diff
@@ -2,6 +2,7 @@ package core
 
 import (
 	"crypto/ecdsa"
+	"github.com/ethereum/go-ethereum/core/rawdb"
```
trung: Please run `goimports`

core/dual_state_test.go
```diff
@@ -1,20 +1,20 @@
 package core
 
 import (
+	"github.com/ethereum/go-ethereum/core/rawdb"
```
trung: Please run `goimports`

eth/downloader/peer.go
```diff
@@ -22,6 +22,7 @@ package downloader
 import (
 	"errors"
 	"fmt"
+	"github.com/ethereum/go-ethereum/consensus"
```
trung: Please run `goimports`

graphql/graphql.go
```diff
@@ -0,0 +1,1111 @@
...
+	ret := false
+	tx, err := t.resolve(ctx)
+	if err != nil || tx == nil {
+		return &ret, err
+	}
+	ret = tx.IsPrivate()
+	return &ret, nil
+}
+
+func (t *Transaction) PrivateInputData(ctx context.Context) (*hexutil.Bytes, error) {
```
trung, line 355: As we already have `IsPrivate` to distinguish between public and private transaction, why not reuse `InputData` instead of introducing a new one? It would align with internal representation of a transaction.

internal/ethapi/api.go
```diff
@@ -1333,20 +1513,54 @@ func (args *SendTxArgs) setDefaults(ctx context.Context, b Backend) error {
...
+	} else {
+		log.Info("nil args.input & args.data")
+	}
+
+	if len(input) > 0 {
+		var data []byte
+		var err error
+		if sendTxn {
+			//Send private transaction to local Constellation node
+			log.Info("sending private tx", "input", fmt.Sprintf("%x", input), "privatefrom", args.PrivateFrom, "privatefor", args.PrivateFor)
```
trung: Security concern. We must not log private payload!

rpc/client.go
```diff
@@ -402,66 +443,70 @@ func (c *Client) Subscribe(ctx context.Context, namespace string, channel interf
...
 		}
 	}
-	c.writeConn.SetWriteDeadline(deadline)
-	err := json.NewEncoder(c.writeConn).Encode(msg)
-	c.writeConn.SetWriteDeadline(time.Time{})
+	err := c.writeConn.Write(ctx, msg)
 	if err != nil {
 		c.writeConn = nil
+		if !retry {
+			return c.write(ctx, msg, true)
```
trung, line 491: - Retry should be handled externally rather than internally
- If there's a good reason doing so, this retry logic needs to be terminated conditionally to avoid stack overflow

rpc/service.go
```diff
@@ -0,0 +1,286 @@
...
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+package rpc
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"github.com/ethereum/go-ethereum/params"
```
trung: Please run `goimports`

signer/core/api.go
```diff
@@ -449,23 +476,49 @@ func logDiff(original *SignTxRequest, new *SignTxResponse) bool {
...
 		err    error
 		result SignTxResponse
+		msgs   *ValidationMessages
 	)
-	msgs, err := api.validator.ValidateTransaction(&args, methodSelector)
-	if err != nil {
-		return nil, err
+	if args.IsPrivate {
+		msgs = new(ValidationMessages)
+		msgs.Info(fmt.Sprintf("Quorum private transaction. Key: %s", args.Data))
```
trung: Is it necessary to leak private payload to validation message?

rpc/client.go
trung: Per logic, there are only 2 attempts. And this aligns with upstream fix

cmd/puppeth/genesis_test.go
vsmk98: Changes done to use t.Skipf().

cmd/puppeth/genesis_test.go
vsmk98: Changes done to use t.Skipf().

graphql/graphql.go
vsmk98: The reason we segregated it is to call to `InputData` returns the tessera hash and `PrivateInputData` returns the actual payload if the node is part of private transaction. This is inline with what we do currently where in `eth_getTransaction` for a node which is part of private transaction returns `input` which is the tessera hash and the using the hash I can make a call to `eth_getQuorumPayload` to get the actual pay load. Pl let me know your thoughts.

internal/ethapi/api.go
vsmk98: Agree removed payload from the log message. Also changed the message to a `Debug` message instead of `Info`.

rpc/service.go
vsmk98: Done

signer/core/api.go
vsmk98: Done

cmd/utils/flags_test.go
vsmk98: Done

jpmsam (approved):

PR #968, opened 2020-04-09
Created special PTM for public-only mode.

prd-fox:
Create subclass of PrivateTransactionManager that returns a not-in-use error on all of the functions. This is used when the node is set to run in public-only mode (`PRIVATE_CONFIG=ignore`)


trung (dismissed):
Just an note that we can use `testify/assert` package to simplify the assertion logic in tests.

LGTM

jpmsam (approved):

PR #971, opened 2020-04-14
Changes to allow ImmutabilityThreshold to be configured at network level

vsmk98:
The PR allows immutability threshold to be passed as  cli parameter when the node is started.


eth/downloader/downloader.go
```diff
@@ -209,10 +209,16 @@ type BlockChain interface {
 
 	// InsertReceiptChain inserts a batch of receipts into the local chain.
 	InsertReceiptChain(types.Blocks, []types.Receipts, uint64) (int, error)
+
+	// Config returns the block chain configuration
+	Config() *params.ChainConfig
```
zzy96: I didn't see any usage of Config()

params/network_params.go
```diff
@@ -59,3 +59,23 @@ const (
 	// the freezer as the cutoff treshold and by clique as the snapshot trust limit.
 	ImmutabilityThreshold = 90000
 )
+
+// //Quorum
+var QuorumImmutabilityThreshold int
+var IsQuorum = false
```
zzy96: We may define ImmutabilityThreshold as var and directly modify it. Although this means we have to modify this network_params file but we can avoid changes in all other files calling to `params.ImmutabilityThreshold`.

params/network_params.go
```diff
@@ -59,3 +59,23 @@ const (
...
+	if IsQuorum {
+		return QuorumImmutabilityThreshold
+	}
+	return ImmutabilityThreshold
+}
+
+// sets the immutability threshold and isQuorum to passed values
+func SetQuorumImmutabilityThreshold(immutabilityThreshold int, isQuorum bool) {
+	QuorumImmutabilityThreshold = immutabilityThreshold
+	IsQuorum = isQuorum
```
zzy96: `SetQuorumImmutabilityThreshold` is called under the condition of `chainConfig.IsQuorum` is true in `eth.New`. Therefore we can directly set IsQuorum to true instead of passing it.

eth/downloader/downloader.go
vsmk98: Yes, had added it when I wanted to check from config if it is Quorum. Code changes done.

params/network_params.go
vsmk98: Want to keep these two distinct and hence added a new variable.

params/network_params.go
vsmk98: Agree. Code changes done.

params/network_params.go
jbhurat: Agree with @vsmk98 here, we should stay as close to upstream as we can, so that we don't have merge conflicts down the line

jbhurat: Also, as `QuorumImmutabilityThreshold` and `IsQuorum` are only accessed using functions `GetImmutabilityThreshold()` and `SetQuorumImmutabilityThreshold()`, there is no need for them to be exported.

vsmk98: Agree Jitu. Have pushed the code changes.

jbhurat (commented):
The changes look good. One small change, can you run `goimports` on the files modified

vsmk98:
ran `goimports` on all files.

jbhurat (approved):
LGTM

vsmk98:
merged with 1.9.7 branch.

PR #989, opened 2020-05-06
eth_estimateGas failing when value is not passed as a part of argumenst

vsmk98:
when `eth.estimateGas` is called with any arguments as shown below, the call fails with an error. 
```
> eth.estimateGas({from: eth.accounts(0)})
Error: method handler crashed
    at web3.js:3143:20
    at web3.js:6347:15
    at web3.js:5081:36
    at <anonymous>:1:1
```
This PR fixes the above error.


jbhurat (dismissed):
LGTM

jpmsam (approved):

PR #1017, opened 2020-06-12
Add  'GO111MODULE=off' comment

icodezjb:
about https://github.com/jpmorganchase/quorum/issues/909

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -3,6 +3,11 @@
 
 Note: Building Quorum requires both a Go (version 1.9 or later) and a C compiler. You can install them using your favourite package manager. 
 
+If Go version >= 1.11
+```
+go env -w GO111MODULE=off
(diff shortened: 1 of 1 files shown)


jbhurat:
Hi @icodezjb, I have tried building `quorum` both inside and outside of `GOPATH` and not having `GO111MODULE=off` and it built just fine using golang `1.11.4` and `1.13.12`. Have you tried using a different variant of linux to see if the issue is not related to `deepin linux`

icodezjb:
> Hi @icodezjb, I have tried building `quorum` both inside and outside of `GOPATH` and not having `GO111MODULE=off` and it built just fine using golang `1.11.4` and `1.13.12`. Have you tried using a different variant of linux to see if the issue is not related to `deepin linux`

Thank you for your feedback

PR #706, opened 2019-05-09
Add JSON-RPC TLS support

denny60004:
[Quorum] TLS Support On Quorum

PR #1019, opened 2020-06-15
plugin/security: protect JSON RPC APIs via a plugin interface

trung:
* Define a security plugin interface
* Protect JSON RPC servers: HTTP and WS with TLS and access controls
* Support connectivity to a protected JSON RPC server in
   * ethclient library
   * geth attach CLI
* Support JSON Batch RPC: return security errors corresponding to
requests

Related to #693 #706


vsmk98 (dismissed):
LGTM

jpmsam (approved):

PR #1026, opened 2020-07-02
cmd/geth: exit gradually when there's a failure in PluginManager

trung:
p2p server handles peer connection before Ethereum service is actually started.
By having PluginManager service (or any service) starts before Ethereum service causing peer
connections being held and geth is unable to exit gradually when
PluginManager service (or any service) returns error.

### DIFF
--- a/cmd/geth/config.go
+++ b/cmd/geth/config.go
@@ -25,15 +25,14 @@ import (
 	"reflect"
 	"unicode"
 
-	"github.com/naoina/toml"
-	"gopkg.in/urfave/cli.v1"
-
 	"github.com/ethereum/go-ethereum/cmd/utils"
 	"github.com/ethereum/go-ethereum/dashboard"
 	"github.com/ethereum/go-ethereum/eth"
 	"github.com/ethereum/go-ethereum/node"
 	"github.com/ethereum/go-ethereum/params"
 	whisper "github.com/ethereum/go-ethereum/whisper/whisperv6"
+	"github.com/naoina/toml"
+	"gopkg.in/urfave/cli.v1"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


chris-j-h (changes requested):
The startup order of services is not guaranteed as the iteration is [performed on a map](https://github.com/jpmorganchase/quorum/blob/master/node/node.go#L231) (although this change does make it more likely for the `PluginManager` to be started after the `Ethereum` service).

Adding `log.Debug("Starting service", "service", kind.String())` to https://github.com/jpmorganchase/quorum/blob/master/node/node.go#L232 and starting a network a handful of times I observed different start orders:

```
DEBUG[07-03|13:23:39.257] Starting service                         service=*eth.Ethereum
DEBUG[07-03|13:23:39.257] Starting service                         service=*plugin.PluginManager
```

```
DEBUG[07-03|13:27:17.980] Starting service                         service=*raft.RaftService
DEBUG[07-03|13:27:18.004] Starting service                         service=*eth.Ethereum
DEBUG[07-03|13:27:18.004] Starting service                         service=*plugin.PluginManager
```

```
DEBUG[07-03|13:30:31.382] Starting service                         service=*plugin.PluginManager
```

The last case was observed the least number of times, but still results in the node being unable to exit gracefully.  

Perhaps we should explicitly hold off starting the `PluginManager` until all other services have been started, or explicitly start the `Ethereum` service before all others?

trung:
Good catch. The order is only maintained during service contructors are called. But not during start time. I think this could be a problem if there are other services being started before Ethereum service and failed.

Best is to make sure the order is same during startup. I'll raise a PR upstream to get additional feedback

vsmk98:
Agree with the observations here. I had similar observation where `plugin` started even before `ethereum` service started and hence node shut down never happened even though `plugin` `start` returned error.

trung:
Raised https://github.com/ethereum/go-ethereum/issues/21292

jpmsam (approved):

PR #1030, opened 2020-07-06
Private contract state extension

vsmk98:
Once a private contract is deployed, it is only available on the nodes where it was initially deployed. Any nodes which are not part of initial contract deployment or new nodes joining the network will not have access to private contract. This PR addresses this issue and allows contract state to be extended to nodes that were not part of original contract deployment.


trung (approved):
LGTM

jpmsam (approved):

PR #1033, opened 2020-07-08
Support additional methods of account management with account plugins

chris-j-h:
This change adds support for account plugins to quorum and clef.  Account plugins can be used to provide alternative methods of managing accounts to the standard `keystore` file-based accounts.

Changes include:
* adds new `accounts.Pluggable` wallet and backend
* adds `geth account plugin` CLI commands
* enables usage of all existing account/signing-related APIs with plugin accounts (excluding `personal_newAccount` and `personal_importRawKey`)
* adds `plugin@account_newAccount` and `plugin@account_importRawKey` APIs
* extends `js` console with `plugin_account.newAccount` and `plugin_account.importRawKey`
* centralises account/key management docs (renames `Safeguarding Private Keys` section to `Account/Key Management`)


internal/ethapi/api.go
```diff
@@ -344,23 +345,55 @@ func (s *PrivateAccountAPI) UnlockAccount(ctx context.Context, addr common.Addre
 	} else {
 		d = time.Duration(*duration) * time.Second
 	}
-	ks, err := fetchKeystore(s.am)
```
vsmk98, line 347: Please merge the branch with latest master.

internal/ethapi/api.go
```diff
@@ -344,23 +345,55 @@ func (s *PrivateAccountAPI) UnlockAccount(ctx context.Context, addr common.Addre
...
+	backend, err := s.am.Backend(acct)
+	if err != nil {
+		return err
+	}
+
+	switch b := backend.(type) {
+	case *pluggable.Backend:
+		return b.TimedUnlock(acct, password, duration)
+	case *keystore.KeyStore:
+		return b.TimedUnlock(acct, password, duration)
```
vsmk98, line 367: please add a `case` for `default` and return an error message. Otherwise for external signers like clef, this function will return `true` though the unlock operation would not have happened

internal/ethapi/api.go
```diff
@@ -344,23 +345,55 @@ func (s *PrivateAccountAPI) UnlockAccount(ctx context.Context, addr common.Addre
...
+	backend, err := s.am.Backend(acct)
+	if err != nil {
+		return err
+	}
+
+	switch b := backend.(type) {
+	case *pluggable.Backend:
+		return b.Lock(acct)
+	case *keystore.KeyStore:
+		return b.Lock(addr)
```
vsmk98, line 395: same as `unlockAccount`. Handle the default condition for any other type of signer and return error

vsmk98 (changes requested):

internal/ethapi/api.go
chris-j-h: Resolved by https://github.com/jpmorganchase/quorum/pull/1033/commits/e9d59669f6f34afcd2f3a50e6f3997ec2d2d7925

internal/ethapi/api.go
chris-j-h: Resolved by https://github.com/jpmorganchase/quorum/pull/1033/commits/e9d59669f6f34afcd2f3a50e6f3997ec2d2d7925

chris-j-h:
I observed an intermittent panic when starting clef with account plugins (approx 1 in 10 times).

I have included a fix with https://github.com/jpmorganchase/quorum/pull/1033/commits/d8977eb55b2a9b9815fa99b18f35c94b5a40b300 and have not been able to reproduce the error since.  However, given the intermittent nature of the issue I will document it here for future reference and in case others encounter it:

#### Observation
The following panic when starting `clef --plugins plugin-config.json --plugins.skipverify` (and after `ok`ing at the startup prompt):

```shell
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x48 pc=0x460413a]

goroutine 47 [running]:
github.com/ethereum/go-ethereum/accounts/pluggable.(*wallet).Open(0xc000172040, 0x0, 0x0, 0x1, 0x0)
        .../go/src/github.com/ethereum/go-ethereum/accounts/pluggable/wallet.go:41 +0x2a
github.com/ethereum/go-ethereum/signer/core.(*SignerAPI).startUSBListener.func1(0xc00248e000, 0xc000bd8230, 0xc000160c00)
        .../go/src/github.com/ethereum/go-ethereum/signer/core/api.go:339 +0xa6
created by github.com/ethereum/go-ethereum/signer/core.(*SignerAPI).startUSBListener
        .../go/src/github.com/ethereum/go-ethereum/signer/core/api.go:335 +0xa4
```

#### Cause

As `--nousb` was not provided, clef starts a `USBListener` goroutine.  Before the fix, this goroutine was being created before the plugin wallet setup was completed.  The `USBListener` attempts to open ALL wallets.  As goroutine execution order is not guaranteed, sometimes this goroutine would be scheduled BEFORE the plugin wallet was ready, resulting in a nil pointer dereference.

#### Fix 
The fix ensures the plugin wallet is setup before creating the `USBListener` goroutine.  If the error persists, including the `--nousb` flag should also provide a fix.

trung (dismissed):
LGTM

vsmk98 (dismissed):
LGTM

trung (approved):

jpmsam (approved):

PR #1034, opened 2020-07-08
Fix punctuation

rdimaio:


jpmsam:
We'll make the edit when we update the documentation.

PR #1036, opened 2020-07-14
Create separate sub-module for crypto/secp256k1 package

chris-j-h:
Once tagged, this will allow external consumers of the crypto/secp256k1 package to depend only on this package and not the entire quorum codebase


trung (approved):
LGTM

PR #1042, opened 2020-07-22
Error out all header verification if one of the header errors out

jbhurat:
`VerifyHeaders()` verifies a batch of headers, while iterating over the headers a potential invalid header might be used to get the snapshot. The fix here is to error out all the subsequent header verification request, when the first one fails


vsmk98 (approved):
LGTM

PR #1046, opened 2020-07-29
private: formalize the usage of common.EncryptedPayloadHash

trung:
`PrivateTransactionManager` interface methods represent API interaction with a private transaction manager. Using `common.EncryptedPayloadHash` enhances the understandability.

### DIFF
--- a/private/privatetransactionmanager/tx_manager.go
+++ b/private/privatetransactionmanager/tx_manager.go
@@ -15,46 +15,50 @@ type PrivateTransactionManager struct {
 	c    *gocache.Cache
 }
 
-func (g *PrivateTransactionManager) Send(data []byte, from string, to []string) (out []byte, err error) {
-	out, err = g.node.SendPayload(data, from, to)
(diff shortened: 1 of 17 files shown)


private/privatetransactionmanager/tx_manager.go
```diff
@@ -15,46 +15,50 @@ type PrivateTransactionManager struct {
...
 		return nil, err
 	}
 	return out, nil
 }
 
-func (g *PrivateTransactionManager) Receive(data []byte) ([]byte, error) {
-	if len(data) == 0 {
-		return data, nil
+func (g *PrivateTransactionManager) Receive(txHash common.EncryptedPayloadHash) ([]byte, error) {
+	if common.EmptyEncryptedPayloadHash(txHash) {
```
nicolae-leonte-go, line 49: I'm not sure that returning txHash.Bytes() is right here... 
An empty transaction.data field is converted to a 0 filled array (which is somewhat different from the previous behaviour).

nicolae-leonte-go (approved):

jpmsam (approved):

PR #1058, opened 2020-08-21
Add documentation to clarify the use of bindingAddress in tessera config

### LINKED ISSUE #807: Move the advertised address out of server configuration
Currently, the URL that is advertised to other peers with our public keys is taken from the P2P server configuration, as the `serverAddress` property.
In order to be able to bind to a different address, the `bindingAddress` property was introduced.

Since the advertised address isn't directly linked to the server configuration (although there is an indirect link to the `P2P` server), it could make sense to move this "advertised address" to a node level property instead of server level. Then, we can remove the distinction between `serverAddress` and `bindingAddress`, only needing one of them (which would be the `bindingAddress`).

This would overall reduce the confusion that can arise between the properties, whilst keeping the server setup distinct from the rest of the node.

SatpalSandhu61:
Fixes jpmorganchase/tessera#807

Clarifies use of bindingAddress in Tessera server configuration.

### DIFF
--- a/docs/Privacy/Tessera/Configuration/Configuration Overview.md
+++ b/docs/Privacy/Tessera/Configuration/Configuration Overview.md
@@ -104,6 +104,10 @@ The servers to be started are provided as a list:
 ]
 ```
 
+Note that there are two configuration entries relating to the server address:
+- `serverAddress` - In normal use, you will only need this (e.g. `http://localhost:9001`).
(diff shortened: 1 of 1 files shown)


Krish1979 (approved):

SatpalSandhu61:
Closing as the checks haven't run for some reason. Will recreate.

Issue #1063, opened 2020-08-28
Raft: Modify to generate deterministic raftIds, remove required `--raftjoinexisting $raftId` flag


libby:
Currently, raftIds are variable,  and sometime depended on the order of the `static-nodes.json` file, while other times they need to be obtained from an outside source. This means that a new node cannot determine its own `$raftId` without being explicitly passed the information via the geth startup param (`--raftjoinexisting $raftId`) at startup.

This is overly complex, and makes automating operations (adding / removing nodes) difficult for projects like [ConsenSys qubernetes](https://github.com/ConsenSys/qubernetes) which supports easily creating, deploying and managing Consensys Quorum networks on Kubernetes. Related to add nodes PR in qubernetes / k8s https://github.com/ConsenSys/qubernetes/pull/60 

### Starting a New Network

*  The RaftId is obtained from the order of the `static-nodes.json` file when starting the network for the first time. 
*  **issue**: this means that every node in the network must have the same exact `static-nodes.json` file with the same exact order. This only holds at bootup time, after bootup  the `static-nodes.json` is not no longer the raftId recorded of truth. 
* **update request**: the raftId should not be dependent on the order of the `static-nodes.json` file, but should be obtained deterministically by the nodes.

### Adding a New Node To Raft Network

1. Add the new node from a node in the current cluster by running `raft.addPeer($enodeId)` and obtain the raftId that is returned (this id is not deterministic but generated by the node the command was executed on, it is supposed to be the next monotonically increasing int of all known nodes in the cluster, the command can only be called once).

2. Take the $raftId returned in step 1 and add the `--raftjoinexsiting $raftId` to the geth startup params for the node that is being added. 

3. Additionally when removing a node and re adding it, it will receive a new `$raftId` and the old `$raftId` can never be used again, this is a current requirement as the id are monotonically increasing, and need to ensure uniqueness.

**Issues**: This has many drawbacks for operations, and makes adding/removing raft nodes overly complex, e.g. deployment configs will be constantly in need of updating . The difficulty comes from managing the state outside of the Quorum node itself, and requiring different config params dependent on whether it is is an initial node in the network or a new node joining the network, or a node that has been added, removed and added again. A new node has no way of knowing its `$raftId` apart from obtaining it from the passed in flag `--raftjoinexsiting $raftId`.

[...]

PR #1071, opened 2020-09-08
[WIP] Make Raft id deterministic, remove  required `--raftjoinexisting $raftId`

libby:
The purpose of this PR is to address issue 1063  https://github.com/ConsenSys/quorum/issues/1063 "Raft: Modify to generate deterministic raftIds, remove required `--raftjoinexisting $raftId` flag "
to simplify adding / removing and managing nodes in a raft Quorum network. 

## Current Raft Flow Adding a Node (before this PR)

The current raft id is a monotonically increasing number that is calculated initially by the index of the nodes in the initial `static-nodes.json` file. The raft id being tied to the node's position in the `static-nodes.json` only holds when the network is being started for the first time, after that,  to add a new node to a quorum network the nodes must:
1. be added to a permisioned-nodes.json file on a cluster node.
2. an operator must run `raft.addPeer(enodeurl)` on a healthy node in the cluster and obtain the returned id (they cannot know this ahead of time).
3. the flag `--raftjoinexisting $RAFTID` must be included in the start up params of the node being added, and the id obtained from the operator must be set in the startup params.

The raftId for a node is not sticky, and if a raft node is removed from the cluster and re-added it must obtain a new raftId. Raft id are never reused to avoid race-conditions, and to compute a unique id when adding other nodes. This divergence of startup parameter makes make it difficult to mange a raft quorum network, e.g. from projects like [qubernetes](https://github.com/ConsenSys/qubernetes), and from other ops standpoints, as the node parameters as constantly changing, and values have to be obtained outside of the node before starting it up.

This PR makes the raftId deterministic, specifically it changes the raftid from a monotonically increasing `uint16` to a deterministic `uint64` Keccack hash of the enode id of each node. 
 
## Changes To Raft Id / Raft Flow

This PR changes the way that the raft id is calculated and greatly simplifies adding / removing and managing raft quorum nodes / network. 

* Each node can now calculate its own raft id as it the Keccack hash of its enode id.
* When a node is removed and re-added it will keep the same raft id.
* The order of the nodes (when initializing a quorum network) in the `static-nodes.json` is no longer important, as the order does not play a role in the calculating the raft id. 
* Removes the need for the  `--raftjoinexisting $raftid` startup parameter for nodes that are being added to the network.
* All nodes can have the same geth startup parameters, there is no difference between the params of an initial node and subsequent nodes being added to the network. 
* Collision are avoided due to the hash properties although it is not 256 but uint64, it should suffice for a reasonable sized raft network up to thousands of nodes.

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -1841,6 +1841,7 @@ func RegisterPermissionService(stack *node.Node) {
 func RegisterRaftService(stack *node.Node, ctx *cli.Context, nodeCfg *node.Config, ethChan chan *eth.Ethereum) {
 	blockTimeMillis := ctx.GlobalInt(RaftBlockTimeFlag.Name)
 	datadir := ctx.GlobalString(DataDirFlag.Name)
+	// TODO: libby make this a boolean or remove it.
 	joinExistingId := ctx.GlobalInt(RaftJoinExistingFlag.Name)
 	useDns := ctx.GlobalBool(RaftDNSEnabledFlag.Name)
 	raftPort := uint16(ctx.GlobalInt(RaftPortFlag.Name))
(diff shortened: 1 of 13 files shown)


robinbryce:
What happened to this feature ? We were very much looking forward to it!

cmd/utils/flags.go
```diff
@@ -1845,43 +1844,26 @@ func RegisterRaftService(stack *node.Node, ctx *cli.Context, nodeCfg *node.Confi
...
-				peerIds[peerIdx] = peerId
-				if peerId == strId {
-					myId = uint16(peerIdx) + 1
-				}
-			}
-
-			if myId == 0 {
-				Fatalf("failed to find local enode ID (%v) amongst peer IDs: %v", strId, peerIds)
-			}
+			Fatalf("Raft-based consensus requires either an initial peers list including all the nodes in the network (in static-nodes.json) including this enode hash (%v)", strId)
```
macfarla, line 1860: this is no longer an either-or statement. Suggest "requires an initial peers list"

raft/minter_test.go
```diff
@@ -19,10 +19,25 @@ import (
...
+const ENODE_ID_NODE3 = "579f786d4e2830bbcc02815a27e8a9bacccc9605df4dc6f20bcc1a6eb391e7225fff7cb83e5b4ecd1f3a94d8b733803f2f66b7e871961e7b029e22c155c3a778"
+const ENODE_ID_NODE4 = "3d9ca5956b38557aba991e31cf510d4df641dce9cc26bfeb7de082f0c07abb6ede3a58410c8f249dabeecee4ad3979929ac4c7c496ad20b8cfdd061b7401b4f5"
+const ENODE_ID_NODE5 = "3701f007bfa4cb26512d7df18e6bbd202e8484a6e11d387af6e482b525fa25542d46ff9c99db87bd419b980c24a086117a397f6d8f88e74351b41693880ea0cb"
+const ENODE_ID_NODE6 = "eacaa74c4b0e7a9e12d2fe5fee6595eda841d6d992c35dbbcc50fcee4aa86dfbbdeff7dc7e72c2305d5a62257f82737a8cffc80474c15c611c037f52db1a3a7b"
+const ENODE_ID_NODE7 = "239c1f044a2b03b6c4713109af036b775c5418fe4ca63b04b1ce00124af00ddab7cc088fc46020cdc783b6207efe624551be4c06a994993d8d70f684688fb7cf"
 
 func TestSignHeader(t *testing.T) {
 	//create only what we need to test the seal
-	var testRaftId uint16 = 5
+	testRaftId := getRaftId(t, ENODE_ID_NODE5) // 5
```
macfarla, line 40: nit: NODE5 is hard to read - any reason not to go with just ENODE_ID_5

PR #1072, opened 2020-09-17
Add API for account dump & transaction trace fix

prd-fox:
A new API named `debug_dumpAddress` has been added in this PR, which fetches the state of an address at a given block height (including `pending` and `latest`). This includes it's code, balance, nonce, storage etc.
Note that the storage can get large, so the response can take some time. The request can also take some time due to needing to rebuild the contracts environment if the state isn't present, so if planning on using this API regularly  it is recommended to run the node in archive mode.

Also included is a bug fix for tracing transactions. If the transaction is public, a write operation would fail and abort the trace due to the private state being used instead of the public state. The public state is now set correctly in this case.

### DIFF
--- a/eth/api.go
+++ b/eth/api.go
@@ -296,27 +296,9 @@ func NewPublicDebugAPI(eth *Ethereum) *PublicDebugAPI {
 // DumpBlock retrieves the entire state of the database at a given block.
 // Quorum adds an additional parameter to support private state dump
 func (api *PublicDebugAPI) DumpBlock(blockNr rpc.BlockNumber, typ *string) (state.Dump, error) {
-	var publicState, privateState *state.StateDB
-	var err error
-	if blockNr == rpc.PendingBlockNumber {
-		// If we're dumping the pending state, we need to request
-		// both the pending block as well as the pending state from
-		// the miner and operate on those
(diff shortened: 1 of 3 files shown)


vsmk98 (dismissed):
LGTM

eth/api.go
```diff
@@ -325,6 +325,42 @@ func (api *PublicDebugAPI) DumpBlock(blockNr rpc.BlockNumber, typ *string) (stat
 	return publicState.RawDump(false, false, true), nil
 }
 
+// Quorum
+// DumpAddress retrieves the state of an address at a given block.
+// Quorum adds an additional parameter to support private state dump
+func (api *PublicDebugAPI) DumpAddress(address common.Address, blockNr rpc.BlockNumber) (state.DumpAccount, error) {
```
trung, line 313: There's an opportunity to extract the below duplicated code which does retrieval of states and reuse in both `DumpBlock` and `DumpAddress`.
`DumpBlock` also has Quorum modification footprints so it makes sense to do the refactoring.

prd-fox: Done, thanks!

trung (approved):
LGTM

jpmsam (approved):

PR #1073, opened 2020-09-22
Privacy Enhancements

nicolae-leonte-go:
Adds the "Privacy Enhancements" feature to quorum.

The feature can be enabled via the `privacyEnhancementsBlock` chain config entry in genesis.json (a value must be agreed at the network level and all parties must run `geth init` to configure the value).

A new privacyFlag parameter was added to the sendTransaction API. The possible values are:
* 0 - StandardPrivate
* 1 - PartyProtection
* 3 - StateValidation

If the parameter is missing the value is assumed to be 0 - StandardPrivate.

The sendTransaction APIs have been enhanced to perform transaction simulation before submitting the transaction to the transaction pool in order to extract the necessary metadata for PartyProtection and StateValidation. 

StandardPrivate - is the value identifying the private transactions quorum has handled in all it's previous releases. We've aimed to keep the StandardPrivate behavior backward compatible (regardless of whether privacy enhancements are enabled/disabled) - unless a StandardPrivate transaction/contract attempts to affect privacy enhanced contracts.

PartyProtection - during simulation we detect all the contracts that the transaction affects. We then attach the necessary proofs/metadata to the transaction payload so that the other parties are able to validate that the sending party has access to all the affected contracts.

StateValidation - includes all the PartyProtection checks. In addition participants for StateValidation transactions are checked against the participants for all of the affected contracts. After simulation an execution hash is calculated from the state of of the affected contracts and is attached to the private transaction payload (which is then verified by the receiving nodes after the transaction is minted).


core/genesis.go
```diff
@@ -248,9 +250,21 @@ func SetupGenesisBlockWithOverride(db ethdb.Database, genesis *Genesis, override
...
 	rawdb.WriteChainConfig(db, stored, newcfg)
+	if storedcfg.PrivacyEnhancementsBlock == nil && newcfg.PrivacyEnhancementsBlock != nil {
+		checkAndPrintPrivacyEnhancementsWarning(newcfg)
+	}
 	return newcfg, stored, nil
 }
 
+func checkAndPrintPrivacyEnhancementsWarning(config *params.ChainConfig) {
+	if config.PrivacyEnhancementsBlock != nil {
+		// TODO - Quorum - Privacy Enhancements - update the tessera version in the warning message
```
SatpalSandhu61, line 261: Note that this TODO still needs to be actioned

docs/Privacy/Privacy-Enhancements/Overview.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Private State Validation (PSV)
+
+This flavor uses all the verifications of counter party protection. On top of those it introduces further checks preventing nodes from state deviation by failing private transactions to 'self' or 'subset of nodes' by sharing the full list of recipients among all nodes which is validated against all subsequent transactions (in standard private/party protection mode only the sending node knows the list of recipients). 
+
+## Key Enhancements
+
+### Privacy Flag
+
+A new parameter `PrivacyFlag` has been added to all Quorum [send](http://docs.goquorum.com/en/latest/Getting%20Started/api/) API methods, being passed from the client to enable the privacy enhancements feature. This flag is an unsigned integer with the following values: 1 for PP (Party Protection) and 3 for PSV (Private State Validation) transactions. If the flag is missing or zero, the transaction is assumed to be a 'non-privacy enhanced' SP (Standard Private) transaction.
```
SatpalSandhu61, line 19: It might be worth mentioning that `privacyFlag` is ignored on public contracts.

trung (changes requested):
**Tollgate Note:** `.github/workflows/pr.yml` changes are to enable comprehensive acceptance tests for this PR. Those tests will be included in `quorum-acceptance-tests` repo. Before merging, please rollback the changes.

core/rawdb/database_quorum.go
```diff
@@ -49,10 +53,23 @@ func GetPrivateStateRoot(db ethdb.Database, blockRoot common.Hash) common.Hash {
...
+}
+
 func WritePrivateStateRoot(db ethdb.Database, blockRoot, root common.Hash) error {
 	return db.Put(append(privateRootPrefix, blockRoot[:]...), root[:])
 }
 
+func WritePrivacyMetadataStateRootForPrivateStateRoot(db ethdb.KeyValueWriter, privateStateRoot, privacyMetadataRoot common.Hash) error {
+	err := db.Put(append(privateRootToPrivacyMetadataRootPrefix, privateStateRoot[:]...), privacyMetadataRoot[:])
+	if err != nil {
+		return fmt.Errorf("unable to persist mapping between private state root to privacy metadata root. Cause: %v", err)
```
jbhurat: We should not wrap the error with the message here, instead just return whatever `db.Put()` returns. The calling function should provide helpful log if required

nicolae-leonte-go: fixed

mkdocs.yml
```diff
@@ -56,6 +55,9 @@ nav:
             - Contract Extension:
                 - Overview: Privacy/Contract-Extension/Overview.md
                 - APIs: Privacy/Contract-Extension/ContractExtension apis.md
+            - Privacy Enhancements:
+                - Overview: Privacy/Privacy-Enhancements/Overview.md
+                - APIs: Privacy/Privacy-Enhancements/Privacy Enhancement apis.md
```
chris-j-h: these changes aren't necessary as `docs/` has been deprecated

extension/privacyExtension/state_set_utilities.go
```diff
@@ -28,10 +28,27 @@ func setState(privateState *state.StateDB, accounts map[string]extension.Account
...
+// updates the privacy metadata
+func setPrivacyMetadata(privateState *state.StateDB, address common.Address, hash string) {
+	privacyMetaData, err := privateState.GetStatePrivacyMetadata(address)
+	if err != nil || privacyMetaData.PrivacyFlag.IsStandardPrivate() {
+		return
+	}
+
+	ptmHash, err := common.Base64ToEncryptedPayloadHash(hash)
+	if err != nil {
+		log.Debug("setting privacy metadata failed", "err", err)
```
jbhurat: Shouldn't this be logged as `Error`

cmd/evm/runner.go
```diff
@@ -217,6 +217,7 @@ func runCmd(ctx *cli.Context) error {
 	execTime := time.Since(tstart)
 
 	if ctx.GlobalBool(DumpFlag.Name) {
+		//TODO - Quorum - what is the point of doing both (commit does IntermediateRoot anyway)?
```
vsmk98: Please get rid of the `TODO` here. I know the call is redundant, but I would leave it there, keeping it aligned with upstream. Also this should not have any impact processing time

core/state_transition_pmh.go
```diff
@@ -0,0 +1,145 @@
...
+		// GetStatePrivacyMetadata is invoked on the privateState (as the tx is private) and it returns:
+		// 1. public contacts: privacyMetadata = nil, err = nil
+		// 2. private contracts of type:
+		// 2.1. StandardPrivate:     privacyMetadata = nil, err = "The provided contract does not have privacy metadata"
+		// 2.2. PartyProtection/PSV: privacyMetadata = <data>, err = nil
+		actualPrivacyMetadata, err := pmh.stAPI.GetStatePrivacyMetadata(addr)
+		//when privacyMetadata should have been recovered but wasnt (includes non-party)
+		//non party will only be caught here if sender provides privacyFlag
+		if err != nil && privacyFlag.IsNotStandardPrivate() {
+			return returnErrorFunc(nil, "PrivacyMetadata unable to be found", "err", err)
```
vsmk98: nitpick - do we want to change the message to `unable to find privacy metadata`?

core/state_transition_pmh.go
```diff
@@ -0,0 +1,145 @@
...
+		exitEarly = true
+		if anError != nil {
+			err = fmt.Errorf("vmerr=%s, err=%s", vmerr, anError)
+		}
+		return
+	}
+	actualACAddresses := pmh.stAPI.AffectedContracts()
+	log.Trace("Verify hashes of affected contracts", "expectedHashes", pmh.receivedPrivacyMetadata.ACHashes, "numberOfAffectedAddresses", len(actualACAddresses))
+	privacyFlag := pmh.receivedPrivacyMetadata.PrivacyFlag
+	actualACHashesLength := 0
```
vsmk98: `actualACHashesLength` can be removed as we are not using it after removing the validation on this.

mkdocs.yml
nicolae-leonte-go: removed

extension/privacyExtension/state_set_utilities.go
vsmk98: Agree Jitu . Will change to `Error`

nicolae-leonte-go: fixed

cmd/evm/runner.go
nicolae-leonte-go: removed

core/state_transition_pmh.go
nicolae-leonte-go: error message updated

core/state_transition_pmh.go
nicolae-leonte-go: removed

jbhurat:
~~Seeing Inconsistent error messages; the steps to reproduce are~~

- ~~Create a contract with `StandardPrivate` `privacyFlag`~~
- ~~Execute a transaction against that contract and same participants using `privacyFlag` `PartyProtection` and there is an error returned `Error: PrivacyMetadata not found: The provided contract does not have privacy metadata: fc1684a8af3e6bf532482c356e5b2cef9cad144f`~~
- ~~Execute a transaction against the same contract with no `privateFor` arguments and using `privacyFlag` `PartyProtection` or `StateValidation`, no error is returned here and the transaction is executed with status `0x1`. Expected an error to be returned and transaction to fail. The state is not updated though~~

`privateFor` flag is used to determine if a transaction is private or not, if it is skipped the transaction is treated as a public transaction which is the case above

core/rawdb/database_quorum.go
```diff
@@ -27,9 +27,11 @@ import (
...
 
 var (
-	privateRootPrefix           = []byte("P")
-	privateBloomPrefix          = []byte("Pb")
-	quorumEIP155ActivatedPrefix = []byte("quorum155active")
+	privateRootPrefix                      = []byte("P")
+	privateBloomPrefix                     = []byte("Pb")
+	quorumEIP155ActivatedPrefix            = []byte("quorum155active")
+	privateRootToPrivacyMetadataRootPrefix = []byte("PSR2PMDR")
+	emptyRoot                              = common.HexToHash("56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421")
```
trung: it may be worth to comment on this value:
```
// emptyRoot is the known root hash of an empty trie. Duplicate from `trie/trie.go#emptyRoot`
```

core/state/database.go
```diff
@@ -109,16 +113,29 @@ func NewDatabase(db ethdb.Database) Database {
...
 	}
 }
 
 type cachingDB struct {
-	db            *trie.Database
+	db *trie.Database
+	// Quorum - Privacy Enhancements
+	privacyMetadataLinker rawdb.PrivacyMetadataLinker
+	// End Quorum - Privacy Enhancements
 	codeSizeCache *lru.Cache
```
trung: There are number of places having the practice of wrapping with `Quorum - xxx` and `End Quorum - xxx` which will result misleading documentation. In particular, `codeSizeCache` will have documentation as `End Quorum - Privacy Enhancements`.

It would be clearer to document the field and its intention in a single line. E.g.:
```go
type cachingDB struct {
	db *trie.Database
	// Quorum:  Privacy enhacements introducing privacyMetadataLinker which maintains mapping between private state root and privacy metadata root. As this struct is the backing store for state, this gives the reference to the linker when needed  
	privacyMetadataLinker rawdb.PrivacyMetadataLinker
```

It can be applied for new methods or  updating existing methods. Giving some context and motivation for the changes.

core/genesis.go
```diff
@@ -248,9 +250,20 @@ func SetupGenesisBlockWithOverride(db ethdb.Database, genesis *Genesis, override
 		return newcfg, stored, compatErr
 	}
 	rawdb.WriteChainConfig(db, stored, newcfg)
+	if storedcfg.PrivacyEnhancementsBlock == nil && newcfg.PrivacyEnhancementsBlock != nil {
```
trung: Can we remove `newcfg.PrivacyEnhancementsBlock != nil` condition?  
It seems already be included in the function `checkAndPrintPrivacyEnhancementsWarning()`

nicolae-leonte-go: removed

core/rawdb/database_quorum.go
nicolae-leonte-go: added the suggested comment

core/state/database.go
nicolae-leonte-go: updated the comment you mentioned and a number of other similar comments.

trung (dismissed):
LGTM. Just make sure to rollback the `pr.yml` to make it clean as noted in my first comment.

vsmk98 (dismissed):
LGTM

trung (approved):
LGTM

vsmk98 (approved):

PR #1086, opened 2020-10-22
ci/cd: Add YY.MM docker image tag on Docker Hub

nmvalera:
- Add YY.MM tag to docker images on Docker Hub
- Update the URL of Quorum repository from which PRs are scrapped to generate release-note

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,13 +19,14 @@ jobs:
         run: |
           REF=${{ github.ref }}
           echo ::set-output name=image_tag::$(echo $REF | sed 's/refs\/tags\/v//g')
+          echo ::set-output name=image_tag_minor_latest::$(echo $REF | sed -e 's/refs\/tags\/v//g' -e 's/\.[[:digit:]]$/-latest/g') 
       - name: 'Build and publish to Docker Hub'
         uses: docker/build-push-action@v1
         with:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_ACCESS_TOKEN }}
           repository: ${{ secrets.DOCKER_REPO }}
-          tags: ${{ steps.extract.outputs.image_tag }}
+          tags: ${{ steps.extract.outputs.image_tag }}, ${{ steps.extract.outputs.image_tag_minor_latest }}
           add_git_labels: true
   build:
     name: 'Build binary for ${{ matrix.os }}'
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -19,13 +19,14 @@ jobs:
         run: |
           REF=${{ github.ref }}
           echo ::set-output name=image_tag::$(echo $REF | sed 's/refs\/tags\/v//g')
+          echo ::set-output name=image_tag_minor_latest::$(echo $REF | sed -e 's/refs\/tags\/v//g' -e 's/\.[[:digit:]]$/-latest/g')
```
trung: @nmvalera please let me know if my understanding is correct here:

| | Git Tag | Docker Tags |
| --- | ---- | ----| 
| 1 | `v20.10` | `20.10`|
| 2 | `v20.10.1` | `20.10.1` and `20.10-latest` |

In case we have more than 10 micro/patch releases in October, the Git Tag would likely be `v20.10.11` for example, then the Docker Image tag for the latest should be `20.10-latest` but from the above regex, it would be `20.10.11-latest`.

Updated: would you consider `v20.10.01` as well as it may benefit sorting?

vsmk98: @nmvalera my suggestion for the latest tag is as below:
`echo ::set-output name=image_tag_minor_latest::$(echo $REF | sed -e 's/refs\/tags\/v//g' -e 's/^\(.....\).*/\1/' -e 's/$/-latest/'`
This will ensure we always get `20.10-latest` as the tag for minor/patch releases.

nmvalera: Thanks a lot.

After rediscussing this the target is actually to set a `YY.MM` docker tag and not `YY.MM-latest`.

I pushed an update

- to update tag format to `YY.MM`
- fix the issue you mentioned

nmvalera: @trung 

What we want to achieve is 


Docker tag | git | Comment
-- | -- | --
YY.MM.Patch | tag YY.MM.Patch | Image is created when pushing git tag YY.MM.Patch and then it is never overridden
YY.MM | latest tag YY.MM.* | Image is created for the first time when pushing git tag YY.MM.0 then it is overridden on every new YY.MM.* git tagOnce bumping to YY’.MM’.0 image is not overridden anymore

vsmk98 (approved):
LGTM

trung (approved):
LGTM

PR #1100, opened 2020-11-23
Add configurable ipc timeout

SatpalSandhu61:
(Re-raising previously approved [PR#145](https://github.com/ConsenSysQuorum/quorum/pull/145), after resolving merge conflicts from privacy enhancements.)

Currently, the IPC timeout, when quorum communicates with tessera, is hard-coded to 5 seconds.This can cause issues in some cases, for example when a large number of private transactions are sequentially sent to Tessera. Here is one such issue raised in Slack: https://go-quorum.slack.com/archives/C825QTQ1Z/p1600188900086900

This PR allows the timeout to be modified. The change leverages the fact that the envar PRIVATE_CONFIG can already be set to a config file instead of the actual socket file, e.g.: `PRIVATE_CONFIG=qdata/dd1/tx-ipc-config.toml`

The toml file should contain contents like:
```
socket = "tm.ipc"
workdir = "qdata/c1"
responseHeaderTimeout = 5
```
Note that the following timeouts can be specified:
- `dialTimeout` is timeout when connecting to socket (seconds), default = 1 second
- `requestTimeout` is timeout for the write to the socket (seconds), default = 5 seconds
- `responseHeaderTimeout` is timeout for reading a response from the socket (seconds), default = 5 seconds


Krish1979 (dismissed):

private/private.go
```diff
@@ -58,42 +58,31 @@ func FromEnvironmentOrNil(name string) PrivateTransactionManager {
 }
 
 func MustNewPrivateTxManager(cfgPath string) PrivateTransactionManager {
```
nmvalera, line 60: Minor: Could be split into two functions 

```go
func NewPrivateTxManager(cfgPath string) (PrivateTransactionManager, err)
```
and

```go
func MustNewPrivateTxManager(cfgPath string) PrivateTransactionManager
```

that calls the first one.

private/private.go
```diff
@@ -58,42 +58,31 @@ func FromEnvironmentOrNil(name string) PrivateTransactionManager {
...
 	return ptm
 }
 
-func unixTransport(socketPath string) *httpunix.Transport {
+func unixTransport(cfg engine.Config) *httpunix.Transport {
 	t := &httpunix.Transport{
-		DialTimeout:           1 * time.Second,
-		RequestTimeout:        5 * time.Second,
-		ResponseHeaderTimeout: 5 * time.Second,
+		DialTimeout:           time.Duration(cfg.DialTimeout) * time.Second,
```
nmvalera, line 90: Looking at `httpunix` doc it seems that those `httpunix.Transport` timeout are deprecated and context should be used instead (c.f. https://godoc.org/github.com/tv42/httpunix#Transport).

Maybe adding some unit-test to make sure that this is working as we expect?

nmvalera: Also worth exploring if Timeout on `http.Client` may be better suited here (c.f. https://golang.org/pkg/net/http/#Client)

nmvalera: Good reference: https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/

private/private.go
```diff
@@ -58,42 +58,31 @@ func FromEnvironmentOrNil(name string) PrivateTransactionManager {
...
-		}
-		socketPath = filepath.Join(cfg.WorkDir, cfg.Socket)
-	}
 
 	client := &engine.Client{
 		HttpClient: &http.Client{
-			Transport: unixTransport(socketPath),
+			Transport: unixTransport(cfg),
 		},
 		BaseURL: "http+unix://c",
```
nmvalera, line 78: You could set `http.Client.Timeout` here

SatpalSandhu61: Ok, that could be useful. The user would need to modify `ClientTimeout` and the `ResponseHeaderTimeout` in conjunction though. That gives more flexibility (albeit with more potential for getting it wrong).

private/private.go
SatpalSandhu61: I'm not entirely comfortable removing these in case it causes issues in existing uses. I'd prefer to just add the `http.Client` timeout as an additional config option.

SatpalSandhu61: Having experimented with the use of the http.Client timeout, I don't think it's actually possible to use it.
Whilst the client.timeout does expire, the socket call still blocks until the responseHeader timeout expires within the Transport.

private/private.go
SatpalSandhu61: Ok, done.

private/private.go
```diff
@@ -63,17 +63,21 @@ func MustNewPrivateTxManager(cfgPath string) PrivateTransactionManager {
...
 	}
 
+	ptm, err := newPrivateTxManager(cfg)
+	if err != nil {
+		panic(fmt.Sprintf("unable to connect to private tx manager using %s due to %s", cfgPath, err))
+	}
+	return ptm
+}
+
+func newPrivateTxManager(cfg engine.Config) (PrivateTransactionManager, error) {
```
nmvalera: Why not exporting the function?

Also I think `NewPrivateTxManager(cfgPath string)` could take the config path as an argument possibly returning an extended error using `fmt.Errorf(...)` so in the end, `MustNewPrivateTxManager(cfgPath string)` only handles the panic behavior:

```go
func MustNewPrivateTxManager(cfgPath string) PrivateTransactionManager {
    ptm, err := NewPrivateTxManager(cfgPath string)
    if err != nil {
        panic(err)
    }
    return ptm
}
```

SatpalSandhu61: The function is only used locally, hence there's no reason to export it.
I feel it's cleaner as implemented, splitting the load config from the txn manager creation, so I prefer to leave it as is.

nmvalera (approved):

PR #1105, opened 2020-12-12
Serve multiple tenants from a single node

### LINKED ISSUE #1115: use dual state in ExtCodeHash opcode
Opcode `EXTCODEHASH` is being introduced in [EIP-1052](https://eips.ethereum.org/EIPS/eip-1052) and implemented in `geth` PR https://github.com/ethereum/go-ethereum/pull/17202

GoQuorum must be updated to reflect the dual state implementation for this opcode.

https://github.com/ConsenSys/quorum/blob/7b726385216c44e55fc6b6191a022fe598c102d4/core/vm/instructions.go#L551-L560

A potential impact is when a private contract tries to use `EXTCODEHASH` of a public contract. However, the lookup is done in the private state instead of public state.

trung:
* Support multiple tenants in a single node while preserving data isolation
   - Introduce `--multitenancy` flag to enable the feature
* Use RPC Security Plugin to support tenancy protection 
   - Design access token scope to provide granular access controls
* Introduce `AccountExtraData` to provide flexibility in managing additional data for account state. `AccountExtraData` is composed of `PrivacyMetadata` and managed parties

Also fixes #1115


core/state/state_object.go
```diff
@@ -468,22 +597,59 @@ func (s *stateObject) setNonce(nonce uint64) {
...
+}
+
+// A new AccountExtraData will be created if not exists.
+// This must be called after successfully acquiring accountExtraDataMutex lock
+func (s *stateObject) journalAccountExtraData() *AccountExtraData {
+	current, _ := s.AccountExtraData()
+	s.db.journal.append(accountExtraDataChange{
 		account: &s.address,
-		prev:    prevPM,
+		prev:    current.copy(),
```
nicolae-leonte-go, line 481: Wouldn't `copy()` fail if current was nil? Maybe do the `nil` test below before the `copy()`.

core/state_transition.go
```diff
@@ -315,12 +340,39 @@ func (st *StateTransition) TransitionDb() (ret []byte, usedGas uint64, failed bo
 	st.refundGas()
 	st.state.AddBalance(st.evm.Coinbase, new(big.Int).Mul(new(big.Int).SetUint64(st.gasUsed()), st.gasPrice))
 
+	if contractCreation && st.evm.SupportsMultitenancy {
```
nicolae-leonte-go: I think we need to remove contractCreation from the test (it may be that an existing contract creates another contract - in which case the transaction has the `to` field populated).

nicolae-leonte-go (dismissed):

core/state/state_object.go
trung: not needed as `copy()` will return `nil` if the target is `nil`

https://github.com/trung/quorum/blob/64f8fa18967d3b50dbf8e52a9a1d6674d7ed8bf3/core/state/state_object.go#L232-L235

I'm adding a test case to reflect the same

core/state_transition.go
trung: Yes, we need to populate the index for all newly created contracts regardless coming from a contract creation tx or message call tx.

Adding

core/blockchain.go
```diff
@@ -2366,3 +2379,7 @@ func (bc *BlockChain) SubscribeLogsEvent(ch chan<- []*types.Log) event.Subscript
 func (bc *BlockChain) SubscribeBlockProcessingEvent(ch chan<- bool) event.Subscription {
 	return bc.scope.Track(bc.blockProcFeed.Subscribe(ch))
 }
+
+func (bc *BlockChain) SupportsMultitenancy(context.Context) (*proto.PreAuthenticatedAuthenticationToken, bool) {
```
nmvalera, line 2383: Why do we need returning `*proto.PreAuthenticatedAuthenticationToken` if always returning `nil` ?

core/state_transition.go
```diff
@@ -304,6 +307,28 @@ func (st *StateTransition) TransitionDb() (ret []byte, usedGas uint64, failed bo
 	}
 	// End Quorum - Privacy Enhancements
 
+	// do the affected contract managed party checks
+	if msg, ok := msg.(PrivateMessage); ok && isQuorum && st.evm.SupportsMultitenancy && msg.IsPrivate() {
```
nmvalera, line 318: MINOR: Surround code block with `// Quorum` tag

core/vm/evm.go
```diff
@@ -165,18 +179,32 @@ type EVM struct {
...
 
 	// these are for privacy enhancements
-	affectedContracts map[common.Address]AffectedType // affected contract account address -> type
-	currentTx         *types.Transaction              // transaction currently being applied on this EVM
+	affectedContracts map[common.Address]*AffectedType // affected contract account address -> type
+	currentTx         *types.Transaction               // transaction currently being applied on this EVM
+	addressStack      []common.Address
+}
+
+type AffectedType struct {
```
nmvalera, line 205: MINOR: a short docstring explaining what Affected type mode and reason are and how they are set/updated would be useful

extension/api.go
```diff
@@ -102,9 +105,49 @@ func (api *PrivateExtensionAPI) checkIfPrivateStateExists(toExtend common.Addres
 	return false
 }
 
+func (api *PrivateExtensionAPI) doMultiTenantChecks(ctx context.Context, address common.Address, txa ethapi.SendTxArgs) error {
+	// do multitenancy checks
+	chainAccessor := api.privacyService.stateFetcher.chainAccessor
```
nmvalera: MINOR: it looks like chainAccessor could be declared in the if block avoiding unnecessary declaration if multitenancy is not activated

extension/api.go
```diff
@@ -102,9 +105,49 @@ func (api *PrivateExtensionAPI) checkIfPrivateStateExists(toExtend common.Addres
...
+		managedParties, err := extraDataReader.ReadManagedParties(address)
+		if err != nil {
+			return err
+		}
+		attributes = append(attributes,
+			multitenancy.NewContractSecurityAttributeBuilder().FromEOA(txa.From).Private().Write().Parties(managedParties).Build(),
+			multitenancy.NewContractSecurityAttributeBuilder().FromEOA(txa.From).Private().Read().Parties(managedParties).Build())
+
+		if authorized, _ := apiHelper.IsAuthorized(ctx, authToken, attributes); !authorized {
+			return fmt.Errorf("not authorized")
```
nmvalera: Would there be a way to make the error message more explicit about the reason for being unauthorized (guiding users about what is missing in their JWT scope)?

eth/filters/api.go
```diff
@@ -431,6 +446,14 @@ func (api *PublicFilterAPI) GetFilterChanges(id rpc.ID) (interface{}, error) {
 		case LogsSubscription:
 			logs := f.logs
 			f.logs = nil
+			ok, err := api.isAuthorized(ctx, logs)
```
nmvalera: If I understand `isAuthorized(...)` correctly if 1 of the logs is unauthorized then JSON-RPC returns an unauthorized error.

Instead of returning an error would it make sense to filter out unauthorized logs and return only authorized ones (possibly returning an empty array if no authorized logs are found)? 

I think it would make the behavior closer to a non multi-tenant world, also avoid possibly leaking info about the activity of private contracts managed by other tenants.

internal/ethapi/api.go
```diff
@@ -854,6 +855,9 @@ type account struct {
 	StateDiff *map[common.Hash]common.Hash `json:"stateDiff"`
 }
 
+// Quorum - Multitenancy
+// Before returning the result, we need to inspect the EVM and
+// perform verification check
 func DoCall(ctx context.Context, b Backend, args CallArgs, blockNrOrHash rpc.BlockNumberOrHash, overrides map[common.Address]account, vmCfg vm.Config, timeout time.Duration, globalGasCap *big.Int) ([]byte, uint64, bool, error) {
```
nmvalera, line 862: Question after reviewing `DoCall`

Is the following correct?

On `eth_call` a user does not get the same JSON-RPC response if calling 
- an address with no contract => user gets a response with empty result
- an address of a contract owned by another tenant => user gets a response with unauthorized error

internal/ethapi/api.go
```diff
@@ -1471,9 +1507,45 @@ func (s *PublicTransactionPoolAPI) GetTransactionReceipt(ctx context.Context, ha
...
+			managedParties, err := extraDataReader.ReadManagedParties(ca)
+			if errors.Is(err, common.ErrNotPrivateContract) {
+				attrBuilder.Public()
+			} else if err != nil {
+				return nil, fmt.Errorf("contract %s not found in the index due to %s", ca.Hex(), err.Error())
+			}
+			attributes = append(attributes, attrBuilder.Parties(managedParties).Build())
+		}
+		if ok, _ := s.b.IsAuthorized(ctx, authToken, attributes); !ok {
+			return nil, fmt.Errorf("not authorized")
```
nmvalera: Have considered options to return
- the private receipt in case the user has a scope with PTM key which was a participant of the private tx
- the public receipt (with empty contract address and logs) otherwise

So we get a behavior closer to non multi-tenant world.

internal/ethapi/api.go
```diff
@@ -1648,6 +1755,190 @@ func SubmitTransaction(ctx context.Context, b Backend, tx *types.Transaction) (c
...
+		privateTx = types.NewContractCreation(tx.Nonce(), tx.Value(), tx.Gas(), tx.GasPrice(), privatePayload)
+	} else {
+		privateTx = types.NewTransaction(tx.Nonce(), *tx.To(), tx.Value(), tx.Gas(), tx.GasPrice(), privatePayload)
+	}
+	return privateTx, privateFrom, nil
+}
+
+// runSimulation runs a simulation of the given transaction.
+// It returns the EVM instance upon completion
+func runSimulation(ctx context.Context, b Backend, from common.Address, tx *types.Transaction) (*vm.EVM, error) {
```
nmvalera, line 1849: Commenting here, but also valid for `DoCall`.

Looks like the approach for processing a transaction with multi-tenancy is 
1. Execute the transaction without dealing with authorizations and recording all affected contracts
2. Check user authorization to affect those contracts were, if not reject the transaction (possibly reverting state)

Would there be a way to instead bring authorization checks closer into the EVM so prior to every contract call the EVM would check the authorization to access the called contract if not then behaves as if the called contract was empty? (would passing the authorization token in the EVM context something possible?)

In the end, it would probably avoid a requirement for
- a double step procedure (exec then check)
- simulating the transaction (which is time consuming)

Again the idea here is to get a behavior closer to a non-multitenant world.

multitenancy/decision_manager.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+type DefaultContractAccessDecisionManager struct {
+}
+
+// IsAuthorized performs authorization check for each security attribute against
+// the granted access inside the pre-authenticated access token.
+//
+// All security attributes must pass.
+func (cm *DefaultContractAccessDecisionManager) IsAuthorized(ctx context.Context, authToken *proto.PreAuthenticatedAuthenticationToken, attributes []*ContractSecurityAttribute) (bool, error) {
```
nmvalera: Few comments about `IsAuthorized()` (I am super glad to discuss this further)

1. I think it would make sense to make the unauthorized error message that is returned to users more explicit, helping them to possibly re-generate a new JWT token (of course being careful not leaking information about other tenants in the unauthorized message)
2. MINOR: Could be broken into 2 functions: a granular non exported one doing the check for a single attribute (easy to test) and `IsAuthorized(...)` doing the check on multiple attributes by looping over the attributes and calling the granular function
3. MINOR: Instead of passing a slice of attributes `[]*ContractSecurityAttribute`, passing a variadic argument `...*ContractSecurityAttribute` thus when doing a check for a few attributes only it is not a requirement to declare a slice
4. MINOR: `IsAuthorized` could be declared has a function used by `*DefaultContractAccessDecisionManager.IsAuthorized` method so it is possible to call it without having to declare a `*DefaultContractAccessDecisionManager` (which is required for interface compatibility):

```go
func IsAuthorized(...) (bool, error) {
    ...
}

func (cm *DefaultContractAccessDecisionManager) IsAuthorized(...) (bool, error) {
    return IsAuthorized(...)
}
```
5. MINOR: Is the `context.Context` argument necessary? Looks like `PreAuthenticatedAuthenticationToken` extraction already happens in higher layers

multitenancy/types.go
```diff
@@ -0,0 +1,180 @@
...
+	QueryOwnedEOA = "owned.eoa"
+	// QueryToEOA query parameter is to capture the EOA address which is the
+	// target account in value transfer scenarios
+	QueryToEOA = "to.eoa"
+	// QueryFromTM query parameter is to capture the Tessera Public Key
+	// which indicates the sender of a private transaction or participant of a private contract
+	QueryFromTM = "from.tm"
+
+	// AnyEOAAddress represents wild card for EOA address
+	AnyEOAAddress = "0x0"
```
nmvalera, line 34: Why not using standard "*" wildcard for address? 

My concern here is that from a end-user perspective (manipulating the scopes) it may be confusing to not always use the same wildcard character

rpc/server.go
```diff
@@ -164,7 +164,7 @@ func (s *Server) authenticateHttpRequest(r *http.Request, cfg securityContextCon
 		if authToken, err := s.authenticationManager.Authenticate(context.Background(), token); err != nil {
 			securityContext = context.WithValue(securityContext, ctxAuthenticationError, &securityError{err.Error()})
 		} else {
-			securityContext = context.WithValue(securityContext, ctxPreauthenticatedToken, authToken)
+			securityContext = context.WithValue(securityContext, CtxPreauthenticatedToken, authToken)
```
nmvalera, line 167: Question (not directly related to this PR): why using a security context different from the request context? Would it makes sense to instead enrich the request context with the authorization data?

My concern is that context is usually error-prone and hard to maintain (because very black box) so having 2 distinct contexts flowing through the code may make things even harder... 

(this being said I think context is great and we should absolutely use it just it comes with some downsides that it is good to mitigate)

multitenancy/decision_manager.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+func toHexAddress(a common.Address) string {
+	if (a == common.Address{}) {
+		return AnyEOAAddress
+	}
+	return strings.ToLower(a.Hex())
+}
+
+func match(attr *ContractSecurityAttribute, ask, granted *url.URL) bool {
```
nmvalera: Commenting here about the general format of the scopes
- I think it would be great to be able to pass multiple account address as part of a single scope (if I understand correctly currently it is only possible pass either one address or a wildcard giving access to all accounts thus forcing to create multiple similar scopes when requiring permissions on multiple addresses which is not convenient)
- I am a bit disturbed by the read/write access control which sounds redundant with permissioning while multi-tenancy is mainly meant to provide state access control. In the end, I feel like we are bringing in another variable for users to configure increasing complexity for users. In the end if we anticipate that users will always use `_` for read/write access then I am not certain we need to make this configurable at the scope level (this is a personal opinion that I am glad to discuss, I am usually more in favor of doing simple so we avoid support questions in the future)

jpmsam: The read/write access for multi-tenancy is meant for a specific node(and it's state) that is linked to an authorization server, not for the entire network. The permissioning model is for read/write to the network, not any specific node and it's also visible to the entire network.

core/blockchain.go
trung: `EthAPIBackend` implements the interface and it would extract the token from the context and return. 
The implementation is also extracted by `APIBackendHelper` to support `PrivateExtensionAPI` logic.

core/state_transition.go
trung: `isQuorum` in the condition may already clarify the intention already without the need of explicit comments. 

Also for a fairly complex logic meant for Quorum, I would rather add some documentation at the method level than inline as it would provide a better overview of what've been changed.

core/vm/evm.go
trung: Sure, it will be added

extension/api.go
trung: Good point, will move it inside

extension/api.go
trung: From security standpoint, i would avoid to tell why users are not authorized in order to hide details from a potential attacker. `geth` log would be the right place to provide enough information for a network operator to debug if necessary.

eth/filters/api.go
trung: I understand that you want to keep users experience as they interact with their own node. Omitting the logs could be a double-edged sword. 

In the context of RPC security, when users request for logs, they do expect some sort of response. A short reason why logs are missing would be helpful.

multitenancy/types.go
trung: `0x0` fits better as it reflects the hex format of an Ethereum account address. Also we don't have to deal with URL encoding ( * = %2a) when having the wild card value in the query string. This would make the scope string more readable.

rpc/server.go
trung: `securityContext` is maintained at connection level (see `rpc/json.go#jsonCodec`), request context is actually enriched with information from `securityContext` (see `rpc/handler.go#L324`). That's why in all APIs methods, we are able to obtain pre-authenticated token from the `ctx` variable passing to the API methods.

multitenancy/decision_manager.go
trung: 1. If token is invalid, has wrong signature, expired, or wrong `aud`, an explicit message will be returned.
2. and 3. Good points, I will refactor
4. I would keep `IsAuthorized` in the struct so it gives a context on what authorization is about and defer the refactoring until we need to use it independently
5. I would keep it for the sake of continuity of the request.

internal/ethapi/api.go
trung: Added runtime multitenancy check during simulation

internal/ethapi/api.go
```diff
@@ -1475,9 +1512,45 @@ func (s *PublicTransactionPoolAPI) GetTransactionReceipt(ctx context.Context, ha
...
+				return nil, err
+			}
+			if ok {
+				filteredLogs = append(filteredLogs, l)
+			}
+		}
+		fields["logs"] = filteredLogs
+		receiptClone := &types.Receipt{PostState: receipt.PostState, Status: receipt.Status, Logs: filteredLogs}
+		fields["logsBloom"] = types.CreateBloom(types.Receipts{receiptClone})
+	}
```
nmvalera, line 1534: 1. It looks like "status" field is never overridden in case of multi-tenancy. So if a user accesses a receipt of a tx that failed and that was sent to a contract the user has no access to then the user would still be able to deduce that the transaction failed from the receipt. Is my understanding correct here? If correct, I think the user should instead see a success status in the receipt (same as if the user was accessing receipt through a node that was not a participant of the transaction)

2. It looks like "contractAddress" is never overridden. So if a user tries to access a receipt of a successful contract deployment and the user has no access to the deployed contract then the user would still be able to deduce that the deployment actually succeeded from the receipt. Is my understanding correct here? If correct, I think the user should instead see an empty "contractAddress" field in this case  (same as if the user was accessing receipt through a node that was not a participant of the transaction)

3. An edge case is for failed deployment tx. I don't see any way to hide that the tx failed to non-authorized users.

core/vm/evm.go
```diff
@@ -493,15 +564,20 @@ func (evm *EVM) create(caller ContractRef, codeAndHash *codeAndHash, gas uint64,
 	// Create a new account on the state
 	snapshot := evm.StateDB.Snapshot()
 	evm.StateDB.CreateAccount(address)
-	evm.affectedContracts[address] = Creation
+	if evm.SupportsMultitenancy && evm.AuthorizeCreateFunc != nil {
```
nmvalera, line 590: Why is the multi-tenancy check happening after calling `CreateAccount`?

core/vm/evm.go
```diff
@@ -607,78 +683,172 @@ func (evm *EVM) Create2(caller ContractRef, code []byte, gas uint64, endowment *
...
+				return evm.lastError
+			}
+		}
+		t.mode = t.mode | actualMode
+	}
+	return nil
+}
+
+// enforceMultitenancyCheck returns true if EVM is enforced to do multitenancy check, false otherwise
+func (evm *EVM) enforceMultitenancyCheck() bool {
```
nmvalera, line 889: Why not calling this method something like `IsMultitenancyEnabled`? Why "enforce"?

core/vm/evm.go
```diff
@@ -164,19 +206,47 @@ type EVM struct {
...
+	_        AffectedReason = iota
+	Creation AffectedReason = iota
 	MessageCall
 )
 
+// AffectedMode defines a mode in which the state is operated as the result of contract code execution.
+type AffectedMode byte
+
+const (
+	ModeUnknown AffectedMode = iota
```
nmvalera, line 234: Would it possible to add a docstring with the human readable value for each AffectedMode?

core/vm/evm.go
```diff
@@ -607,78 +683,172 @@ func (evm *EVM) Create2(caller ContractRef, code []byte, gas uint64, endowment *
...
+	evm.addressStack = evm.addressStack[:l-1]
+}
+
+// Quorum
+//
+// captureOperationMode stores the type of operation being applied on the current
+// affected contract whose address is on top of the stack.
+// For multitenancy, it checks if the mode is allowed. Also it bubbles up the last error
+// captured. This helps to avoid "evm: execution revert" generic error
+func (evm *EVM) captureOperationMode(isWriteOperation bool) error {
```
nmvalera, line 829: This function is a bit misleading as it does a lot of things also naming does not seem to really correspond to what the functions does.

Could we instead of this function do the following

Declare next functions 
- `func (evm *EVM) AddressStackTop() common.Address` that returns the address at the top of the address stack
- `func ModeOf(isWrite bool) AffectedMode` creating a mode from a boolean
- `func (mode AffectedMode) IsAuthorized(actualMode AffectedMode)` that basically returns `mode&actualMode != actualMode`

Then at the loop level do something like
```go
if evm.enforceMultitenancyCheck() && !evm.AffectedMode(evm.AddressStackTop()).IsAuthorized(ModeOf(operation.writes) {
       	log.Trace(...)
        evm.lastError = multitenancy.ErrNotAuthorized
	return nil, evm.lastError
    }
}
```

eth/backend.go
```diff
@@ -101,6 +99,10 @@ type Ethereum struct {
 	netRPCService *ethapi.PublicNetAPI
 
 	lock sync.RWMutex // Protects the variadic fields (e.g. gas price and etherbase)
+
+	// Quorum - Multitenancy
+	// contractAccessDecisionManager is set after node starts instead in New()
+	contractAccessDecisionManager multitenancy.ContractAccessDecisionManager
```
nmvalera: `ContractAccessDecisionManager` is a bit verbose would `ContractAccessManager` not be enough?

eth/filters/api.go
```diff
@@ -574,3 +587,35 @@ func decodeTopic(s string) (common.Hash, error) {
...
+
+// Quorum
+// Perform authorization check for each logs based on the contract addresses
+func (api *PublicFilterAPI) filterUnAuthorized(ctx context.Context, logs []*types.Log) ([]*types.Log, error) {
+	if len(logs) == 0 {
+		return logs, nil
+	}
+	if authToken, ok := api.backend.SupportsMultitenancy(ctx); ok {
+		filteredLogs := make([]*types.Log, 0)
+		for _, l := range logs {
```
nmvalera, line 599: This for block is quite massive and quite hard to follow. Would there be a way to split in more singular functions? Also the more authorization logic we have under the package `multitenancy` the best?

eth/filters/api.go
```diff
@@ -574,3 +587,35 @@ func decodeTopic(s string) (common.Hash, error) {
...
+			attributes := make([]*multitenancy.ContractSecurityAttribute, 0)
+			contractAddress := l.Address
+			extraDataReader, err := api.backend.AccountExtraDataStateReaderByNumber(ctx, rpc.BlockNumber(l.BlockNumber))
+			if err != nil {
+				return nil, fmt.Errorf("no account extra data reader at block %v: %w", l.BlockNumber, err)
+			}
+			managedParties, err := extraDataReader.ReadManagedParties(contractAddress)
+			attrBuilder := multitenancy.NewContractSecurityAttributeBuilder().Read().Private()
```
nmvalera, line 604: MINOR: why not testing for error before doing `attrBuilder := multitenancy.NewContractSecurityAttributeBuilder().Read().Private()` ?

eth/filters/api.go
```diff
@@ -574,3 +587,35 @@ func decodeTopic(s string) (common.Hash, error) {
...
+			if errors.Is(err, common.ErrNotPrivateContract) {
+				attrBuilder.Public()
+			} else if err != nil {
+				return nil, fmt.Errorf("contract %s not found in the index due to %s", contractAddress.Hex(), err.Error())
+			}
+			attributes = append(attributes, attrBuilder.Parties(managedParties).Build())
```
nmvalera: MINOR: It looks like only one attribute is required could we avoid using the attributes slice?

eth/filters/api.go
```diff
@@ -574,3 +587,35 @@ func decodeTopic(s string) (common.Hash, error) {
...
+// Perform authorization check for each logs based on the contract addresses
+func (api *PublicFilterAPI) filterUnAuthorized(ctx context.Context, logs []*types.Log) ([]*types.Log, error) {
+	if len(logs) == 0 {
+		return logs, nil
+	}
+	if authToken, ok := api.backend.SupportsMultitenancy(ctx); ok {
+		filteredLogs := make([]*types.Log, 0)
+		for _, l := range logs {
+			attributes := make([]*multitenancy.ContractSecurityAttribute, 0)
+			contractAddress := l.Address
```
nmvalera: MINOR: Is declaring `contractAddress` necessary? Can we not use `l.Address`?

multitenancy/types.go
```diff
@@ -0,0 +1,188 @@
...
+	Action      ContractAction     // create/read/write
+	PrivateFrom string             // TM Key, only if Visibility is private, for write/create
+	Parties     []string           // TM Keys, only if Visibility is private, for read
+}
+
+func (csa *ContractSecurityAttribute) String() string {
+	return fmt.Sprintf("%v visibility=%s action=%s privateFrom=%s parties=%v", csa.AccountStateSecurityAttribute, csa.Visibility, csa.Action, csa.PrivateFrom, csa.Parties)
+}
+
+type ContractSecurityAttributeBuilder struct {
```
nmvalera, line 73: I am not sure to understand why we need the builder layer? 

Could we not do the same in-place mutable approach directly `ContractSecurityAttribute` struct?

e.g. 
```go
func (attr *ContractSecurityAttribute) FromEOA(eoa common.Address) *ContractSecurityAttribute {
	attr.AccountStateSecurityAttribute.From = eoa
	return attr
}
```

extension/api.go
```diff
@@ -102,9 +105,47 @@ func (api *PrivateExtensionAPI) checkIfPrivateStateExists(toExtend common.Addres
...
+		if len(txa.PrivateFrom) == 0 {
+			return errors.New("You must specify 'privateFrom' when running in a multitenant node")
+		}
+		// check whether the user has access to txa.PrivateFrom and the txa.From eth account
+		attributes := make([]*multitenancy.ContractSecurityAttribute, 0)
+		attributes = append(attributes,
+			multitenancy.NewContractSecurityAttributeBuilder().FromEOA(txa.From).Private().Create().PrivateFrom(txa.PrivateFrom).Build(),
+			multitenancy.NewContractSecurityAttributeBuilder().FromEOA(txa.From).Private().Write().Party(txa.PrivateFrom).Build(),
+			multitenancy.NewContractSecurityAttributeBuilder().FromEOA(txa.From).Private().Read().Party(txa.PrivateFrom).Build())
+		chainAccessor := api.privacyService.stateFetcher.chainAccessor
```
nmvalera, line 116: MINOR: The next 3 line up to `extraDataReader` declaration are quite verbose with intermediate variable declaration. Would there be a way to make it less verbose?

extension/api.go
```diff
@@ -309,8 +359,25 @@ func (api *PrivateExtensionAPI) CancelExtension(extensionContract common.Address
...
+		currentBlock := apiHelper.CurrentBlock().Number().Int64()
+		extraDataReader, err := apiHelper.AccountExtraDataStateReaderByNumber(ctx, rpc.BlockNumber(currentBlock))
+		if err != nil {
+			return "", fmt.Errorf("no account extra data reader at block %v: %w", currentBlock, err)
+		}
+		managedParties, err := extraDataReader.ReadManagedParties(extensionContract)
+		if err != nil {
+			return "", err
+		}
+		attributes := []*multitenancy.ContractSecurityAttribute{
```
nmvalera: MINOR: is this slice necessary? It looks like we only use one attribute

extension/privacyExtension/state_set_utilities.go
```diff
@@ -46,7 +54,43 @@ func setPrivacyMetadata(privateState *state.StateDB, address common.Address, has
...
+		log.Error("setting privacy metadata failed", "err", err)
+		return
+	}
+
+	_, managedParties, _, _, err := ptm.Receive(ptmHash)
+	newManagedParties := appendSkipDuplicates(existingManagedParties, managedParties)
+	privateState.WriteManagedParties(address, newManagedParties)
+}
+
+func appendSkipDuplicates(list1 []string, list2 []string) (result []string) {
```
nmvalera: MINOR: 
- I would declare this as an exported utility function (maybe under a new file `common/slice.go`) 
- Also to be consistent with the `append` function signature `appendSkipDuplicates(list []string, ss ...string) []string`

extension/privacyExtension/state_set_utilities.go
```diff
@@ -46,7 +54,43 @@ func setPrivacyMetadata(privateState *state.StateDB, address common.Address, has
...
+	result = list1
+	for _, val := range list2 {
```
nmvalera: Is this optimal? Look like we are doing `len(list1) x len(list2)` complexity. Could we do `len(list1) + len(list2)` by using a map storing value that have already been added?

Something like 
```go
cache := make(map[string]bool)
for _, l := range list1 {
    cache[l] =true
}

for _, l :=range list2 {
    if !cache[l] {
         list1 = append(list1, l)
         cache[l] =true
     }
}
```

core/vm/interface.go
```diff
@@ -24,18 +26,29 @@ import (
 	"github.com/ethereum/go-ethereum/core/types"
 )
 
+type AccountExtraDataStateReader interface {
+	// Return nil for public contract
+	ReadPrivacyMetadata(addr common.Address) (*state.PrivacyMetadata, error)
```
nmvalera: MINOR: For consistency with the MinimalApiState I would rename those function `Get...` better than `Read...`

core/vm/interface.go
```diff
@@ -24,18 +26,29 @@ import (
...
+	ReadManagedParties(addr common.Address) ([]string, error)
+}
+
+type AccountExtraDataStateWriter interface {
```
nmvalera: Question: why do we need to split the interface between reader and writer? Thanks

core/vm/interface.go
```diff
@@ -24,18 +26,29 @@ import (
...
+	WritePrivacyMetadata(addr common.Address, pm *state.PrivacyMetadata)
```
nmvalera: MINOR: For consistency with the MinimalApiState I would rename those function `Set...` better than `Write...`

core/rawdb/database_quorum.go
```diff
@@ -81,28 +86,39 @@ func GetPrivateBlockBloom(db ethdb.Database, number uint64) (bloom types.Bloom)
...
-	PrivacyMetadataRootForPrivateStateRoot(privateStateRoot common.Hash) common.Hash
-	LinkPrivacyMetadataRootToPrivateStateRoot(privateStateRoot, privacyMetadataRoot common.Hash) error
+// AccountExtraDataLinker maintains mapping between root hash of the state trie
+// and root hash of state.AccountExtraData trie
+type AccountExtraDataLinker interface {
+	// Find returns the root hash of the state.AccountExtraData trie from
+	// the given root hash of the state trie.
+	//
+	// It returns an empty hash if not found.
+	Find(stateRoot common.Hash) common.Hash
```
nmvalera: MINOR: I would rename 
- `Find` --> `GetAccountExtraDataRoot`
- `Link` --> `WriteAccountExtraDataRoot`

Would be clearer to understand in the rest of the code

core/rawdb/database_quorum.go
```diff
@@ -81,28 +86,39 @@ func GetPrivateBlockBloom(db ethdb.Database, number uint64) (bloom types.Bloom)
 	return bloom
 }
 
-type PrivacyMetadataLinker interface {
-	PrivacyMetadataRootForPrivateStateRoot(privateStateRoot common.Hash) common.Hash
-	LinkPrivacyMetadataRootToPrivateStateRoot(privateStateRoot, privacyMetadataRoot common.Hash) error
+// AccountExtraDataLinker maintains mapping between root hash of the state trie
+// and root hash of state.AccountExtraData trie
+type AccountExtraDataLinker interface {
```
nmvalera, line 91: QUESTION: why to we need to create a `AccountExtraDataLinker` interface? I am asking because for accessing the private state root we did not need that. Thanks

core/state_transition.go
```diff
@@ -315,12 +347,41 @@ func (st *StateTransition) TransitionDb() (ret []byte, usedGas uint64, failed bo
...
+		}
+	}
+
 	if isPrivate {
 		return ret, 0, vmerr != nil, err
 	}
 	return ret, st.gasUsed(), vmerr != nil, err
 }
 
+func containsAll(source, target []string) bool {
```
nmvalera: MINOR: this could be declared as an exported utility function (e.g. under `common/slice.go`)

core/state_transition.go
```diff
@@ -315,12 +347,41 @@ func (st *StateTransition) TransitionDb() (ret []byte, usedGas uint64, failed bo
...
+	for _, str := range target {
```
nmvalera: MINOR: it looks like we are doing `len(source) x len(target)` complexity. Could we do `len(source) + len(target)` by using a map?

eth/filters/filter.go
```diff
@@ -26,12 +26,16 @@ import (
...
 	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/core/vm"
 	"github.com/ethereum/go-ethereum/ethdb"
 	"github.com/ethereum/go-ethereum/event"
+	"github.com/ethereum/go-ethereum/multitenancy"
 	"github.com/ethereum/go-ethereum/rpc"
 )
 
 type Backend interface {
+	multitenancy.OperationalSupport
```
nmvalera: MINOR: OperationalSupport is not very clear.

internal/ethapi/api.go
```diff
@@ -1475,9 +1512,45 @@ func (s *PublicTransactionPoolAPI) GetTransactionReceipt(ctx context.Context, ha
...
 	return fields, nil
 }
 
+func (s *PublicTransactionPoolAPI) hasAccessToContract(ctx context.Context, authToken *proto.PreAuthenticatedAuthenticationToken, extraDataReader vm.AccountExtraDataStateReader, addr common.Address) (bool, error) {
+	attributes := make([]*multitenancy.ContractSecurityAttribute, 0)
```
nmvalera: MINOR: it looks like we are using only one attribute do we need the slice?

internal/ethapi/api.go
```diff
@@ -1475,9 +1512,45 @@ func (s *PublicTransactionPoolAPI) GetTransactionReceipt(ctx context.Context, ha
...
+			}
+		}
+		fields["logs"] = filteredLogs
+		receiptClone := &types.Receipt{PostState: receipt.PostState, Status: receipt.Status, Logs: filteredLogs}
+		fields["logsBloom"] = types.CreateBloom(types.Receipts{receiptClone})
+	}
 	return fields, nil
 }
 
+func (s *PublicTransactionPoolAPI) hasAccessToContract(ctx context.Context, authToken *proto.PreAuthenticatedAuthenticationToken, extraDataReader vm.AccountExtraDataStateReader, addr common.Address) (bool, error) {
```
nmvalera: MINOR: rename `hasAccessToContract` -> `IsContractAuthorized` ?

internal/ethapi/api.go
```diff
@@ -1648,6 +1753,161 @@ func SubmitTransaction(ctx context.Context, b Backend, tx *types.Transaction) (c
...
+			return multitenancy.ErrNotAuthorized
+		}
+	}
+	currentBlock := b.CurrentBlock().Number().Int64()
+	extraDataReader, err := b.AccountExtraDataStateReaderByNumber(ctx, rpc.BlockNumber(currentBlock))
+	if err != nil {
+		return fmt.Errorf("no account extra data reader at block %v: %w", currentBlock, err)
+	}
+	// create callbacks to support runtime multitenancy checks during simulation
+	createContractSA := multitenancy.NewContractSecurityAttributeBuilder().
```
nmvalera, line 1780: MINOR: I don't think we need to declare `createContractSA` could be merge in `IsAuthorized` call

internal/ethapi/api.go
```diff
@@ -1648,6 +1753,161 @@ func SubmitTransaction(ctx context.Context, b Backend, tx *types.Transaction) (c
...
+	}
+	var authorizeMessageCallFunc multitenancy.AuthorizeMessageCallFunc = func(contractAddress common.Address) (bool, bool, error) {
+		managedParties, err := extraDataReader.ReadManagedParties(contractAddress)
+		isPrivate := true
+		if errors.Is(err, common.ErrNotPrivateContract) {
+			isPrivate = false
+		} else if err != nil {
+			return false, false, fmt.Errorf("%s not found in the index, error: %s", contractAddress.Hex(), err.Error())
+		}
+		readSecAttr := multitenancy.NewContractSecurityAttributeBuilder().FromEOA(fromEOA).PrivateIf(isPrivate).PartiesOnlyIf(isPrivate, managedParties).Read().Build()
```
nmvalera, line 1795: MINOR: I don't think we need to declare `readSecAttr` could be merge in `IsAuthorized` call

internal/ethapi/api.go
```diff
@@ -1648,6 +1753,161 @@ func SubmitTransaction(ctx context.Context, b Backend, tx *types.Transaction) (c
...
+		authorizedRead, _ := b.IsAuthorized(ctx, authToken, readSecAttr)
+		writeSecAttr := multitenancy.NewContractSecurityAttributeBuilder().FromEOA(fromEOA).PrivateIf(isPrivate).PartiesOnlyIf(isPrivate, managedParties).Write().Build()
```
nmvalera, line 1798: MINOR: I don't think we need to declare `writeSecAttr` could be merge in `IsAuthorized` call

internal/ethapi/api.go
```diff
@@ -943,8 +949,36 @@ func DoCall(ctx context.Context, b Backend, args CallArgs, blockNrOrHash rpc.Blo
...
+					isPrivate = false
+				} else if err != nil {
+					return false, false, fmt.Errorf("%s not found in the index, error: %s", contractAddress.Hex(), err.Error())
+				}
+				readSecAttr = multitenancy.NewContractSecurityAttributeBuilder().FromEOA(addr).PrivateIf(isPrivate).PartiesOnlyIf(isPrivate, managedParties).Read().Build()
+			}
+			authorizedRead, _ := b.IsAuthorized(ctx, authToken, readSecAttr)
+			return authorizedRead, false, nil
+		}
+		enrichedCtx = context.WithValue(enrichedCtx, multitenancy.CtxKeyAuthorizeMessageCallFunc, authorizeMessageCallFunc)
```
nmvalera, line 977: This looks a bit black box would there not be a cleared way to pass the `authorizeMessageCallFunc` into the EVM?

internal/ethapi/api.go
```diff
@@ -943,8 +949,36 @@ func DoCall(ctx context.Context, b Backend, args CallArgs, blockNrOrHash rpc.Blo
 	// this makes sure resources are cleaned up.
 	defer cancel()
 
+	enrichedCtx := ctx
+	// create callbacks to support runtime multitenancy checks during the run
+	if authToken, ok := b.SupportsMultitenancy(ctx); ok {
+		var authorizeMessageCallFunc multitenancy.AuthorizeMessageCallFunc = func(contractAddress common.Address) (bool, bool, error) {
```
nmvalera, line 954: Could we not add the `AuthorizeMessageCallFunc` method on the `Backend` interface? For example a method named `IsAuthorizedMessage`.

internal/ethapi/api.go
trung: 2. Without multitenancy, `contractAddress` is always visible even if user was not a participant.
3. and 1. seem having the same context, do you foresee any further security concern after knowing if a contract is accessible or not? I would argue to keep the original tx status to reflect the correct status of the transaction execution. Without multitenancy, party-protection transactions also produce fail status for transactions sent by non-party.

core/vm/evm.go
trung: Indeed, at this stage, EVM is doing simulation for multi tenancy. But `isDoingSimulationForMultitenancy` is a bit tedious and `isDoingSimulation` alone is inadequate. EVM has a context attribute `SupportsMultitenancy` which might get confused with `IsMultitenancyEnabled`.

"enforce" makes sense to me as this opcode intercept logic is an additional check and to enforce other multitenancy checks which have been done with `Authorize*` function callbacks.

core/vm/evm.go
trung: moved to after

core/vm/evm.go
trung: added comment

core/vm/evm.go
trung: i simplified the function with singular functions as suggested

eth/backend.go
trung: changed

eth/filters/api.go
trung: I simplified the for loop to make it clearer. As the data being used to construct security attributes are in that method, moving to `multitenancy` package may cause circular dependency.

eth/filters/api.go
trung: moved

eth/filters/api.go
trung: removed the slice

eth/filters/api.go
trung: it's for the log statement. But we can remove the declaration

multitenancy/types.go
trung: There are other construction logic which would be better in the builder

extension/api.go
trung: Those attributes are now moved to `multitenancy.FullAccessContractSecurityAttributes` which is also reused in `ethapi/api.go`

extension/api.go
trung: simplified

extension/privacyExtension/state_set_utilities.go
trung: moved to `common/slice.go`

extension/privacyExtension/state_set_utilities.go
trung: logic changed

core/vm/interface.go
trung: changed

core/vm/interface.go
trung: Having them separate and explicit makes it easier to reason for `AccountExtraDataStateReaderByNumber` which returns only the reader for a block height.

core/vm/interface.go
trung: changed

core/rawdb/database_quorum.go
trung: `Find` --> `GetAccountExtraDataRoot` makes sense. Changed.
I would keep `Link` function as it reflects the interface's responsibility

core/rawdb/database_quorum.go
trung: You are right, we don't access private state root with the linker. It only helps to maintain the mapping for AccountExtraData trie

core/state_transition.go
trung: done

core/state_transition.go
trung: changed

eth/filters/filter.go
trung: changed to make it clearer

internal/ethapi/api.go
trung: removed the slice

internal/ethapi/api.go
trung: changed

internal/ethapi/api.go
trung: i added a log to print out the value. It is very useful when debugging. Same for the similar comments below

internal/ethapi/api.go
trung: Keeping `GetEVM` signature makes merging with upstream less conflicts. Also the logic is encapsulated inside the `GetEVM`, `EVM` has the explicit declaration of those callbacks.

internal/ethapi/api.go
trung: Callback implementation is different for `eth_call` and simulation. Also the data required for the closure are sourced in the same function.

core/state/state_object.go
```diff
@@ -113,12 +120,134 @@ type Account struct {
...
+	})
+}
+
+// Quorum
+// AccountExtraData is to contain extra data that supplements existing Account data.
+// It is also maintained in a trie to support rollback.
+// Note:
+// - update copy() method
+// - update DecodeRLP and EncodeRLP when adding new field
+type AccountExtraData struct {
```
ricardolyn: to improve the structure, I think would be better to move `AccountExtraData` to a separate file (e.g `core/state/account_extra_data.go`). moreover, it could be applied the same approach to `PrivacyMetadata`.

core/vm/interpreter.go
```diff
@@ -217,7 +217,9 @@ func (in *EVMInterpreter) Run(contract *Contract, input []byte, readOnly bool) (
 		if in.evm.quorumReadOnly && operation.writes {
 			return nil, fmt.Errorf("VM in read-only mode. Mutating opcode prohibited")
 		}
-
+		if err := in.evm.captureOperationMode(operation.writes); err != nil {
```
ricardolyn, line 220: should we add the Quorum section comments?

internal/ethapi/api.go
```diff
@@ -1648,6 +1751,159 @@ func SubmitTransaction(ctx context.Context, b Backend, tx *types.Transaction) (c
...
+		log.Error("Simulated execution for multitenancy", "error", err)
+		return err
+	}
+	return nil
+}
+
+// Quorum
+//
+// Retrieve private payload and construct the original transaction
+func buildPrivateTransaction(tx *types.Transaction) (*types.Transaction, error) {
```
ricardolyn, line 1816: those functions seem not directly related to the `api.go` and to avoid more conflicts on geth upgrades, could be moved to a separate file.

core/state_transition.go
```diff
@@ -304,6 +314,28 @@ func (st *StateTransition) TransitionDb() (ret []byte, usedGas uint64, failed bo
 	}
 	// End Quorum - Privacy Enhancements
 
+	// do the affected contract managed party checks
+	if msg, ok := msg.(PrivateMessage); ok && isQuorum && st.evm.SupportsMultitenancy && msg.IsPrivate() {
+		if len(managedPartiesInTx) > 0 {
```
ricardolyn, line 319: non-blocking: would make sense to move blocks of code like this to be done in other external functions to make less impact on geth code?

core/vm/evm.go
```diff
@@ -62,8 +64,38 @@ func run(evm *EVM, contract *Contract, input []byte, readOnly bool) ([]byte, err
 		// During DelegateCall() CodeAddr is the address of the delegated account
 		address := *contract.CodeAddr
 		if _, ok := evm.affectedContracts[address]; !ok {
-			evm.affectedContracts[address] = MessageCall
+			evm.affectedContracts[address] = newAffectedType(MessageCall, ModeUnknown)
+			// for multitenancy, we cache the expectation of AffectedMode to be used later
+			// when an opcode is executed.
+			if evm.SupportsMultitenancy && evm.AuthorizeMessageCallFunc != nil {
```
ricardolyn: same question here

core/rawdb/database_quorum.go
```diff
@@ -81,28 +86,39 @@ func GetPrivateBlockBloom(db ethdb.Database, number uint64) (bloom types.Bloom)
...
+	// GetAccountExtraDataRoot returns the root hash of the state.AccountExtraData trie from
+	// the given root hash of the state trie.
+	//
+	// It returns an empty hash if not found.
+	GetAccountExtraDataRoot(stateRoot common.Hash) common.Hash
+	// Link saves the mapping between root hash of the state trie and
```
robinbryce, line 97: 1. Is the thrust of this change adding a state trie per contract account ?
2. Is it true that the node operator would unavoidably still see the plaintext of the transaction, but remote users of the node would never be able see transactions, receipts or logs in appropriate for their 'tenant' ?
3. Is the tenancy determined by the privateFor at contract deployment time (or something else) and can it be changed afterwords ?
4. Or, is there something else I can read that might help me understand the nature of these changes ?

internal/ethapi/api.go
nmvalera: 2. Without multitenancy, a non participant would see an empty `contractAddress` and not the effective deployment address (since a non-participant has no access to the tx payload for deployment). Or am I missing something here?

1. 3. In a non-multitenant world, a non-participant always sees a success status for private transactions. Or am I missing something here?

core/vm/evm.go
nmvalera: Ok I get your point differentiating between multitenancy on simulation and on transaction processing.

Indeed I would not name this after `Simulation` but probably something like `IsInterpreterMultitenancyEnabled`

core/vm/evm.go
nmvalera: I think it `captureOperationMode` can still be cleaned.

I am still not fan of keeping multi-tenancy checks under a function named `captureOperationMode`, when reading the code in the interpreter loop it should be clear that multi-tenancy checks are happening

eth/filters/api.go
nmvalera: Yes I don't think it is necessary. 

Usually, when seeing this kind of variable copying I expect the first variable to be modified later in the code while needing to keep track of the old value

multitenancy/decision_manager.go
```diff
@@ -0,0 +1,222 @@
...
+	ErrNotAuthorized               = errors.New("not authorized")
+	CtxKeyAuthorizeCreateFunc      = "AUTHORIZE_CREATE_FUNC"
+	CtxKeyAuthorizeMessageCallFunc = "AUTHORIZE_MESSAGE_CALL_FUNC"
+)
+
+// AccountAccessDecisionManager performs authorization checks for Ethereum Account
+// based on what is entitled in the proto.PreAuthenticatedAuthenticationToken
+// and what is asked in ContractSecurityAttribute list.
+// Note: place holder for future, this is to protect Value Transfer between accounts.
+type AccountAccessDecisionManager interface {
```
nmvalera: `AccessDecisionManager` naming is not very standard also this is not really a manager since it only does checks.

I think `AccountAuthorizationChecker` would be clearer.

multitenancy/decision_manager.go
```diff
@@ -0,0 +1,222 @@
...
+
+type AuthorizeCreateFunc func() bool
+
+// AuthorizeMessageCallFunc returns if a contract is authorized to be read / write
+type AuthorizeMessageCallFunc func(contractAddress common.Address) (authorizedRead bool, authorizedWrite bool, err error)
+
+// ContractAccessDecisionManager performs authorization checks for contract
+// based on what is entitled in the proto.PreAuthenticatedAuthenticationToken
+// and what is asked in ContractSecurityAttribute list.
+type ContractAccessDecisionManager interface {
```
nmvalera: Same goes here: renaming `ContractAccessDecisionManager` to `ContractAuthorizationChecker` would be clearer.

multitenancy/decision_manager.go
```diff
@@ -0,0 +1,222 @@
...
+// and what is asked in ContractSecurityAttribute list.
+// Note: place holder for future, this is to protect Value Transfer between accounts.
+type AccountAccessDecisionManager interface {
+	IsAuthorized(ctx context.Context, authToken *proto.PreAuthenticatedAuthenticationToken, attr *AccountStateSecurityAttribute) (bool, error)
+}
+
+type AuthorizeCreateFunc func() bool
+
+// AuthorizeMessageCallFunc returns if a contract is authorized to be read / write
+type AuthorizeMessageCallFunc func(contractAddress common.Address) (authorizedRead bool, authorizedWrite bool, err error)
```
nmvalera: AuthorizeMessageCallFunc is misleading: it returns authorizations to be checked elsewhere in the codebase which I don't think should happen. 

Authorization checks logic should be centralized in the multitenancy package that then exposes `IsAuthorized` functions to the rest of the codebase.

multitenancy/decision_manager.go
nmvalera: Actually re-reading this what is the upside of splitting into 2 interfaces?

Could we not have a single interface:
```go
type Checker interface {
     IsAccountAuthorized(...)
     IsContractAuthorized(...)
}
```
?

multitenancy/types.go
```diff
@@ -0,0 +1,200 @@
...
+	To   common.Address
+}
+
+func (assa *AccountStateSecurityAttribute) String() string {
+	return fmt.Sprintf("from=%s to=%s", assa.From.Hex(), assa.To.Hex())
+}
+
+// ContractSecurityAttribute contains security configuration ask
+// which are defined for a secure contract account
+type ContractSecurityAttribute struct {
```
nmvalera, line 61: I am not a fan of `SecurityAttribute` naming which is not standard for authentication stuff.

What about "Claim"?

multitenancy/types.go
nmvalera: Do you have any examples? Are we currently using those? Thanks

core/state/state_object.go
nmvalera: +1

multitenancy/decision_manager.go
nmvalera: Actually something quite misleading overall is also that we not clearly differentiate between multi-tenancy
- at block processing time
- at simulation time

I think this should be captured in some interfaces declared under `multitenancy` package and then imported.

core/vm/evm.go
```diff
@@ -258,7 +249,7 @@ func ModeOf(isWrite bool) AffectedMode {
 }
 
 func (mode AffectedMode) IsAuthorized(actualMode AffectedMode) bool {
-	return mode&actualMode != actualMode
+	return (mode&ModeUpdated == ModeUpdated) && (mode&actualMode != actualMode)
```
nmvalera: Why do we need to do also check for `ModeUpdated`?

core/vm/evm.go
```diff
@@ -705,6 +696,7 @@ func getDualState(evm *EVM, addr common.Address) StateDB {
 
 	if evm.PrivateState().Exist(addr) {
 		state = evm.PrivateState()
+		evm.captureAffectedContract(addr, ModeUnknown)
```
nmvalera, line 718: Good! :+1:

core/vm/evm.go
```diff
@@ -838,7 +830,54 @@ func (evm *EVM) captureOperationMode(isWriteOperation bool) error {
...
+		evm.affectedContracts[address] = newAffectedType(MessageCall, mode)
+	}
+	return evm.updateAffectedMode(address)
+}
+
+// Quorum
+//
+// updateAffectedMode is usied in simulation/eth_call for multitenancy, it sets the expectation of AffectedMode
+// to be verified later when an opcode is executed.
+func (evm *EVM) updateAffectedMode(address common.Address) error {
```
nmvalera: `updateAffectedMode` looks to do a lot could we instead do: 

1. ```(mode AffectedMode) Update(authorizedRead, authorizedWrite bool) AffectedMode``` that updates the mode
2. Update ```(evm *EVM) captureAffectedContract(...) *AffectedType``` (returning the Affected type but not doing `updateAffectedMode` so not returning any error)
3. In `getDualState` do
```
affectedType, err := evm.captureAffectedMode(...)
...
if evm.AuthorizeMessageCallFunc != nil {
    authorizedRead, authorizedWrite, err := evm.AuthorizeMessageCallFunc(address)
	...
	affectedType.mode.Update(authorizedRead, authorizedWrite)
}
```

core/vm/evm.go
trung: That is to indicate that the expectation has been set. We need this new mode as we set default to `ModeUnknown`  when capturing affected contracts.

core/vm/evm.go
```diff
@@ -838,7 +849,43 @@ func (evm *EVM) captureOperationMode(isWriteOperation bool) error {
...
+// to be verified later when an opcode is executed.
+func (evm *EVM) captureAffectedContract(address common.Address, mode AffectedMode) error {
+	if _, ok := evm.affectedContracts[address]; !ok {
+		evm.affectedContracts[address] = newAffectedType(MessageCall, mode)
+	}
+
+	if !evm.SupportsMultitenancy || evm.AuthorizeMessageCallFunc == nil {
+		return nil
+	}
+	affectedType, found := evm.affectedContracts[address]
```
nmvalera, line 860: On line 860 you are already looking for `evm.affectedContracts[address]` we probably do not need to do it a second time here.

core/vm/evm.go
```diff
@@ -838,7 +849,43 @@ func (evm *EVM) captureOperationMode(isWriteOperation bool) error {
...
+	}
+	authorizedRead, authorizedWrite, err := evm.AuthorizeMessageCallFunc(address)
+	if err != nil {
+		return err
+	}
+	authorizedMode := affectedType.mode.Update(authorizedRead, authorizedWrite)
+	log.Debug("AffectedMode changed", "address", address.Hex(), "old", affectedType.mode, "new", authorizedMode, "read", authorizedRead, "write", authorizedWrite)
+	// if we don't authorize either read/write, it's unauthorized access
+	// and we need to inform EVM
+	if !authorizedRead && !authorizedWrite {
```
nmvalera: Why not checking this right after `authorizedRead, authorizedWrite, err := evm.AuthorizeMessageCallFunc(address)` on line 871? 

Actually would it even make sense to return an error from `evm.AuthorizeMessageCallFunc()` in case both `authorizedRead` and `authorizedWrite` are `false` ?

So then we can directly do 
`affectedType.mode = affectedType.mode.Update(authorizedRead, authorizedWrite)`

nmvalera (approved):
LGTM

PR #1109, opened 2020-12-29
Enable http connection to be used for Txn manager

SatpalSandhu61:
Allow http/https connection to be used instead of the unix socket when connecting to privacy txn manager.
Leverages the toml configuration file that can currently be used when specifying the txn manager connection in the `PRIVATE_CONFIG` environment variable.
Also adds command line options which can be used instead of the `PRIVATE_CONFIG` environment variable and toml config file.

### Config File
The config file can use any of the forms shown below.

**For socket connection:**
socket = "tm.ipc"
workdir = "qdata/c1"
timeout = 5
dialTimeout = 1

**For HTTP connection:**
httpUrl = "HTTP://127.0.0.1:9101"
timeout = 5
httpWriteBufferSize = 4096
httpReadBufferSize = 4096

**For HTTPS connection:**
httpUrl = "HTTPS://127.0.0.1:9101"
tlsMode = "STRICT"
tlsRootCA = "/path/to/ca-root.cert.pem"
tlsClientCert = "/path/to/client-ca-chain.cert.pem"
tlsClientKey = "/path/to/client.key.pem"
timeout = 5
httpIdleConnTimeout = 10
httpWriteBufferSize = 4096
httpReadBufferSize = 4096

### Command Line Parameters

```
QUORUM PRIVATE TRANSACTION MANAGER OPTIONS:
  --ptm.socket value                  Path to the ipc file when using unix domain socket for the private transaction manager connection
  --ptm.url value                     URL when using http connection to private transaction manager
  --ptm.timeout value                 Timeout (seconds) for the private transaction manager connection. Zero value means timeout disabled. (default: 0)
  --ptm.dialtimeout value             Dial timeout (seconds) for the private transaction manager connection. Zero value means timeout disabled. (default: 0)
  --ptm.http.idletimeout value        Idle timeout (seconds) for the private transaction manager connection. Zero value means timeout disabled. (default: 0)
  --ptm.http.writebuffersize value    Size of the write buffer (bytes) for the private transaction manager connection. Zero value uses http.Transport default. (default: 0)
  --ptm.http.readbuffersize value     Size of the read buffer (bytes) for the private transaction manager connection. Zero value uses http.Transport default. (default: 0)
  --ptm.tls.mode value                If "off" then TLS disabled (default). If "strict" then will use TLS for http connection to private transaction manager
  --ptm.tls.rootca value              Path to file containing root CA certificate for TLS connection to private transaction manager (defaults to host's certificates)
  --ptm.tls.clientcert value          Path to file containing client certificate (or chain of certs) for TLS connection to private transaction manager
  --ptm.tls.clientkey value           Path to file containing client's private key for TLS connection to private transaction manager
  --ptm.tls.insecureskipverify        Disable verification of server's TLS certificate on connection to private transaction manager
```


private/engine/config.go
```diff
@@ -1,52 +1,121 @@
...
+const (
+	socketConnection connectionType = iota
+	httpConnection
+)
+
 type Config struct {
-	Socket  string `toml:"socket"`  // socket filename
-	WorkDir string `toml:"workdir"` // path to socket file
+	ConnectionType connectionType
+	SocketConfig   socketConfig `toml:",omitempty"`
```
nmvalera: For connection type naming I think it would be better to use "unix" instead of "socket". 

Indeed a socket can refer to other things than a unix socket thus may be confusing for users

private/engine/config.go
```diff
@@ -1,52 +1,121 @@
...
+	HttpConfig     httpConfig   `toml:",omitempty"`
+}
 
-	DialTimeout           uint // timeout for connecting to socket (seconds)
-	RequestTimeout        uint // timeout for writing to socket (seconds)
-	ResponseHeaderTimeout uint // timeout for reading from socket (seconds)
+type socketConfig struct {
```
nmvalera: Instead of using two distinct nested configuration structs `httpConfig` and `socketConfig`, could we rely on the single `Config` struct containing `ConnectionType` field and all fields for both socket and http? 

In the end, 
- it would result in a single section in the toml file which would be easier to setup from a user perspective
- It would avoid duplicated fields

private/engine/config.go
```diff
@@ -1,52 +1,121 @@
...
+	RequestTimeout        uint   // timeout for writing to socket (seconds)
+	ResponseHeaderTimeout uint   // timeout for reading from socket (seconds)
+}
 
-	// Deprecated
-	SocketPath string `toml:"socketPath"`
+type httpConfig struct {
+	Url             string // transaction manager URL for HTTP connection
+	Tls             string // if "OFF" then TLS disabled, if "STRICT" then TLS enabled
+	RootCA          string // path to file containing certificate for root CA
```
nmvalera: It is common to trust multiple CA. I think it would make sense to allow to pass RootCA as either
- a list of file paths
- a directory path (then we load all CAs in the folder)
- or both

You can refer to Golang default system root loading: https://github.com/golang/go/blob/df81a15819/src/crypto/x509/root_unix.go#L32

prd-fox (dismissed):

private/engine/config.go
```diff
@@ -1,52 +1,121 @@
...
+	DialTimeout           uint   // timeout for connecting to socket (seconds)
+	RequestTimeout        uint   // timeout for writing to socket (seconds)
+	ResponseHeaderTimeout uint   // timeout for reading from socket (seconds)
+}
 
-	// Deprecated
-	SocketPath string `toml:"socketPath"`
+type httpConfig struct {
+	Url             string // transaction manager URL for HTTP connection
+	Tls             string // if "OFF" then TLS disabled, if "STRICT" then TLS enabled
```
nmvalera: I think we need at least 3 modes for TLS configuration on client side
1. `off`: deactivated
2. `require`: forces a TLS connexion but does not verify that server certificate is CA signed (in terms of code it consists in setting `tls.Config.InsecureSkipVerify=true`)
3. `verify-full`: forces a TLS connexion, verifies that server certificate is CA signed and verifies that server host is listed in server certificate authorized domains

We could also add
4. `verify-ca`: opens a TLS connexion, verifies that server certificate is CA signed and but does not verifies that server host  is listed in server certificate authorized domains

2. Is convenient because it forces the server open a TLS connexion while not having to proceed to heavy certificate set-up
3. Is the most secured but is cumbersome to set-up
4. Is a good in between 2. & 3. but it is not straighforward to implement in Go

Currently, we have 1. and 3. would be good to add 2.

In terms of tls mode naming, I based on postgres client sslmode which are broadly used (c.f. https://www.postgresql.org/docs/9.1/libpq-ssl.html)

private/engine/config.go
```diff
@@ -1,52 +1,121 @@
...
+
+	if cfg.SocketConfig.Socket != "" {
+		cfg.ConnectionType = socketConnection
+	} else if cfg.HttpConfig.Url != "" {
+		cfg.ConnectionType = httpConnection
+		switch cfg.HttpConfig.Tls {
+		case "OFF":
+			//no action needed
+		case "STRICT":
+			if cfg.HttpConfig.RootCA == "" || cfg.HttpConfig.ClientCert == "" || cfg.HttpConfig.ClientKey == "" {
```
nmvalera: I think we are missing some options here
- RootCA can be empty in which case Go defaults to host certificates
- ClientCert & ClientKey can be empty and opening a TLS connexion is still possible. Usually, ClientCert & ClientKey are used when required by a server for mutual authentication

private/private.go
```diff
@@ -66,35 +68,119 @@ func MustNewPrivateTxManager(cfgPath string) PrivateTransactionManager {
...
+			Timeout:   time.Duration(cfg.HttpConfig.ClientTimeout) * time.Second,
+			Transport: transport,
+		},
+		BaseURL: cfg.HttpConfig.Url,
+	}
+	return client, nil
+}
+
+func httpTransportUsingTLS(cfg engine.Config) (*http.Transport, error) {
+	rootCAPool := x509.NewCertPool()
```
nmvalera: Would it not be possible to implement a single `CreateHttpTransport` function doing the job for both TLS and non-TLS? (would avoid some code duplication)

private/private.go
```diff
@@ -66,35 +68,119 @@ func MustNewPrivateTxManager(cfgPath string) PrivateTransactionManager {
...
+func httpTransport(cfg engine.Config) *http.Transport {
+	t := &http.Transport{
+		IdleConnTimeout: time.Duration(cfg.HttpConfig.IdleConnTimeout) * time.Second,
+		WriteBufferSize: cfg.HttpConfig.WriteBufferSize,
+		ReadBufferSize:  cfg.HttpConfig.ReadBufferSize,
+	}
+	return t
+}
+
+func createHTTPClientUsingTLS(cfg engine.Config) (*engine.Client, error) {
```
nmvalera: Would it not be possible to implement a single function `CreateClient` instead of 3 different ones? (would reduce code duplication)

private/engine/config.go
SatpalSandhu61: I intentionally used separate sections in the toml as a number of fields are exclusive to either the socket, or to the http config. If I don't use separate sections in the config then it means:
1. A load of validation code would need to be added, to ensure that config fields have only been defined for the relevant type of connection.
2. It would be more confusing and potentially lead to more user errors.

I actually feel the config file is easier to set up this way, and cleaner. It's certainly simpler to document.

cmd/utils/flags.go
```diff
@@ -875,6 +875,56 @@ var (
 		Usage: "Default minimum difference between two consecutive block's timestamps in seconds",
 		Value: eth.DefaultConfig.Istanbul.BlockPeriod,
 	}
+
+	// Quorum Private Transaction Manager connection options
+	QuorumPTMUnixSocketFlag = DirectoryFlag{
+		Name:  "ptm.socket",
```
nmvalera, line 884: Would `ptm.ipc-path`? or `ptm.ipc` be clearer?

private/private.go
```diff
@@ -46,55 +49,183 @@ type PrivateTransactionManager interface {
...
 	if err != nil {
-		return nil, fmt.Errorf("unable to connect to private tx manager using %s due to %s", cfgPath, err)
+		return nil, fmt.Errorf("unable to connect to private tx manager due to: %s", err)
 	}
+
+	isPrivacyEnabled = true
 	return ptm, nil
 }
 
+func createClient(cfg engine.Config) (*engine.Client, error) {
```
nmvalera: It feels like next functions are not opinionated of ptm and thus could be moved under a common package and exported (e.g under `/common/http/client.go`, `/common/http/transport.go`, `/common/http/certificate.go`):
- createClient
- unixTransport
- httpTransport
- setHttpTransportToUseTLS
- loadRootCACerts
- loadRootCAFromFile

`struct Config` could also be moved to `/common/http/config.go`

private/private.go
```diff
@@ -46,55 +49,183 @@ type PrivateTransactionManager interface {
...
+func httpTransport(cfg engine.Config) *http.Transport {
+	t := &http.Transport{
+		IdleConnTimeout: time.Duration(cfg.HttpIdleConnTimeout) * time.Second,
+		WriteBufferSize: cfg.HttpWriteBufferSize,
+		ReadBufferSize:  cfg.HttpReadBufferSize,
+	}
+	return t
+}
+
+func setHttpTransportToUseTLS(cfg engine.Config, transport *http.Transport) error {
```
nmvalera: I think this function would be more convenient with next signature `NewTLSConfig(cfg engine.Config) (*tls.Config, error)` (only generating the tls.Config from configuration). 

It can then be easily re-used by upper-level functions creating the http.Transport

cmd/utils/flags.go
SatpalSandhu61: I left the naming consistent with the toml config file (which has the name unchanged for backwards compatibility), in order to prevent confusion. However, if you think that's not important then I can change it.
Note that it's not completely consistent with the toml file as the command line param includes the path, rather than having a separate flag for the path.

private/private.go
SatpalSandhu61: Done

private/private.go
SatpalSandhu61: Done

private/private.go
SatpalSandhu61: Done

private/private.go
SatpalSandhu61: Done

private/engine/config.go
SatpalSandhu61: First point is done.
Second point is not done as Tessera does not support 1-way TLS. (It could be added if Tessera is changed to implement this in future).

private/engine/config.go
SatpalSandhu61: Modes `off` and `verify-full` are implemented but called `off` and `strict` to match the naming using in the Tessera config (there isn't any standard naming convention, but off/strict is used in other places e.g. by mysql).

Rather than add a `require` mode, I've added `insecureSkipVerify` as a separate config option, which is consistent with the security plugin and other TLS-related work we’ve done for Quorum.

I didn't add `verify-ca`, as I'm not sure it adds a lot of benefit and as per your statement, it would not be straightforward to implement.

SatpalSandhu61:
Draft PR for the documentation changes has been created: https://github.com/ConsenSys/doc.goquorum/pull/62

nmvalera (approved):
LGTM

PR #1113, opened 2021-01-19
Add new RPC APIs for IBFT and Clique

amalrajmani:
This PR adds new RPC APIs to Clique and IBFT consensus. The details of the RPC APIs as follows:

**Clique:**
_clique_status(from, to)_ - returns the signing status of blocks for a given block range. If block range is not provided it returns status of last 64 blocks by default.

**IBFT**
_istanbul_status(from, to)_ - returns the signing status of blocks for a given block range. If block range is not provided it returns status of last 64 blocks by default.
_istanbul_isValidator(blockNo)_ - returns true if this node was a validator for the given _blockNo_ . If _blockNo_ is not provided it takes the latest block number by default.


nmvalera (approved):
LGTM

PR #1124, opened 2021-02-04
Replace etcd/raft module with modified 3.3.20 from Consensys

ricardolyn:
## Summary

Apply upstream fix on raft 3.4 to the current version we use (3.3.20), to make the learner node sync up with the leader.

## Changes

* Use a modified version of v3.3.20 in the repo https://github.com/ConsenSys/etcd/tree/v3.3.20-quorum

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,7 +2,10 @@ module github.com/ethereum/go-ethereum
 
 go 1.15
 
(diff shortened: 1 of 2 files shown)


nmvalera (approved):
LGTM

PR #1126, opened 2021-02-08
Use `etcd` module from Quorum with geth 1.9.7

ricardolyn:
### Summary

Quorum used a modified version of etcd 3.3.13. By moving to go modules, we then had to create our own fork with the modified version on https://github.com/ConsenSys/etcd/tree/v3.3.13-quorum197. 

In the future, the plan is to move to the latest version of etcd (without modifications) and adapt our raft code to use the latest API.


amalrajmani (approved):
LGTM

raft/constants.go
```diff
@@ -1,11 +1,18 @@
...
+	etcdRaft "github.com/coreos/etcd/raft"
+)
+
 const (
 	//protocolName           = "raft"
 	//protocolVersion uint64 = 0x01
 
 	//raftMsg = 0x00
 
+	minterRole = etcdRaft.LEADER
```
nmvalera, line 13: I am not fan of re-declaring an etcd constant as a constant here (why not just using the constant itself?)

raft/handler.go
```diff
@@ -668,24 +669,34 @@ func (pm *ProtocolManager) isVerifier(rid uint16) bool {
 	return false
 }
 
-func (pm *ProtocolManager) handleRoleChange(raftRole etcdRaft.StateType) {
+func (pm *ProtocolManager) handleRoleChange(roleC <-chan interface{}) {
```
nmvalera, line 672: Why not using and `chan etcd.StateType`?

nmvalera (approved):
LGTM

raft/constants.go
ricardolyn: @amalrajmani only applied the code on quorum-geth-1.9.7. we need in the future to re-do this using the right StateType API on this issue: https://github.com/ConsenSysQuorum/quorum/issues/189

raft/handler.go
ricardolyn: @amalrajmani only applied the code on quorum-geth-1.9.7. we need in the future to re-do this using the right StateType API on this issue: https://github.com/ConsenSysQuorum/quorum/issues/189

PR #1128, opened 2021-02-10
[Lint] Increase timeout fix issues with PR pipelines

ricardolyn:
### Summary

Currently, there are many failures with timeouts in the linter so this should fix the problem.

### Changes

* Increase timeout to 5minutes


nmvalera (approved):
LGTM

PR #1129, opened 2021-02-10
Support eth and consensus subprotocols 

libby:
Enable the eth service to support multiple subprtocols, e.g.
"eth" AND "istanbul/100", instead of a single protocol, e.g. "eth" OR "istanbul/99".

When istanbul consensus was originally added, the eth service was
modified to override the eth subprotocol maintained by the eth service
to an quorum specific consensus protocol, e.g. "istanbul/99" meaning all
eth p2p messages would be sent over a consensus specific protocol in
addition to any consensus protocol message. In the case of istanbul,
the "eth" subprotocol would no long be included as a devp2p capability/
subprotocol for the node.

This commit changes that, and enables the eth service to return multiple
subprotocols, e.g. "eth", "istanbul/100".

"istanbul/100" has been added as a new istanbul subprotocol version, which supports
the multple subprotocols design. peers running older versions of istanbul will
continue to override the "eth" service and will communicate over a single subprotocol:

istanbul/99 <-> istanbul/99
istanbul/100, eth <-> istanbul/100, eth

To maintain backwards compatibility, when establishing a connection with a
new peer that does not support multiple protocols, but rather a "legacy"
protocol, e.g. "istanbul/64" "istanbul/99" "clique" (v2.6), the nodes will communicate
over the single "legacy" protocol and not "eth" + consensus subprotocol.

When running two subprotocols, "eth" + "consensus", the eth subprotocol
manages the lifecycle of the peer for the eth service, and the consensus
subprotocol uses this peer for its communication, registering a separate
protocol read writer on it.

maintains backwards compatibility with: istanbul, raft, clique.

### DIFF
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -211,7 +211,9 @@ func New(ctx *node.ServiceContext, config *Config) (*Ethereum, error) {
 		dbVer = fmt.Sprintf("%d", *bcVersion)
 	}
 	log.Info("Initialising Ethereum protocol", "name", protocolName, "versions", ProtocolVersions, "network", config.NetworkId, "dbversion", dbVer)
-	log.Info("Initialising Quorum consensus protocol", "name", quorumConsensusProtocolName, "versions", quorumConsensusProtocolVersions, "network", config.NetworkId, "dbversion", dbVer)
+	if chainConfig.IsQuorum {
+		log.Info("Initialising Quorum consensus protocol", "name", quorumConsensusProtocolName, "versions", quorumConsensusProtocolVersions, "network", config.NetworkId, "dbversion", dbVer)
+	}
 
 	if !config.SkipBcVersionCheck {
 		if bcVersion != nil && *bcVersion > core.BlockChainVersion {
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -186,10 +186,19 @@ func New(ctx *node.ServiceContext, config *Config) (*Ethereum, error) {
 	}
 
 	// Quorum: Set protocol Name/Version
+	// keep `var protocolName = "eth"` as is, and only update the quorum consensus specific protocol
+	// This is used to enable the eth service to return multiple devp2p subprotocols.
+	// Previously, for istanbul/64 istnbul/99 and clique (v2.6) `protocolName` would be overridden and
+	// set to the consensus subprotocol name instead of "eth", meaning the node would no longer
+	// communicate over the "eth" subprotocol, e.g. "eth" or "istanbul/99" but not eth" and "istanbul/99".
+	// With this change, support is added so that the "eth" subprotocol remains and optionally a consensus subprotocol
+	// can be added allowing the node to communicate over "eth" and an optional consensus subprotocol, e.g. "eth" and "istanbul/100"
 	if chainConfig.IsQuorum {
 		quorumProtocol := eth.engine.Protocol()
-		protocolName = quorumProtocol.Name
-		ProtocolVersions = quorumProtocol.Versions
+		// set the quorum specific consensus devp2p subprotocol, eth subprotocol remains set to protocolName as in upstream geth.
+		quorumConsensusProtocolName = quorumProtocol.Name
+		quorumConsensusProtocolVersions = quorumProtocol.Versions
+		quorumConsensusProtocolLengths = quorumProtocol.Lengths
 	}
 
 	// force to set the istanbul etherbase to node key address
@@ -202,6 +211,7 @@ func New(ctx *node.ServiceContext, config *Config) (*Ethereum, error) {
 		dbVer = fmt.Sprintf("%d", *bcVersion)
 	}
 	log.Info("Initialising Ethereum protocol", "name", protocolName, "versions", ProtocolVersions, "network", config.NetworkId, "dbversion", dbVer)
+	log.Info("Initialising Quorum consensus protocol", "name", quorumConsensusProtocolName, "versions", quorumConsensusProtocolVersions, "network", config.NetworkId, "dbversion", dbVer)
 
 	if !config.SkipBcVersionCheck {
 		if bcVersion != nil && *bcVersion > core.BlockChainVersion {
@@ -567,6 +577,17 @@ func (s *Ethereum) Downloader() *downloader.Downloader { return s.protocolManage
 func (s *Ethereum) Synced() bool                       { return atomic.LoadUint32(&s.protocolManager.acceptTxs) == 1 }
 func (s *Ethereum) ArchiveMode() bool                  { return s.config.NoPruning }
 
+// Quorum
+// adds quorum specific protocols to the Protocols() function which in the associated upstream geth version returns
+// only one subprotocol, "eth", and the supported versions of the "eth" protocol.
+// Quorum uses the eth service to run configurable consensus protocols, e.g. istanbul. Thru release v20.10.0
+// the "eth" subprotocol would be replaced with a modified subprotocol, e.g. "istanbul/99" which would contain all the "eth"
+// messages + the istanbul message and be communicated over the consensus specific subprotocol ("istanbul"), and
+// not over "eth".
+// Now the eth service supports multiple protocols, e.g. "eth" and an optional consensus
+// protocol, e.g. "istanbul/100".
+// /Quorum
+
 // Protocols implements node.Service, returning all the currently configured
 // network protocols to start.
 func (s *Ethereum) Protocols() []p2p.Protocol {
@@ -579,6 +600,15 @@ func (s *Ethereum) Protocols() []p2p.Protocol {
 	if s.lesServer != nil {
 		protos = append(protos, s.lesServer.Protocols()...)
 	}
+
+	// /Quorum
+	// add additional quorum consensus protocol if set and if not set to "eth", e.g. istanbul
+	if quorumConsensusProtocolName != "" && quorumConsensusProtocolName != protocolName {
+		quorumProtos := s.quorumConsensusProtocols()
+		protos = append(protos, quorumProtos...)
+	}
+	// /end Quorum
+
 	return protos
 }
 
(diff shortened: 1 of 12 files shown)


libby:
Following scenarios scenarios should be tested for this PR:

1. A network running all new nodes (istanbul, clique, raft). These nodes should agree on `istanbul/100` and `eth/64` subprotcols.
2. A network running new nodes, and the latest stable version `20.10.0` 

This can be done using `qctl` and minikube see https://github.com/ConsenSys/qubernetes/pull/120
build a local docker image from this branch using the [steps in qubernetes](https://github.com/ConsenSys/qubernetes/tree/master/dev)
```
> git clone https://github.com/ConsenSysQuorum/quorum.git
> cd quorum
> git checkout support-eth-and-consensus-protos
```

start minikube
```
minikube start --memory 6144
```

copy the file in `dev/docker-helpers/` from https://github.com/ConsenSys/qubernetes/pull/120 to the base `quorum` directory
Run the script to build the local container, we'll call it quorum-mul-protos
```
./quorum-build-all.sh quorum-mul-protos
```
check that the image was built in the minikube docker env
```
$> eval $(minikube docker-env)
$> docker images
quorum-mul-protos                         latest        51502f91c6f1   About an hour ago   1.29GB
...
k8s.gcr.io/pause                          3.1           da86e6ba6ca1   3 years ago         742kB
gcr.io/k8s-minikube/storage-provisioner   v1.8.1        4689081edb10   3 years ago         80.8MB
```
At this point there will be a local image in the minikube docker context that we can access when creating our local quorum networks.

#### Create a Quorum network of new nodes
```
$> qctl init --consensus=istanbul --qimagefull=quorum-mul-protos --num 3
$> qctl generate network --create
$> qctl deploy network --wait
```

#### Create a Quorum network of new nodes and latest stable release
```
$> qctl init --consensus=istanbul --qimagefull=quorum-mul-protos --num 3
```
Add 1 or more stable release nodes to the minimal config
```
$> qctl add node --qversion=20.10.0 quorum-node4
```
create and generate the network
```
$> qctl generate network --create
$> qctl deploy network --wait
```
### once the network is deployed
connect to a node
```
$> qctl connect node1
```
open the logs in vi
```
$> vi $QHOME/quorum.log
```
search for 'Starting Protocol' to see the protocols that each nodes uses to connect.

eth/handler.go
```diff
@@ -391,6 +392,15 @@ func (pm *ProtocolManager) handle(p *peer) error {
 			return err
 		}
 	}
+
+	// Quorum notify other subprotocols that the eth peer is ready, and has been added to the peerset.
+	select {
+	case p.EthPeerRegistered <- true:
```
nmvalera: I am not fan of notifying through a `chan bool` 

A more standard option is to use a `chan struct{}` and close it to broadcast notification.

In this case, since it is required to notify `true` or `false` I would instead use 2 channels 
- EthPeerReady
- EthPeerDisconnected

eth/quorum_protocol.go
```diff
@@ -0,0 +1,216 @@
...
+			}
+			protos[i] = s.protocolManager.makeQuorumConsensusProtocol(quorumConsensusProtocolName, vsn, length)
+		}
+	}
+	return protos
+}
+
+// istanbul/64, istanbul/99, clique/63, clique/64 all override the "eth" subprotocol.
+func isLegacyProtocol(name string, version uint) bool {
+	legacyProtocols := map[string][]uint{"istanbul": {64, 99}, "clique": {63, 64}}
```
nmvalera: Can we declare `legacyProtocols` as a global variable?

eth/backend.go
```diff
@@ -202,6 +211,7 @@ func New(ctx *node.ServiceContext, config *Config) (*Ethereum, error) {
 		dbVer = fmt.Sprintf("%d", *bcVersion)
 	}
 	log.Info("Initialising Ethereum protocol", "name", protocolName, "versions", ProtocolVersions, "network", config.NetworkId, "dbversion", dbVer)
+	log.Info("Initialising Quorum consensus protocol", "name", quorumConsensusProtocolName, "versions", quorumConsensusProtocolVersions, "network", config.NetworkId, "dbversion", dbVer)
```
ricardolyn: should we check `if chainConfig.IsQuorum` before the log? or show a different log

eth/backend.go
```diff
@@ -579,6 +600,15 @@ func (s *Ethereum) Protocols() []p2p.Protocol {
 	if s.lesServer != nil {
 		protos = append(protos, s.lesServer.Protocols()...)
 	}
+
+	// /Quorum
+	// add additional quorum consensus protocol if set and if not set to "eth", e.g. istanbul
+	if quorumConsensusProtocolName != "" && quorumConsensusProtocolName != protocolName {
```
ricardolyn, line 608: Q: in which situation `quorumConsensusProtocolName` could get the value `eth`?

eth/quorum_protocol.go
```diff
@@ -0,0 +1,216 @@
...
+
+// protocol Length describe the number of messages support by the protocol/version map[uint]uint64{Istanbul64: 18, Istanbul99: 18, Istanbul100: 18}
+var quorumConsensusProtocolLengths map[uint]uint64
+
+// makeQuorumConsensusProtocol is similar to eth/handler.go -> makeProtocol. Called from eth/handler.go -> Protocols.
+// returns the supported subprotocol to the p2p server.
+// The Run method starts the protocol and is called by the p2p server. The quorum consensus subprotocol,
+// leverages the peer created and managed by the "eth" subprotocol.
+// The quorum consensus protocol requires that the "eth" protocol is running as well.
+func (pm *ProtocolManager) makeQuorumConsensusProtocol(ProtoName string, version uint, length uint64) p2p.Protocol {
```
ricardolyn, line 40: do we have unit tests for this new code or other tests that are covering this?

ricardolyn (commented):
is there a plan to add Acceptance Tests for this case?

eth/backend.go
libby: In the case where the consensus engine return `eth` as the consensus protocol, e.g. `clique`.

eth/backend.go
```diff
@@ -186,10 +186,19 @@ func New(ctx *node.ServiceContext, config *Config) (*Ethereum, error) {
...
+	// set to the consensus subprotocol name instead of "eth", meaning the node would no longer
+	// communicate over the "eth" subprotocol, e.g. "eth" or "istanbul/99" but not eth" and "istanbul/99".
+	// With this change, support is added so that the "eth" subprotocol remains and optionally a consensus subprotocol
+	// can be added allowing the node to communicate over "eth" and an optional consensus subprotocol, e.g. "eth" and "istanbul/100"
 	if chainConfig.IsQuorum {
 		quorumProtocol := eth.engine.Protocol()
-		protocolName = quorumProtocol.Name
-		ProtocolVersions = quorumProtocol.Versions
+		// set the quorum specific consensus devp2p subprotocol, eth subprotocol remains set to protocolName as in upstream geth.
+		quorumConsensusProtocolName = quorumProtocol.Name
```
nmvalera, line 199: Is it necessary to use `quorumConsensusProtocolName`, `quorumConsensusProtocolVersions`, etc. 
global variable?

libby: We are mirroring the way the eth `protocolName` `ProtocolVersions` is set for consistency:
eth/backend.go
```
func (s *Ethereum) Protocols() []p2p.Protocol {
	protos := make([]p2p.Protocol, len(ProtocolVersions))
	for i, vsn := range ProtocolVersions {
```	
```
func (s *Ethereum) quorumConsensusProtocols() []p2p.Protocol {
	protos := make([]p2p.Protocol, len(quorumConsensusProtocolVersions))
	for i, vsn := range quorumConsensusProtocolVersions {
```

eth/quorum_protocol.go
libby: We can, but it is only used in the `isLegacyProtocol` method atm.

eth/backend.go
libby: sure, added a `if chainConfig.IsQuorum` check.

eth/handler.go
libby: Changed `EthPeerRegistered` to buffered `chan struct{}{}` and added `EthPeerDisconnected` instead of using chan bool and sending `true` `false`.

eth/backend.go
nmvalera: Ok I get the point.

Still I see a difference

```go
func (s *Ethereum) Protocols() []p2p.Protocol {
	protos := make([]p2p.Protocol, len(ProtocolVersions))
	for i, vsn := range ProtocolVersions {
```

uses `ProtocolVersions` global variable which is constant while

```go
func (s *Ethereum) quorumConsensusProtocols() []p2p.Protocol {
	protos := make([]p2p.Protocol, len(quorumConsensusProtocolVersions))
	for i, vsn := range quorumConsensusProtocolVersions {
```

uses `quorumConsensusProtocolVersions` global variable which is not constant. 

`quorumConsensusProtocolVersions` is indirectly set by `New(...) (*Ethereum, error)` which I think is not a good practice and maybe error-prone (e.g it creates coupling between `Ethereum` structs through a global variable which is not constant).

I think it would be better to have a `quorumConsensusProtocolVersions` attribute on the `Ethereum` struct that is set on `New(...)` then `quorumConsensusProtocols()` could re-use the attribute or something like this.

What do you think?

libby: Before this change, the `ProtocolVersions` and `ProtocolName` vars would be overridden here to set them to the  quorum specific subprototol obtained from the consensus engine.  In upstream (same geth version), they would not be overridden. This change leaves `ProtocolName` `ProtocolVersions` always set to the `eth` subprotocol. 
With this initial dual subprotocol approach, keeping the changes as minimally invasive to the existing code as possible was the approach. In the (near) future, this will be refactored more (consensus may be a service with its own protocol, when we get up to date with the current geth, eth service will be refactored to support more than one subprotocol, e.g. `snap`, `eth`, `quorumconsensus`. To the client, they won't see a difference once they are running the dual subprotocols, but the Quorum code can refactored to support a different subprotocol design.

nmvalera: Ok

nmvalera (approved):
LGTM

PR #1134, opened 2021-02-16
Count consensus messages and change to prepare state handling commit

### LINKED ISSUE #1133: Istanbul BFT could fail to achieve a consensus on a hash-locked proposal
#### System information

Geth version: `geth version`

#### Scenario triggering the issue 

There are 4 validators. They failed to achieve a consensus in round 0 because of the harsh network environment.
At the end of round 0, two of them locked the proposal receiving 4 PREPARE messages. (They will be called `hash-locked` node in the following description.) Other nodes didn't lock the proposal because they received only 2 PREPARE messages in the round for some reason.  (They will be called `non-hash-locked` nodes in the following description.)

In round 1, a hash-locked proposer sends PREPREPARE message containing the hash-locked proposal to others. Handling PREPREPARE message, 2 hash-locked nodes set their state to `Prepared` and send COMMIT message sine the proposal was already prepared in round 0. The other nodes set their state to `Preprepared` and send PREPARE message. As a result, 2 PREPARE and 2 COMMIT messages are broadcasted on the network. (Consider the network is healthy now) 

#### Expected behaviour

They should achieve a consensus because they all agreed on the proposal. 

#### Actual behaviour

They could fail to achieve a consensus depends on the message order.

If non-hash-locked nodes receive 2 PREPARE messages before 2 COMMIT messages, they fail to achieve a consensus. If non-hash-locked nodes receive a PREPARE message after 1 or 2 COMMIT messages, they will achieve a consensus. 

#### What is the problem?

The following diagram describes parts of Istanbul BFT code related to hash-lock. I think the red-colored part should exist, but it is missing in the current code. (I cannot find a document describing the consensus process for the hash-locked proposal)

In the normal IBFT consensus, a node changes its state step by step: `AcceptRequest` -> `Preprepared` -> `Prepared` -> `Committed`.  If a validator receives a valid PREPREPARE message, its state is changed from `AcceptRequest` to `Preparepared`. If a validator receives more than 2f PREPARE messages, its state is changed from `Preprepared` to `Prepared`. If a validator receives more than 2f COMMIT messages, its state is changed from `Prepared` to `Committed`.

When a hash-locked proposal is handled, however, hash-locked nodes skip `Prepared` state and sending a PREPARE message. Therefore, non-hash-locked nodes cannot receive more than 2f PREPARE messages when there are more than f hash-locked nodes exist. Fortunately, non-hash-locked nodes can change their state to `Prepared` since they count COMMIT messages as well as PREPARE messages. However, the change is only triggered by PREPARE message. If they receive only COMMIT messages after 2f messages, they cannot change to `Prepared` state even though 2f+1 Prepare or Commit messages are received. 

The problem can be solved if the red-colored logic in the following diagram is introduced.

image

To reproduce the hash-lock scenario, I used a code-patched version containing the following changes. 
- Sort validator addresses and the first half of validators drop PREPARE messages in round 0.
- Log the receiving consensus messages to distinguish the order of received consensus messages. 
Then, you can see that the result of consensus depends on the order of consensus messages.

aidan-kwon:
Istanbul BFT could fail to achieve a consensus on a hash-locked proposal depends on the consensus message order.
With this change, the problem can be resolved. The detail about the problem is described in https://github.com/ConsenSys/quorum/issues/1133 

Resolve https://github.com/ConsenSys/quorum/issues/1133

### DIFF
--- a/consensus/istanbul/core/commit.go
+++ b/consensus/istanbul/core/commit.go
@@ -68,6 +68,15 @@ func (c *core) handleCommit(msg *message, src istanbul.Validator) error {
 
 	c.acceptCommit(msg, src)
 
+	// Change to Prepared state if we've received enough PREPARE/COMMIT messages or it is locked
+	// and we are in earlier state before Prepared state.
+	if ((c.current.IsHashLocked() && commit.Digest == c.current.GetLockedHash()) || c.current.GetPrepareOrCommitSize() >= c.QuorumSize()) &&
+		c.state.Cmp(StatePrepared) < 0 {
+		c.current.LockHash()
+		c.setState(StatePrepared)
+		c.sendCommit()
+	}
+
 	// Commit the proposal once we have enough COMMIT messages and we are not in the Committed state.
 	//
 	// If we already have a proposal, we may have chance to speed up the consensus process
--- a/consensus/istanbul/core/prepare.go
+++ b/consensus/istanbul/core/prepare.go
@@ -57,7 +57,7 @@ func (c *core) handlePrepare(msg *message, src istanbul.Validator) error {
 
 	c.acceptPrepare(msg, src)
 
-	// Change to Prepared state if we've received enough PREPARE messages or it is locked
+	// Change to Prepared state if we've received enough PREPARE/COMMIT messages or it is locked
 	// and we are in earlier state before Prepared state.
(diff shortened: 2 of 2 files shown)


antonydenyer:
Closing as should be solved with QBFT

PR #1136, opened 2021-02-16
[Fix] Freezer DB quit when using RAFT mode

ricardolyn:
### Summary

After merging this https://github.com/ethereum/go-ethereum/pull/21010 as part of the geth v1.9.14 upgrade, it changed the way the Freezer DB is quit when closing. That leads to issues when running in RAFT mode where the raft service will get stuck trying to stop the `chainDb` (aka freezerDB) as the new `quit` channel added had no subscribers.

This PR fixes the issue by adding an error (similar pattern used in the freezer.Close() method) and don't get the function stuck.

### Changes

* Handled the case where the `freezer.Close()` is called twice and handled the error gracefully.
* Added unit test to replicate the issue


core/rawdb/freezer_test.go
```diff
@@ -0,0 +1,85 @@
...
+	mockFreezer := newFreezerMock(t)
+
+	// when
+	err := mockFreezer.Close()
+
+	// then
+	assert.Nil(t, err)
+}
+
+func Test_shouldCloseGracefully_whenNoSubscribers(t *testing.T) {
```
nmvalera: MINOR: I don't understand difference between `whenNoSubscribers` and `whenHaveSubscribers`. 

Honestly, I think you can gather everything under one test `TestClose`

ricardolyn: if the channel is closed, a panic will be raised.

In this case, is not an issue of closed it seems. is that there is no one receiving in the other side of the channel (subscribers).

so I'm testing both scenarios: the happy path and the path where closed is called twice.

However, it can be merged into one but then is not specific for each case (as the given is different). I can merge it together and add comments to explicitly explain the steps

ricardolyn: changed to 1 test

nmvalera (approved):
LGTM

PR #1137, opened 2021-02-17
[Fix] Use right checkout action version to v2

ricardolyn:
Update to the right checkout action version as the other configurations are using (PR & Release)


nmvalera (approved):
LGTM

PR #1130, opened 2021-02-11
[Upgrade] Go-Ethereum release v1.9.14

ricardolyn:
### Plan & Analyse

- [x] Review the Release Notes
- [x] Review PRs in the section below

As you review, list extra changes and/or tests to be implemented to ensure compatibility with GoQuorum specific features.

### Build & Test

- [x] Pull and checkout PR branch locally, then merge GoQuorum `master` into this branch
- [x] Resolve conflicts, taking into account the prior analysis
- [x] Implement required changes until all unit tests pass
- [x] Implement required changes until acceptance tests pass
- [x] Implement extra changes and/or tests

### Extra Changes & Tests

* **Changes in TransitionDb()**: As part of https://github.com/ethereum/go-ethereum/pull/20830, it was refactored the TransitionDb() to return a `ExecutionResult` struct. As this function has logic for the privacy tx handling and privacy enhancements, it was changed to now support the new data struct. It was also created new errors for those, following the pattern used in other cases.
* **Change dependencies to use math.U256Bytes()**: As changed by https://github.com/ethereum/go-ethereum/pull/21020.
* **Changed the error response data**: Changes in https://github.com/ethereum/go-ethereum/pull/20830 affected how the reverted error comes back (from release notes "we've reworked gas estimation so that reverting transactions will give you a proper error, bubbling up the EVM revert reason ([#20830](https://github.com/ethereum/go-ethereum/pull/20830))"). This will require changing tests when handling reverts from `evm: execution reverted` to `Error processing transaction request: execution reverted`. Done on https://github.com/ConsenSys/quorum-acceptance-tests/pull/76

## Go-Ethereum Release: Pink Marble (v1.9.14)

* Version: v1.9.14
* Published: 2020-05-13T09:36:41Z

### Release notes 

Geth v1.9.14 is a regular maintenance release, but is also packs a punch with some interesting new features!

Ethereum mainnet currently contains over 700M transactions. Each full node maintains a search index, stating that transactions with `hash H` is stored in `block B`. This allows you to look up an arbitrary transaction from the past (at a significant storage cost). But how often do you look up transactions from years ago?

[...]

PR #1138, opened 2021-02-17
[Fix] Wrong order when checking the privacy block in the genesis file against stored config

ricardolyn:
### Summary

There was an issue with the order when calling `checkAndPrintPrivacyEnhancementsWarning` inside the `SetupGenesisBlock`.

This created an issue in the AT `Partial Network upgrade from an old version (q2.5.0/t0.10.5) to privacy enhancements enabled version of geth and tessera ` as the test was expecting that message to come before the fatal error.

Introduced by https://github.com/ConsenSys/quorum/pull/1130.

### Changes

* Change order of `checkAndPrintPrivacyEnhancementsWarning` check

### DIFF
--- a/core/genesis.go
+++ b/core/genesis.go
@@ -238,14 +238,18 @@ func SetupGenesisBlock(db ethdb.Database, genesis *Genesis) (*params.ChainConfig
 	if height == nil {
 		return newcfg, stored, fmt.Errorf("missing block number for head header hash")
 	}
+
+	// Quorum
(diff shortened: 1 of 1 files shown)


achraf17 (approved):
looks good to go

ricardolyn:
Merge https://github.com/ConsenSys/quorum-acceptance-tests/pull/78 after this PR is merged to improve the test stability.

ricardolyn:
Not the right logic after discussion with @nicolae-leonte-go . closing

PR #1152, opened 2021-03-19
Only initialise PTM for nodes

prd-fox:
When starting up Quorum subcommands (e.g. account creation, console attaching), the PTM gets initialised and outputs a log message stating the private tens are disabled, since the user would not have set the PRIVATE_CONFIG envvar for these operations.
This misleads users into thinking the node they are actually running is not enabled to use private transactions, which is not the case.

This PR moves the initialisation to only when running an actual node instance, and stops the log message being output when not needed.


ricardolyn (approved):
LGTM

PR #1154, opened 2021-03-23
Send a message to disconnect sub protocol when eth protocol returns an error

jbhurat:
In p2p/peer.go::startProtocols() each protocol is started by calling `proto.Run()`  call. At the beginning of this method, the number of protocols running on that peer is added to `WaitGroup` delta. When any of the protocol returns an error it writes to the the `protoErr` channel which causes the peer removal. This should have been the expected behavior. But instead the call p2p/peer.go::run() never returns as it is blocked due to line `p.wg.Wait()`.  

When `proto.Run()` called from p2p/peer.go::startProtocols() returns an error for one of the many sub-protocols, it writes to the `protoErr` channel, but this is not acted upon as the p2p/peer.go::run() method gets blocked. 

A solution is to write to `EthPeerDisconnected` channel so that the `Run()` of subprotocol also returns whenever `eth` protocol returns an error

### DIFF
--- a/eth/handler.go
+++ b/eth/handler.go
@@ -350,12 +350,24 @@ func (pm *ProtocolManager) handle(p *peer, protoName string) error {
 	)
 	if err := p.Handshake(pm.networkID, td, hash, genesis.Hash(), forkid.NewID(pm.blockchain), pm.forkFilter, protoName); err != nil {
 		p.Log().Debug("Ethereum handshake failed", "protoName", protoName, "err", err)
+
+		// Quorum
+		// When the Handshake() returns an error, the Run method corresponding to `eth` protocol returns with the error, causing the peer to drop, signal subprotocol as well to exit the `Run` method
+		p.EthPeerDisconnected <- struct{}{}
+		// End Quorum
+
 		return err
 	}
 
 	// Register the peer locally
 	if err := pm.peers.Register(p, pm.removePeer, protoName); err != nil {
 		p.Log().Error("Ethereum peer registration failed", "err", err)
+
+		// Quorum
+		// When the Register() returns an error, the Run method corresponding to `eth` protocol returns with the error, causing the peer to drop, signal subprotocol as well to exit the `Run` method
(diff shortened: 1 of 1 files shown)


trung (dismissed):
LGTM

nmvalera (dismissed):
LGTM

eth/handler.go
```diff
@@ -350,12 +350,16 @@ func (pm *ProtocolManager) handle(p *peer, protoName string) error {
 	)
 	if err := p.Handshake(pm.networkID, td, hash, genesis.Hash(), forkid.NewID(pm.blockchain), pm.forkFilter, protoName); err != nil {
 		p.Log().Debug("Ethereum handshake failed", "protoName", protoName, "err", err)
+		// When the Handshake() returns an error, the Run method corresponding to `eth` protocol returns with the error, causing the peer to drop, signal subprotocol as well to exit the `Run` method
```
ricardolyn, line 355: can we add `// Quorum` and `// End Quorum` tags around?

eth/handler.go
```diff
@@ -350,12 +350,16 @@ func (pm *ProtocolManager) handle(p *peer, protoName string) error {
...
+		p.EthPeerDisconnected <- struct{}{}
 		return err
 	}
 
 	// Register the peer locally
 	if err := pm.peers.Register(p, pm.removePeer, protoName); err != nil {
 		p.Log().Error("Ethereum peer registration failed", "err", err)
+		// When the Register() returns an error, the Run method corresponding to `eth` protocol returns with the error, causing the peer to drop, signal subprotocol as well to exit the `Run` method
```
ricardolyn, line 367: can we add `// Quorum` and `// End Quorum` tags around?

ricardolyn (dismissed):
Looks good

eth/handler.go
jbhurat: Done

eth/handler.go
jbhurat: Done

ricardolyn (approved):

PR #1163, opened 2021-04-09
introduce multiple private states and multi-tenancy support

trung:
### Summary

Multiple Private States is a feature that allows a GoQuorum node to manage more than one private state. The private states are separated logically in the shared database.

This functionality lays the foundation for supporting multiple tenants on a single node. 

**The current multi-tenancy feature will be replaced by this PR**

### Performance - Update from Nam:

PASSED: Performance tests have been executed to determine the impact of having multiple private states and results demonstrated that the difference aren’t significant; moreover, the increase in the number of tenants in a single node does not cause a drop in the network’s performance.

### Note

* `import`/`export` commands are not supported in this PR


consensus/clique/snapshot_test.go
```diff
@@ -450,7 +450,7 @@ func TestClique(t *testing.T) {
 		// Pass all the headers through clique and ensure tallying succeeds
 		chain, err := core.NewBlockChain(db, nil, &config, engine, vm.Config{}, nil, nil)
 		if err != nil {
-			t.Errorf("test %d: failed to create test chain: %v", i, err)
+			t.Errorf("test %d`: failed to create test chain: %v", i, err)
```
ricardolyn: mistake?

core/blockchain.go
```diff
@@ -666,22 +693,52 @@ func (bc *BlockChain) Processor() Processor {
...
+func (bc *BlockChain) State() (*state.StateDB, mps.PrivateStateRepository, error) {
 	return bc.StateAt(bc.CurrentBlock().Root())
 }
 
-// StateAt returns a new mutable state based on a particular point in time.
-func (bc *BlockChain) StateAt(root common.Hash) (*state.StateDB, *state.StateDB, error) {
-	publicStateDb, privateStateDb, err := state.NewDual(root, bc.stateCache, bc.snaps, bc.db, bc.privateStateCache, nil)
+// StatePSI returns a new mutable public state and a mutable private state for given PSI,
+// based on the current HEAD block.
+func (bc *BlockChain) StatePSI(psi types.PrivateStateIdentifier) (*state.StateDB, *state.StateDB, error) {
```
ricardolyn, line 704: I think is always better to add Quorum functions at the end of the file as it simplifies geth upgrades

core/state/statedb.go
```diff
@@ -159,6 +159,7 @@ func New(root common.Hash, db Database, snaps *snapshot.Tree) (*StateDB, error)
 
 // Quorum
 // NewDual - Create a public and private state from a given public and private tree
+// TODO - rebase - this is not used (except in tests). It was used in the trace APIs. Should we remove it?
```
ricardolyn: if not used, then we should remove it. should not be left only for tests

miner/worker.go
```diff
@@ -785,25 +786,46 @@ func (w *worker) updateSnapshot() {
 }
 
 func (w *worker) commitTransaction(tx *types.Transaction, coinbase common.Address) ([]*types.Log, error) {
-	snap := w.current.state.Snapshot()
-	privateSnap := w.current.privateState.Snapshot()
-
+	workerEnv := w.current
```
ricardolyn, line 789: would make sense to encapsulate this block of code before the `core.ApplyTransaction(..)` into a different quorum specific function?

rpc/context.go
```diff
@@ -0,0 +1,100 @@
+package rpc
```
ricardolyn, line 2: could we add some notes at the top of the file or a different name of the file to know that this is a Quorum specific file?

ricardolyn (commented):
Minor comments. Some functions might be missing the Quorum comments

consensus/clique/snapshot_test.go
trung: fixed

core/blockchain.go
trung: added

core/state/statedb.go
trung: removed

rpc/context.go
trung: added

ricardolyn (approved):
LGTM 👍

PR #1178, opened 2021-04-26
go.mod: fix crypto/secp256k1 package replacement

trung:
- `go.mod` must replace the local package rather than a remote so the build will be consistently run against the current code base


ricardolyn (approved):
👍

PR #1185, opened 2021-04-29
Fix failure on release pipeline with missing artifact

ricardolyn:
### Changes

* Use the * to include all files.
* Fail upload if the file is not there
* Add a listing to make sure the file is there


achraf17 (approved):
LGTM

PR #1194, opened 2021-05-11
Do not log error if header data is missing in ancient or level db

### LINKED ISSUE #1193: Missing header from ancient or level db should not be logged as error
In `core/rawdb/accessors_chain.go::ReadHeader()` if the call to `readHeaderRLP()` returns `nil` header it is logged as an `error`. This should be not logged as an `error` as it is completely valid to not have a recent header in ancient or level db. 

Consider the scenario where a block was just mined on node 1 and the block was propagated to node 2, `block_fetcher.go::importBlocks()` spawns a goroutine to import/insert this block.  As part of the insert, it checks if the block header is present in ancient/level db, which obviously is not the case. An error `Invalid block header RLP`, is then logged and the `block_fetcher.go::importBlocks()` continues with the block insertion.

The log in question is https://github.com/ConsenSys/quorum/blob/c862562528104dfd3c266771d30c3e645be6f8a7/core/rawdb/accessors_chain.go#L342

jbhurat:
Fixes https://github.com/ConsenSys/quorum/issues/1193

When reading header data from ancient or level db, if encoded header is nil, then do not log an error

### DIFF
--- a/core/rawdb/accessors_chain.go
+++ b/core/rawdb/accessors_chain.go
@@ -337,7 +337,11 @@ func HasHeader(db ethdb.Reader, hash common.Hash, number uint64) bool {
 
 // ReadHeader retrieves the block header corresponding to the hash.
 func ReadHeader(db ethdb.Reader, hash common.Hash, number uint64) *types.Header {
-	_, header := readHeaderRLP(db, hash, number)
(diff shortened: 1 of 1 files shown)


ricardolyn (approved):
LGTM

PR #1195, opened 2021-05-14
[Fix] Use right block number when checking the current block number on gas estimation

ricardolyn:
## Summary

The quorum specific code that takes into account the possibility of a zero value tx could be public or private had a bug which it was not checking the right block number when the configuration had `istanbulBlock` as 0.

Original fix by @prd-fox 

## Changes

* Use current header to get current block number
* Add unit test

## References

Merge after https://github.com/ConsenSys/quorum-acceptance-tests/pull/101

### DIFF
--- a/internal/ethapi/api_test.go
+++ b/internal/ethapi/api_test.go
@@ -7,6 +7,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/ethereum/go-ethereum/common/math"
+
 	"github.com/ethereum/go-ethereum/accounts"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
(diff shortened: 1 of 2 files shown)


internal/ethapi/api_test.go
```diff
@@ -133,6 +148,24 @@ func teardown() {
 	log.Root().SetHandler(log.DiscardHandler())
 }
 
+func TestDoEstimateGas_whenNoValueTx_Homestead(t *testing.T) {
```
prd-fox: Homestead is taken into account for contract deployments (53k stipend instead of 21k). I think the test should reflect that, although probably needs some actual data instead of random bytes.
Additionally, we would need to add the homestead config like we do for the Istanbul test below.

ricardolyn: on `QuorumTestChainConfig` homestead is by default 0, that's why I didn't add. do we need?

`Homestead is taken into account for contract deployments (53k stipend instead of 21k)`: you mean private contracts?

prd-fox: Ah okay that's fine then.

Thinking about the Homestead issue again, I think it actually doesn't matter. The only thing making a difference to the intrinsic gas cost between public and private is the data field, not whether it's a contract deployment or not. So if we did mess up the value we pass in for homestead, it would happen both times and although the absolute intrinsic gas would be wrong, the *difference* between them would be the same, cancelling out.

tl;dr: i don't think it matters to test the homestead value. Maybe the test name should be `TestDoEstimateGas_whenNoValueTx_Pre_Istanbul` to be clearer though, as that's what we're testing.

ricardolyn: ok, I will change the naming.

prd-fox (approved):

nmvalera (approved):
LGTM

PR #1202, opened 2021-05-26
go.mod: remove replace crypto package

nmvalera:
Remove `github.com/ethereum/go-ethereum/crypto/secp256k1 => ./crypto/secp256k1` replace in `go.mod` to avoid issues when importing GoQ. packages

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,10 +3,7 @@ module github.com/ethereum/go-ethereum
 go 1.15
 
 // Quorum - Replace Go modules that use modifications done by us
-replace (
-	github.com/coreos/etcd => github.com/Consensys/etcd v3.3.13-quorum197+incompatible
-	github.com/ethereum/go-ethereum/crypto/secp256k1 => ./crypto/secp256k1
-)
+replace github.com/coreos/etcd => github.com/Consensys/etcd v3.3.13-quorum197+incompatible
 
 // End Quorum
 
@@ -34,7 +31,6 @@ require (
 	github.com/eapache/channels v1.1.0
 	github.com/eapache/queue v1.1.0 // indirect
 	github.com/edsrzf/mmap-go v0.0.0-20160512033002-935e0e8a636c
-	github.com/ethereum/go-ethereum/crypto/secp256k1 v0.0.0
 	github.com/fatih/color v1.7.0
 	github.com/fjl/memsize v0.0.0-20180418122429-ca190fb6ffbc
 	github.com/gballet/go-libpcsclite v0.0.0-20190607065134-2772fd86a8ff
--- a/crypto/secp256k1/go.mod
+++ b/crypto/secp256k1/go.mod
@@ -1,3 +1,0 @@
-module github.com/ConsenSys/quorum/crypto/secp256k1
-
-go 1.13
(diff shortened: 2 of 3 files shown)


ricardolyn (approved):
👍

trung:
GoQ is LGPL, `secp256k1` is BSD. Removing this will take away the licensing awareness for any non-LGPL project (like [quorum-account-plugin-hashicorp-vault](https://github.com/ConsenSys/quorum-account-plugin-hashicorp-vault/blob/master/go.mod)) depending on `secp256k1`.

go.mod
```diff
@@ -3,10 +3,7 @@ module github.com/ethereum/go-ethereum
 go 1.15
```
darioAnongba, line 3: shouldn't the module be `github.com/consensys/quorum` ? I think that's one of the main issues because otherwise it will not appear when doing a `go get`

ricardolyn: that's why I asked before if this will solve the imports.

I think to properly solve it, we need to change the path here and the path on the code imports.

chris-j-h (changes requested):
As @trung mentioned, the secp256k1 package has a BSD-3 licence and was made a module so that other projects can import it directly, without also depending on the rest of goquorum which has an LGPL licence.  

This was so that dependent projects are free to be licensed as desired (any go project that depends on a go LGPL package must also be licensed as LGPL).

So, this should not be removed without first addressing the dependency in the [quorum-account-plugin-hashicorp-vault](https://github.com/ConsenSys/quorum-account-plugin-hashicorp-vault/blob/master/go.mod).

If this sub-module is causing significant problems that warrant the change, an option could be to create a standalone project for the secp256k1 code so that other projects can continue to use it.  I'll review the BSD-3 licensing to see if that's permitted.

---

edit: the BSD-3 licence does permit copying.  The secp256k1 package is relatively stable but does see updates every so often.  If we were to go the route of creating a standalone project then any updates should be made sure to be pulled across (this could be done with an Actions workflow).

VGLoic:
After discussion, the path for this issue will be:
1. a standalone repository for `crypto/secp256k1`  containing the modifications done in GoQuorum will be created, it will be licensed under 3-Clause BSD. The name of the repository has been chosen as `goquorum-crypto-secp256k1`. The import path has been chosen as `github.com/ConsenSys/goquorum-crypto-secp256k1`.
2. this library will be used instead of the previous `github.com/jpmorganchase/quorum/crypto/secp256k1`. Dependencies need to be updated in https://github.com/ConsenSys/quorum-account-plugin-hashicorp-vault/blob/master/go.mod and https://github.com/ConsenSys/quorum-go-utils/blob/master/go.mod.
3. the GoQuorum repository will remove the isolated `go.mod` for `github.com/jpmorganchase/quorum/crypto/secp256k1`, remove the associated replace in the root `go.mod` and the dependencies on `github.com/jpmorganchase/quorum/crypto/secp256k1`.
4. a github action will be added in the new repository in order to pull the updates from the GoQuorum repository and publish a new version

ricardolyn:
moving to done as repository is ready and public: https://github.com/ConsenSys/goquorum-crypto-secp256k1

PR #1204, opened 2021-06-03
Upgradebot workflow

baptiste-b-pegasys:
- workflow for checking go-ethereum releases

### DIFF
--- a/.github/workflows/upgradebot.yml
+++ b/.github/workflows/upgradebot.yml
@@ -21,7 +21,7 @@ jobs:
         id: go
       - name: Clone upgrade geth bot
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/upgradebot.yml
```diff
@@ -0,0 +1,27 @@
+name: Upgrade geth bot
+on:
+  push:
+  schedule:
+    # * is a special character in YAML so you have to quote this string
+    - cron:  '0 6,12 * * *'
```
ricardolyn, line 6: I would shift the 6 to 4, to be earlier also for.

do we need the 12? not sure if we need more than daily.

PR #1205, opened 2021-06-07
Add a scheduled job in order to check geth upgrades everyday

baptiste-b-pegasys:
## Changes

* add a workflow in github action that is a cron job that will execute the upgrade geth bot
* if needed, a pull-request is created with the latest geth upgrade

### DIFF
--- a/.github/workflows/upgradebot.yml
+++ b/.github/workflows/upgradebot.yml
@@ -21,7 +21,7 @@ jobs:
         id: go
       - name: Clone upgrade geth bot
         run: |
-          git clone https://github.com/pegasys-dev/quorum-bots.git $GOPATH/src/upgradebot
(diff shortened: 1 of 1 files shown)


.github/workflows/upgradebot.yml
```diff
@@ -0,0 +1,26 @@
+name: Upgrade geth bot
+on:
+  schedule:
```
baptiste-b-pegasys, line 4:
```suggestion
  workflow_dispatch:
  schedule:
```

.github/workflows/upgradebot.yml
```diff
@@ -0,0 +1,27 @@
+name: Upgrade geth bot
```
ricardolyn:
```suggestion
name: Run Geth Upgrade
```

ricardolyn (approved):
LGTM 👍

PR #1209, opened 2021-06-07
QBFT Consensus Implementation

jbhurat:
This PR introduces a new QBFT Consensus based on the [The Istanbul BFT Consensus Algorithm](https://arxiv.org/abs/2002.03613) by @hmoniz. The paper addresses liveness issues found in the GoQuorum IBFT implementation. 

The PR introduces a new fork block `qbftBlock` which can be used to upgrade IBFT networks to the new QBFT network.


consensus/qibft/core/core.go
```diff
@@ -0,0 +1,304 @@
...
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/event"
+	"github.com/ethereum/go-ethereum/log"
+	metrics "github.com/ethereum/go-ethereum/metrics"
+	"gopkg.in/karalabe/cookiejar.v2/collections/prque"
+)
+
+// New creates an Istanbul consensus core
+func New(backend istanbul.Backend, config *istanbul.Config) Engine {
+	r := metrics.NewRegistry()
```
nmvalera: Why creating a new metrics registry? Can we not use the global one (similarly to other places in the code using for example `metrics.NewRegisteredMeter(...,nil)`)

consensus/qibft/types.go
```diff
@@ -0,0 +1,89 @@
...
+	"fmt"
+	"io"
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/rlp"
+)
+
+// Proposal supports retrieving height and serialized block to be used during Istanbul consensus.
+type Proposal interface {
```
nmvalera: Why do we need to re-declare `Proposal` interface that is already in `istanbul` ?

consensus/qibft/types.go
```diff
@@ -0,0 +1,89 @@
...
+	String() string
+}
+
+// View includes a round number and a sequence number.
+// Sequence is the block number we'd like to commit.
+// Each round has a number and is composed by 3 steps: preprepare, prepare and commit.
+//
+// If the given block is not accepted by validators, a round change will occur
+// and the validators start a new round with round+1.
+type View struct {
```
nmvalera: Is `qbft.View` different from `istanbul.View` ?

core/blockchain.go
```diff
@@ -1739,8 +1739,10 @@ func (bc *BlockChain) addFutureBlock(block *types.Block) error {
 //
 // After insertion is done, all accumulated events will be fired.
 func (bc *BlockChain) InsertChain(chain types.Blocks) (int, error) {
+	//log.Warn("QBFT: CALLING INSERT CHAIN CHAIN CHAIN", "len", len(chain))
```
nmvalera: Multiple commented logs in scripts, should they be removed?

core/blockchain.go
```diff
@@ -1754,10 +1756,11 @@ func (bc *BlockChain) InsertChain(chain types.Blocks) (int, error) {
 	// Do a sanity check that the provided chain is actually ordered and linked
 	for i := 1; i < len(chain); i++ {
 		block = chain[i]
+		//log.Warn("QBFT: BLOCK", "#", block.Number())
 		prev = chain[i-1]
 		if block.NumberU64() != prev.NumberU64()+1 || block.ParentHash() != prev.Hash() {
 			// Chain broke ancestry, log a message (programming error) and skip insertion
-			log.Error("Non contiguous block insert", "number", block.Number(), "hash", block.Hash(),
+			log.Error("QBFT: Non contiguous block insert", "number", block.Number(), "hash", block.Hash(),
```
nmvalera: Why adding QBFT prefix here?

core/types/istanbul.go
```diff
@@ -105,3 +122,123 @@ func IstanbulFilteredHeader(h *Header, keepSeal bool) *Header {
...
+	if err != nil {
+		return nil, err
+	}
+	return qbftExtra, nil
+}
+
+// QbftFilteredHeader returns a filtered header which some information (like committed seals, round, validator vote)
+// are clean to fulfill the Istanbul hash rules. It returns nil if the extra-data cannot be
+// decoded/encoded by rlp.
+func QbftFilteredHeader(h *Header) *Header {
```
nmvalera: In order to avoid code duplication, could this be `QbftFilteredHeaderWithRound(h, 0)` ?

consensus/istanbul/backend.go
```diff
@@ -36,14 +36,14 @@ type Backend interface {
 	EventMux() *event.TypeMux
 
 	// Broadcast sends a message to all validators (include self)
-	Broadcast(valSet ValidatorSet, payload []byte) error
+	Broadcast(valSet ValidatorSet, code uint64, payload []byte) error
```
ricardolyn, line 39: Q: what is the `code`?

consensus/istanbul/backend/backend.go
```diff
@@ -172,15 +186,32 @@ func (sb *backend) Commit(proposal istanbul.Proposal, seals [][]byte) error {
...
 	}
+
+	// Remove ValidatorSet added to ProposerPolicy registry, if not done, the registry keeps increasing size with each block height
+	sb.config.ProposerPolicy.ClearRegistry()
+
 	// update block's header
 	block = block.WithSeal(h)
 
-	sb.logger.Info("Committed", "address", sb.Address(), "hash", proposal.Hash(), "number", proposal.Number().Uint64())
+	sb.logger.Info("QBFT: Committed", "address", sb.Address(), "hash", proposal.Hash(), "number", proposal.Number().Uint64())
```
ricardolyn: might not be QBFT, right? depends on `sb.IsQIBFTConsensus()`

consensus/istanbul/backend/backend.go
```diff
@@ -316,3 +352,87 @@ func (sb *backend) HasBadProposal(hash common.Hash) bool {
...
+
+	if sb.chain != nil && sb.chain.CurrentHeader().Number.Cmp(sb.config.QibftBlock) >= 0 {
+		return true
+	}
+	return false
+}
+
+// IsQIBFTConsensusCrossed returns whether qbft consensus block has crossed
+// This is useful to see when the snapshot headers should be applied
+func (sb *backend) IsQIBFTConsensusCrossed() bool {
```
ricardolyn: this code is a duplicate of `IsQIBFTConsensus`, except of `>= 1` instead of `>= 0`. can we not re-use the same function with an argument to avoid duplication of most of the code?

consensus/istanbul/backend/backend.go
```diff
@@ -316,3 +352,87 @@ func (sb *backend) HasBadProposal(hash common.Hash) bool {
 func (sb *backend) Close() error {
 	return nil
 }
+
+// IsQIBFTConsensus returns whether qbft consensus should be used
+func (sb *backend) IsQIBFTConsensus() bool {
```
ricardolyn: by looking to this code, is the same as using ``IsQIBFTConsensusForHeader(sb.chain.CurrentHeader())` no? to avoid duplication

consensus/istanbul/backend/snapshot.go
```diff
@@ -178,84 +178,191 @@ func (s *Snapshot) apply(headers []*types.Header) (*Snapshot, error) {
...
-			snap.Tally = make(map[common.Address]Tally)
-		}
-		// Resolve the authorization key and check against validators
-		validator, err := ecrecover(header)
-		if err != nil {
-			return nil, err
-		}
-		if _, v := snap.ValSet.GetByAddress(validator); v == nil {
-			return nil, errUnauthorized
+		if isQBFTConsensus && header.Number.Uint64() > qbftBlockNumber {
```
ricardolyn: why do we need `header.Number.Uint64() > qbftBlockNumber`? `isQBFTConsensus ` should already handle that case

consensus/istanbul/config.go
```diff
@@ -16,29 +16,99 @@
...
 )
 
+// ProposerPolicy represents the Validator Proposer Policy
+type ProposerPolicy struct {
+	Id       ProposerPolicyId    // Could be RoundRobin or Sticky
+	Registry []ValidatorSet      // Holds the ValidatorSet for a given block height
+	By       validatorSortByFunc // func that defines how the ValidatorSet should be sorted
+}
+
+// NewRoundRobinProposerPolicy returns a RoundRobin ProposerPolicy with ValidatorSortByStringFunc as default sort function
```
ricardolyn: what are the rules to do one or another?

consensus/istanbul/core/core.go
```diff
@@ -197,6 +197,17 @@ func (c *core) startNewRound(round *big.Int) {
 		logger = c.logger.New("old_round", c.current.Round(), "old_seq", c.current.Sequence())
 	}
 
+	logger.Trace("Start new ibft round")
+
+	// If new round is 0, then check if qibftConsensus needs to be enabled
+	if round.Uint64() == 0 && c.backend.IsQIBFTConsensus() {
+		logger.Trace("Starting qibft consensus as qibftBlock has passed")
+		if err := c.backend.StartQIBFTConsensus(); err != nil {
+			// TODO Do we panic or continue and try to start qibft when processing the next sequence
```
ricardolyn: what can make QIBFT fail to start? will this create inconsistencies between nodes if we don't panic?

core/blockchain.go
ricardolyn: to remove debug logs? same in other places on the file

docs/Consensus/ibft/ibft-parameters.md
```diff
@@ -0,0 +1,75 @@
...
+configuration. 
+
+The options are as follows: 
+```
+{
+    "config": {
+        "istanbul": {
+            "epoch": 30000,
+            "policy": 0,
+            "ceil2Nby3Block": 0
```
ricardolyn: is not missing the example of `qibftBlock`?

eth/handler.go
```diff
@@ -424,6 +424,7 @@ func (pm *ProtocolManager) handle(p *peer, protoName string) error {
 // handleMsg is invoked whenever an inbound message is received from a remote
 // peer. The remote connection is torn down upon returning any error.
 func (pm *ProtocolManager) handleMsg(p *peer) error {
+	//log.Warn("QBFT: PM handleMsg")
```
ricardolyn: remove?

consensus/istanbul/backend.go
jbhurat: The code identifies the type of message `0x12` for `PREPREPARE`, `0x13` for `PREPARE`, `0x14` for `COMMIT` and `0x15` for `ROUNDCHANGE` qbft message. This would be `0x11` for legacy IBFT

consensus/istanbul/backend/backend.go
jbhurat: `IsQIBFTConsensusForHeader` checks if the qbft consensus was enabled for a given height (header), where as `IsQIBFTConsensus()` checks if the qbft consensus is enabled for the current height/block

consensus/istanbul/backend/snapshot.go
jbhurat: `Snapshot.go::apply()` processes headers from previous block. The check here is to ensure if qbft consensus was enabled for that particular block height identified by the header

consensus/istanbul/config.go
jbhurat: There are two types of validator sorting, sort by string and sort by byte. Legacy IBFT uses sort by string, where as QBFT uses sort by byte. When QBFT consensus is started, the sorting is set to sort by byte

consensus/istanbul/core/core.go
jbhurat: This is legacy IBFT to QBFT transition flow, as part of starting QIBFT, we try to stop legacy IBFT and that can return an error. What we decided was to continue if that happens, so that QIBFT can be started for the next sequence. 

I will take out the TODO from there and add another comment explaining this

consensus/istanbul/backend/backend.go
jbhurat: Right. I pushed a commit to fix the log

consensus/istanbul/backend/backend.go
jbhurat: Done, pushed a commit

consensus/istanbul/core/core.go
jbhurat: Done

core/blockchain.go
jbhurat: Done

docs/Consensus/ibft/ibft-parameters.md
jbhurat: Done

eth/handler.go
jbhurat: Done

consensus/istanbul/backend/backend.go
jbhurat: Actually removed `IsQIBFTConsensusCrossed()` as it looks it might not be needed

consensus/istanbul/backend/backend.go
ricardolyn: non-blocking: ok, there are 2 differences: `sb.chain.CurrentHeader()` and not checking for `sb.qibftConsensusEnabled`. is there a way that we can re-use the same? as the rest is similar.

jbhurat: I pushed a commit, that should re-use same code between the two methods

docs/Consensus/ibft/ibft-parameters.md
```diff
@@ -0,0 +1,76 @@
+# IBFT parameters
```
chris-j-h, line 1: This should be moved out to the doc.goquorum project

core/blockchain.go
```diff
@@ -1757,7 +1757,7 @@ func (bc *BlockChain) InsertChain(chain types.Blocks) (int, error) {
...
```
chris-j-h: Message prefix should be removed

consensus/protocol.go
```diff
@@ -26,7 +26,7 @@ var (
 		Versions: []uint{Istanbul100, Istanbul99, Istanbul64},
 		// istanbul/100 has to have 18 message to be backwards compatible although at the p2p layer it only has
 		// 1 message with msg.Code 17
-		Lengths: map[uint]uint64{Istanbul100: 18, Istanbul99: 18, Istanbul64: 18},
+		Lengths: map[uint]uint64{Istanbul100: 22, Istanbul99: 18, Istanbul64: 18},
```
chris-j-h: Is the requirement of length `18` for `istanbul/100` backward compatibility mentioned in the comment no longer true?  If so the comment should be updated.

chris-j-h: Are there any potential impacts for users by making this change to the `istanbul/100` protocol?

consensus/istanbul/config.go
```diff
@@ -16,29 +16,99 @@
...
+// NewStickyProposerPolicy return a Sticky ProposerPolicy with ValidatorSortByStringFunc as default sort function
+func NewStickyProposerPolicy() *ProposerPolicy {
+	return NewProposerPolicy(Sticky)
+}
+
+func NewProposerPolicy(id ProposerPolicyId) *ProposerPolicy {
+	return &ProposerPolicy{Id: id, By: ValidatorSortByStringFunc}
+}
+
+type proposerPolicyToml struct {
```
chris-j-h, line 53: may be able to achieve the same with `toml:"-"` struct tags in `ProposerPolicy`

consensus/istanbul/config.go
```diff
@@ -16,29 +16,99 @@
...
+	if err != nil {
+		return err
+	}
+	p.Id = pp.Id
+	p.By = ValidatorSortByStringFunc
+	return nil
+}
+
+// Use sets the validatorSortByFunc for the given ProposerPolicy and sorts the validatorSets according to it
+func (p *ProposerPolicy) Use(v validatorSortByFunc) {
```
chris-j-h: might be worth adding some unit tests for this because it requires pointers to always be passed around. A unit test will protect against a potential future change by an unsuspecting dev passing vals around

consensus/istanbul/backend/backend.go
```diff
@@ -316,3 +352,64 @@ func (sb *backend) HasBadProposal(hash common.Hash) bool {
...
+	}
+	if sb.chain != nil {
+		return sb.IsQBFTConsensusForHeader(sb.chain.CurrentHeader())
+	}
+	return false
+}
+
+// IsQBFTConsensusForHeader checks if qbft consensus is enabled for the block height identified by the given header
+func (sb *backend) IsQBFTConsensusForHeader(header *types.Header) bool {
+	// If qbftBlock is not defined in genesis, then use legacy ibft
```
chris-j-h: invalid comment, copy-paste error by the looks of it

consensus/istanbul/backend/backend.go
```diff
@@ -316,3 +352,64 @@ func (sb *backend) HasBadProposal(hash common.Hash) bool {
...
+		return true
+	}
+	return false
+}
+
+// qbftBlockNumber returns the qbftBlock fork block number
+func (sb *backend) qbftBlockNumber() uint64 {
+	if sb.config.QbftBlock == nil {
+		return 0
+	}
```
chris-j-h: `sb.config.QbftBlock == nil` probably should not result in forkBlock being returned `0` as `nil` config implies qbft is disabled.  Callers may not expect this behaviour.

consensus/istanbul/backend/backend_test.go
```diff
@@ -184,6 +185,35 @@ func TestGetProposer(t *testing.T) {
 	}
 }
 
+func TestIsQBFTConsensus(t *testing.T) {
+	// Uses Istanbul DefaultConfig which has qbftBlock set to 0
```
chris-j-h: delete comment as it is incorrect - `newLegacyBlockchain` overrides the default and sets `qbftBlock` to `nil`

consensus/istanbul/backend/backend_test.go
```diff
@@ -184,6 +185,35 @@ func TestGetProposer(t *testing.T) {
...
+	block := makeBlock(chain, engine, chain.Genesis())
+	_, err := chain.InsertChain(types.Blocks{block})
+	if err != nil {
+		t.Errorf("Error inserting block: %v", err)
+	}
+
+	qbftConsensus = engine.IsQBFTConsensus()
+	if !qbftConsensus {
+		t.Errorf("IsQBFTConsensus() should return true after block insertion")
+	}
```
chris-j-h, line 219: cleanup/teardown required?

consensus/istanbul/backend/engine_test.go
```diff
@@ -73,7 +75,46 @@ func newBlockChain(n int) (*core.BlockChain, *backend) {
...
+// block by one node. Otherwise, if n is larger than 1, we have to generate
+// other fake events to process Istanbul.
+func newLegacyBlockChain(n int) (*core.BlockChain, *backend) {
+	genesis, nodeKeys := getGenesisAndKeys(n, false)
+	memDB := rawdb.NewMemoryDatabase()
+	config := istanbul.DefaultConfig
+	// Use the first key as private key
+	b, _ := New(config, nodeKeys[0], memDB).(*backend)
+
+	config.QbftBlock = nil
```
chris-j-h: does this end up changing the `defaultConfig` for subsequent tests (we are changing param of package-level struct pointer)?

consensus/protocol.go
jbhurat: I will be updating the comment. I don't see any impact of updating number of messages to 22 during upgrade

docs/Consensus/ibft/ibft-parameters.md
jbhurat: @bgravenorst, can you please check if we can reuse this page when looking at qbft documentation

core/blockchain.go
jbhurat: Done

consensus/istanbul/config.go
jbhurat: Added a test for ProposerPolicy

consensus/istanbul/backend/backend.go
jbhurat: Updated the comment

consensus/istanbul/backend/backend.go
jbhurat: Returning `-1` when `QbftBlock` is not defined

consensus/istanbul/backend/backend_test.go
jbhurat: Done

consensus/protocol.go
jbhurat: I updated the comment to reflect that `istanbul/100` need to have 22 messages. In case of legacy IBFT, only 18 are used and rest 4 messages are used when QBFT consensus is enabled

docs/Consensus/ibft/ibft-parameters.md
bgravenorst: Yes, this can definitely be used.

consensus/istanbul/backend/backend.go
```diff
@@ -156,14 +161,23 @@ func (sb *backend) Gossip(valSet istanbul.ValidatorSet, payload []byte) error {
...
+
+			var outboundCode uint64 = istanbulMsg
+			if _, ok := qbftMessage.MessageCodes()[code]; ok {
+				outboundCode = code
+			}
+			if sb.IsQBFTConsensus() {
+				go p.SendQbftConsensus(outboundCode, payload)
+			} else {
+				go p.SendConsensus(outboundCode, payload)
+			}
```
baptiste-b-pegasys: I don't think it is clear that for non QBFT, the `code` is still `istanbulMsg`.
Here it will ensure it is the case:
```suggestion

			if sb.IsQBFTConsensus() {
				var outboundCode uint64 = istanbulMsg
				if _, ok := qbftMessage.MessageCodes()[code]; ok {
					outboundCode = code
				}
				go p.SendQbftConsensus(outboundCode, payload)
			} else {
				go p.SendConsensus(istanbulMsg, payload)
			}
```

consensus/istanbul/backend/backend.go
```diff
@@ -172,11 +186,28 @@ func (sb *backend) Commit(proposal istanbul.Proposal, seals [][]byte) error {
 	}
 
 	h := block.Header()
-	// Append seals into extra-data
-	err := writeCommittedSeals(h, seals)
-	if err != nil {
-		return err
+
+	// Append Round Number and seals into extra-data of qbft consensus
+	if sb.IsQBFTConsensus() {
```
baptiste-b-pegasys: for all these if-else blocs, can we use an interface and two implementations, a strategy pattern?

consensus/qbft/message/errors.go
```diff
@@ -0,0 +1,40 @@
...
+		// errFailedDecodePrepare is returned when the PREPARE message is malformed.
+		errFailedDecodePrepare = errors.New("failed to decode PREPARE")
+		// errFailedDecodeCommit is returned when the COMMIT message is malformed.
+		errFailedDecodeCommit = errors.New("failed to decode COMMIT")
+		// errFailedDecodePiggybackMsgs is returned when the Piggyback messages are malformed
+		errFailedDecodePiggybackMsgs = errors.New("failed to decode Piggyback Messages")
+		// errInvalidSigner is returned when the message is signed by a validator different than message sender
+		errInvalidSigner = errors.New("message not signed by the sender")
+		// errInvalidPreparedBlock is returned when prepared block is not validated in round change messages
+		errInvalidPreparedBlock = errors.New("invalid prepared block in round change messages")*/
```
baptiste-b-pegasys: commented code added, may not be useful
```suggestion
```

consensus/qbft/core/backlog.go
```diff
@@ -0,0 +1,177 @@
+// Copyright 2017 The go-ethereum Authors
```
baptiste-b-pegasys: so if there is a qbft package, why would the istanbul package be impacted?

baptiste-b-pegasys (dismissed):

consensus/istanbul/backend/backend.go
ricardolyn: yes, that will be a better approach. the implementation can be injected into the backend

consensus/istanbul/backend/backend.go
```diff
@@ -316,3 +352,63 @@ func (sb *backend) HasBadProposal(hash common.Hash) bool {
...
+	}
+	sb.logger.Trace("Stopped legacy IBFT consensus")
+	sb.coreMu.Lock()
+	defer sb.coreMu.Unlock()
+	// Set the core to qbft
+	sb.core = qbftCore.New(sb, sb.config)
+
+	sb.logger.Trace("Starting qbft")
+	sb.config.ProposerPolicy.Use(istanbul.ValidatorSortByByteFunc)
+	if err := sb.core.Start(); err != nil {
```
ricardolyn: if this fails and we wait for the next block, how is the legacy IBFT is restarted?

consensus/qbft/core/backlog.go
jbhurat: The initial backend work for both `istanbul` and `qbft` consensus is common, hence instead of duplicating that piece of code, `qbft` uses that as well. There are few changes though around how the `extradata` fields are populated and verified, hence `istanbul` package is impacted

consensus/istanbul/backend/backend_test.go
jbhurat: Pushed a commit to stop the engine at the end of test

consensus/istanbul/backend/engine_test.go
jbhurat: Creating a copy of the config and acting on it instead of DefaultConfig

consensus/istanbul/backend/backend.go
jbhurat: Done

consensus/qbft/message/errors.go
jbhurat: Removed commented error messages

consensus/istanbul/backend/backend.go
nmvalera: :+1:

consensus/istanbul/backend/backend.go
```diff
@@ -316,3 +352,63 @@ func (sb *backend) HasBadProposal(hash common.Hash) bool {
...
+func (sb *backend) IsQBFTConsensusForHeader(header *types.Header) bool {
+	// If qbftBlock is not defined in genesis qbft consensus is not used
+	if sb.config.QbftBlock == nil {
+		return false
+	}
+	if sb.config.QbftBlock.Uint64() == 0 {
+		return true
+	}
+
+	if sb.chain != nil && header.Number.Cmp(sb.config.QbftBlock) >= 0 {
```
nmvalera: Why do we need  `sb.chain != nil` ?

consensus/istanbul/backend/backend.go
```diff
@@ -316,3 +352,63 @@ func (sb *backend) HasBadProposal(hash common.Hash) bool {
...
+		return true
+	}
+	if sb.chain != nil {
+		return sb.IsQBFTConsensusForHeader(sb.chain.CurrentHeader())
+	}
+	return false
+}
+
+// IsQBFTConsensusForHeader checks if qbft consensus is enabled for the block height identified by the given header
+func (sb *backend) IsQBFTConsensusForHeader(header *types.Header) bool {
```
nmvalera: Could we not implement `IsQBFTConsensusForHeader` on `istanbul.Config` object?

consensus/istanbul/backend/backend.go
```diff
@@ -316,3 +352,63 @@ func (sb *backend) HasBadProposal(hash common.Hash) bool {
...
+	}
+
+	if sb.chain != nil && header.Number.Cmp(sb.config.QbftBlock) >= 0 {
+		return true
+	}
+	return false
+}
+
+// qbftBlockNumber returns the qbftBlock fork block number, returns -1 if qbftBlock is not defined
+func (sb *backend) qbftBlockNumber() int64 {
```
nmvalera: Could we not implement `qbftBlockNumber` this on istanbul.Config object?

Something like
```go
func (sb *Backend) qbftBlockNumber() int64 {
	return sb.config.QBFTBlock()
}
```

consensus/istanbul/backend/engine.go
```diff
@@ -101,29 +102,54 @@ var (
...
 }
 
 // Signers extracts all the addresses who have signed the given header
 // It will extract for each seal who signed it, regardless of if the seal is
 // repeated
-func (sb *backend) Signers(header *types.Header) ([]common.Address, error) {
-	extra, err := types.ExtractIstanbulExtra(header)
-	if err != nil {
-		return []common.Address{}, err
+func (sb *backend) Signers(header *types.Header, isQbft bool) ([]common.Address, error) {
```
nmvalera: Does `Signers` needs to be a method on the backend object? Feels like it would be clearer implemented as functions on `istanbul/core` and `qbft/core` (also would increase re-usability).

consensus/istanbul/backend/engine.go
```diff
@@ -296,19 +327,30 @@ func (sb *backend) verifyCommittedSeals(chain consensus.ChainHeaderReader, heade
 		return err
 	}
 
-	extra, err := types.ExtractIstanbulExtra(header)
-	if err != nil {
-		return err
+	var committedSeal [][]byte
+	if sb.IsQBFTConsensus() {
```
nmvalera: Same remark as above instead of having `if` blocks we could implement `VerifyCommittedSeals` on `instanbul/core` and `qbft/core` maybe (taking `ValSet` as input so we do not need to call `sb.snapshot(...)`

consensus/istanbul/backend/engine.go
```diff
@@ -101,29 +102,54 @@ var (
 // block, which may be different from the header's coinbase if a consensus
 // engine is based on signatures.
 func (sb *backend) Author(header *types.Header) (common.Address, error) {
-	return ecrecover(header)
+	if sb.IsQBFTConsensus() {
```
nmvalera: We could implement `Author()`  on `instanbul/core` and `qbft/core` so we do not need `if` block

consensus/istanbul/backend/engine.go
```diff
@@ -152,14 +178,19 @@ func (sb *backend) verifyHeader(chain consensus.ChainHeaderReader, header *types
 	}
 
 	// Ensure that the extra data format is satisfied
-	if _, err := types.ExtractIstanbulExtra(header); err != nil {
-		return errInvalidExtraDataFormat
+	if sb.IsQBFTConsensus() {
```
nmvalera: We could implement `VerifyHeader`() on instanbul/core and qbft/core so we do not need if block

consensus/istanbul/backend/engine.go
```diff
@@ -347,6 +389,10 @@ func (sb *backend) VerifySeal(chain consensus.ChainHeaderReader, header *types.H
 // Prepare initializes the consensus fields of a block header according to the
 // rules of a particular engine. The changes are executed inline.
 func (sb *backend) Prepare(chain consensus.ChainHeaderReader, header *types.Header) error {
+	// Check if QBFT Consensus is used, call the QbftPrepare() is that is the case
+	if sb.IsQBFTConsensus() {
```
nmvalera: Same remark as above instead of having if blocks we could implement `Prepare` on instanbul/core and qbft/core maybe

consensus/istanbul/backend/engine.go
```diff
@@ -498,15 +606,19 @@ func (sb *backend) Seal(chain consensus.ChainHeaderReader, block *types.Block, r
 // update timestamp and signature of the block based on its number of transactions
 func (sb *backend) updateBlock(parent *types.Header, block *types.Block) (*types.Block, error) {
 	header := block.Header()
-	// sign the hash
-	seal, err := sb.Sign(sigHash(header).Bytes())
-	if err != nil {
-		return nil, err
-	}
+	if sb.IsQBFTConsensus() {
```
nmvalera: `UpdateBlock()` could be implemented on `istanbul/core` and `qbft/core`

consensus/istanbul/backend/api.go
```diff
@@ -83,7 +83,7 @@ func (api *API) signers(header *types.Header) (*BlockSigners, error) {
 		return nil, err
 	}
 
-	committers, err := api.istanbul.Signers(header)
+	committers, err := api.istanbul.Signers(header, api.istanbul.IsQBFTConsensusForHeader(header))
```
nmvalera: Why do we need to pass `api.istanbul.IsQBFTConsensusForHeader(header)` input? We are calling a function `api.istanbul.Signers(...)` which should already be aware of it.

consensus/istanbul/backend/backend.go
```diff
@@ -316,3 +352,63 @@ func (sb *backend) HasBadProposal(hash common.Hash) bool {
...
+	if sb.config.QbftBlock == nil {
+		return -1
+	}
+	return sb.config.QbftBlock.Int64()
+}
+
+// StartQBFTConsensus stops existing legacy ibft consensus and starts the new qbft consensus
+func (sb *backend) StartQBFTConsensus() error {
```
nmvalera: To be covered when updating migrations

`StartQBFTConsensus` is a bit misleading as it also stops IBFT.

I think it would be clearer to have independant functions
- `StartQBFTConsensus`
- `StopQBFTConsensus`
- `StartIBFTConsensus`
- `StopIBFTConsensus`

consensus/istanbul/backend/engine.go
```diff
@@ -406,6 +452,69 @@ func (sb *backend) Prepare(chain consensus.ChainHeaderReader, header *types.Head
 	return nil
 }
 
+// Prepare initializes the consensus fields of a block header according to the
+// rules of a particular engine. The changes are executed inline.
+func (sb *backend) QbftPrepare(chain consensus.ChainHeaderReader, header *types.Header) error {
```
nmvalera: `QbftPrepare` could be implemented on `qbft/core` maybe taking different arguments `chainHead` and `ValSet` so we do not need access to `sb`

consensus/istanbul/backend/engine.go
```diff
@@ -541,11 +653,26 @@ func (sb *backend) Start(chain consensus.ChainHeaderReader, currentBlock func()
 	sb.currentBlock = currentBlock
 	sb.hasBadBlock = hasBadBlock
 
+	// Check if qbft Consensus needs to be used after chain is set
+	if sb.IsQBFTConsensus() {
+		sb.core = qbftCore.New(sb, sb.config)
```
nmvalera: We could reuse functions `StartQBFT` and `StartQBFT` suggested above

consensus/istanbul/backend/engine.go
```diff
@@ -541,11 +653,26 @@ func (sb *backend) Start(chain consensus.ChainHeaderReader, currentBlock func()
 	sb.currentBlock = currentBlock
 	sb.hasBadBlock = hasBadBlock
 
+	// Check if qbft Consensus needs to be used after chain is set
+	if sb.IsQBFTConsensus() {
```
nmvalera, line 272: It feels like this is the only place where `if sb.IsQBFTConsensus()` is necessary (so we can set the expected `sb.core`) then all other `if sb.IsQBFTConsensus()` could be avoided by implementing suggested method on the `Core` interface itself.

consensus/istanbul/backend/engine.go
```diff
@@ -541,11 +653,26 @@ func (sb *backend) Start(chain consensus.ChainHeaderReader, currentBlock func()
...
+	}
+
 	if err := sb.core.Start(); err != nil {
 		return err
 	}
 
 	sb.coreStarted = true
+
+	if sb.IsQBFTConsensus() {
+		sb.logger.Trace("Started qbft consensus")
```
nmvalera: By using `StartQBFT` and `StartIBFT` this block could be avoided

consensus/istanbul/backend/handler.go
```diff
@@ -49,18 +53,24 @@ func (sb *backend) Protocol() consensus.Protocol {
 
 func (sb *backend) decode(msg p2p.Msg) ([]byte, common.Hash, error) {
 	var data []byte
-	if err := msg.Decode(&data); err != nil {
-		return nil, common.Hash{}, errDecodeFailed
+	if sb.IsQBFTConsensus() {
```
nmvalera, line 55: Is it possible that we have switched to QBFT but still receive IBFT messages?

consensus/istanbul/backend/snapshot.go
```diff
@@ -178,84 +178,191 @@ func (s *Snapshot) apply(headers []*types.Header) (*Snapshot, error) {
...
+			}
 		}
+	}
+	snap.Number += uint64(len(headers))
+	snap.Hash = headers[len(headers)-1].Hash()
+
+	return snap, nil
+}
+
+func (s *Snapshot) legacyApply(header *types.Header) error {
```
nmvalera: In a concern of separating concerns which are proper to `qbft` and `ibft` should we move `legacyApply` under `istanbul` package? Maybe `ApplyHeaderToSnap(header *types.Header, s *backend.Snapshot)` ?

consensus/istanbul/backend/snapshot.go
```diff
@@ -178,84 +178,191 @@ func (s *Snapshot) apply(headers []*types.Header) (*Snapshot, error) {
...
-				Authorize: authorize,
-			})
+		delete(s.Tally, header.Coinbase)
+	}
+	return nil
+}
+
+// qbftApply creates a new authorization snapshot using qbftExtra by applying the given headers to
+// the original one.
+func (s *Snapshot) qbftApply(header *types.Header) error {
```
nmvalera: Same remarks

consensus/istanbul/validator.go
```diff
@@ -34,16 +36,39 @@ type Validator interface {
...
+}
+
+type validatorSorter struct {
+	validators Validators
+	by         validatorSortByFunc
+}
+
+type validatorSortByFunc func(v1 Validator, v2 Validator) bool
+
+var ValidatorSortByStringFunc validatorSortByFunc = func(v1 Validator, v2 Validator) bool {
```
nmvalera: We should not declare a function as a global variablen this is quite error prone as any other place in the code could reset it.

Instead, you could use a function, something like

`func ValidatorSortByString() validatorSortByFunc {
    return func(v1 Validator, v2 Validator) bool {...}
}` 

It may require to export type `validatorSortByFunc`

consensus/istanbul/core/core.go
```diff
@@ -197,6 +197,17 @@ func (c *core) startNewRound(round *big.Int) {
 		logger = c.logger.New("old_round", c.current.Round(), "old_seq", c.current.Sequence())
 	}
 
+	logger.Trace("Start new ibft round")
+
+	// If new round is 0, then check if qbftConsensus needs to be enabled
+	if round.Uint64() == 0 && c.backend.IsQBFTConsensus() {
```
nmvalera: To be covered when updating migrations

Why is the switch from IBFT to QBFT triggered here? Should it not be triggered by in `backend.Commit()`? So concerns are well separated between `ibft` and `qbft`?

consensus/istanbul/core/final_committed.go
```diff
@@ -21,6 +21,6 @@ import "github.com/ethereum/go-ethereum/common"
 func (c *core) handleFinalCommitted() error {
 	logger := c.logger.New("state", c.state)
 	logger.Trace("Received a final committed proposal")
-	c.startNewRound(common.Big0)
+	go c.startNewRound(common.Big0)
```
nmvalera, line 24: Why do we need a goroutine that was not there before?

consensus/protocol.go
```diff
@@ -74,4 +74,7 @@ type Peer interface {
 
 	// SendConsensus sends the message to this p2p peer using the consensus specific devp2p subprotocol
 	SendConsensus(msgcode uint64, data interface{}) error
+
+	// SendQbftConsensus is used to send consensus subprotocol messages from an "eth" peer without encoding the payload
+	SendQbftConsensus(msgcode uint64, payload []byte) error
```
nmvalera: Should we uppercase `Qbft`: `SendQbftConsensus` -> `SendQBFTConsensus`

consensus/qbft/core/backlog.go
```diff
@@ -0,0 +1,177 @@
...
+func (c *core) storeQBFTBacklog(msg message.QBFTMessage) {
+	src := msg.Source()
+	logger := c.logger.New("from", src, "state", c.state)
+
+	if src == c.Address() {
+		logger.Warn("Backlog from self")
+		return
+	}
+
+	logger.Trace("Store future message")
```
nmvalera: "Store future backlog message"

consensus/qbft/core/backlog.go
```diff
@@ -0,0 +1,177 @@
...
+
+	c.backlogsMu.Lock()
+	defer c.backlogsMu.Unlock()
+
+	logger.Debug("Retrieving backlog queue", "for", src, "backlogs_size", len(c.backlogs))
+	backlog := c.backlogs[src]
+	if backlog == nil {
+		backlog = prque.New()
```
nmvalera: `c.backlogs[src] = backlog` could be done under the `if` statement for clarity

consensus/istanbul/config.go
```diff
@@ -16,29 +16,99 @@
...
-	RoundRobin ProposerPolicy = iota
+	RoundRobin ProposerPolicyId = iota
 	Sticky
 )
 
+// ProposerPolicy represents the Validator Proposer Policy
+type ProposerPolicy struct {
+	Id       ProposerPolicyId    // Could be RoundRobin or Sticky
+	Registry []ValidatorSet      // Holds the ValidatorSet for a given block height
+	By       validatorSortByFunc // func that defines how the ValidatorSet should be sorted
```
nmvalera: Look at this during update to migration

This `By validatorSortByFunc` looks to be quite error prone as it can be set anywhere in the code and second it assumes that an engineer knows some implementation details such as the difference of sorting between `qbft` and `ibft`

Instead, could we use an unexported `type string` field which value can be either `qbft` or `ibft`. Then `ProposerPolicy` exposes a method `SortValidators` that uses the underlying correct `validatorSortByFunc` depending on the `type`

consensus/qbft/core/core.go
```diff
@@ -0,0 +1,304 @@
...
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/event"
+	"github.com/ethereum/go-ethereum/log"
+	metrics "github.com/ethereum/go-ethereum/metrics"
+	"gopkg.in/karalabe/cookiejar.v2/collections/prque"
+)
+
+// New creates an Istanbul consensus core
+func New(backend istanbul.Backend, config *istanbul.Config) Engine {
+	r := metrics.NewRegistry()
```
nmvalera: Why creating a new registry? Should we not use the global one similarly to other places in the code?

raft/handler.go
```diff
@@ -487,6 +487,10 @@ func (pm *ProtocolManager) startRaft() {
 			for _, entry := range entries {
 				if entry.Type == raftpb.EntryNormal {
 					var block types.Block
+					if len(entry.Data) == 0 {
```
nmvalera: Is this related to QBFT? Should this be covered as part of a separated PR?

consensus/qbft/core/qbft_msg_set.go
```diff
@@ -0,0 +1,162 @@
...
+import (
+	"fmt"
+	"io"
+	"math/big"
+	"strings"
+	"sync"
+
+	istanbul2 "github.com/ethereum/go-ethereum/consensus/qbft"
+	"github.com/ethereum/go-ethereum/consensus/qbft/message"
+
```
nmvalera: Unnecessary skipped line

consensus/qbft/message/prepare.go
```diff
@@ -0,0 +1,88 @@
...
+	"io"
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/rlp"
+)
+
+// A QBFT PREPARE message.
+type Prepare struct {
+	SignedPreparePayload
```
nmvalera: Why do we need intermediary type `SignedPreparePayload` ?

consensus/qbft/core/justification.go
```diff
@@ -0,0 +1,112 @@
+package core
+
+import (
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/consensus/qbft/message"
+	"github.com/ethereum/go-ethereum/log"
+
```
nmvalera: Unnecessary skipped line

consensus/qbft/core/justification.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/ethereum/go-ethereum/common"
+
```
nmvalera: Unnecessary skipped line

consensus/qbft/core/justification.go
```diff
@@ -0,0 +1,112 @@
...
+)
+
+// Returns true if the `proposal` is justified by the set `roundChangeMessages` of ROUND-CHANGE messages
+// and by the set `prepareMessages` of PREPARE messages.
+// For this we must either have:
+//     - a quorum of ROUND-CHANGE messages with preparedRound and preparedBlockDigest equal to nil; or
+//     - a ROUND-CHANGE message (1) whose preparedRound is not nil and is equal or higher than the
+//           preparedRound of `quorumSize` ROUND-CHANGE messages and (2) whose preparedRound and
+//           preparedBlockDigest match the round and block of `quorumSize` PREPARE messages.
+func justify(
```
nmvalera: Should it be named `isJustified` ?

consensus/qbft/core/justification.go
```diff
@@ -0,0 +1,112 @@
...
+
+	// If there are PREPARE messages, they all need to have the same round and match `proposal`
+	var preparedRound *big.Int
+	iteration := 0
+	for _, spp := range prepareMessages {
+		if iteration == 0 { // Get the round of the first message
+			preparedRound = spp.Round
+		}
+		if preparedRound.Cmp(spp.Round) != 0 || proposal.Hash() != spp.Digest {
+			return false
```
nmvalera: Instead of returning a bool, it could return an error with extra details about the reason for not being justified.

consensus/qbft/core/justification.go
```diff
@@ -0,0 +1,112 @@
...
+	// Check the size of the set of PREPARE messages
+	if len(prepareMessages) != 0 && len(prepareMessages) < quorumSize {
+		return false
+	}
+
+	// If there are PREPARE messages, they all need to have the same round and match `proposal`
+	var preparedRound *big.Int
+	iteration := 0
+	for _, spp := range prepareMessages {
+		if iteration == 0 { // Get the round of the first message
```
nmvalera: It seems that the iteration variabl could be avoided by doing something like

```go
var preparedRound *big.Int
if len(prepareMessages) > 0 {
    preparedRound = prepareMessages[0].Round
    for _, spp := range prepareMessages {
        if preparedRound.Cmp(spp.Round) != 0 || proposal.Hash() != spp.Digest {
            return false
        }
    }
}

File: consensus/qbft/core/preprepare.go [hunk 65]
```diff
@@ -0,0 +1,128 @@
...
+
+		// Set the preprepareSent to the current round
+		c.current.preprepareSent = curView.Round
+	}
+}
+
+func (c *core) handlePreprepareMsg(preprepare *message.Preprepare) error {
+	logger := c.logger.New("state", c.state)
+
+	c.logger.Info("QBFT: handlePreprepareMsg", "view", preprepare.View(), "m", preprepare)
```
nmvalera: Is there a rationale for using `c.logger` and not `logger` ?

consensus/qbft/core/preprepare.go
```diff
@@ -0,0 +1,128 @@
...
+
+	// Check if the message comes from current proposer
+	logger.Warn("QBFT who's proposer?", "source", preprepare.Source(), "proposer", c.valSet.GetProposer().Address())
```
nmvalera: If I understand correctly this will log a warning message on every received `preprepare` message.

Instead should we log source address on the above Info log "QBFT: handlePreprepareMsg" and Warn only if the proposer is invalid

consensus/qbft/core/preprepare.go
```diff
@@ -0,0 +1,128 @@
...
+	if !c.valSet.IsProposer(preprepare.Source()) {
+		logger.Warn("Ignore preprepare messages from non-proposer")
+		return errNotFromProposer
+	}
+
+	// Justification
+	if preprepare.Round.Uint64() > 0 && !justify(preprepare.Proposal, preprepare.JustificationRoundChanges, preprepare.JustificationPrepares, c.QuorumSize()) {
+		logger.Error("QBFT: Unable to justify PRE-PREPARE message")
```
nmvalera: Should the message better be something like "QBFT: invalid PRE-PREPARE message's justification"?

nmvalera: Also, should it be a warning (same as for invalid proposer)?

consensus/qbft/core/preprepare.go
nmvalera: Log message could be cleaned:

```
.Info("QBFT: handle PRE-PREPARE message", "view", preprepare.View(), "msg", preprepare)
```

consensus/qbft/core/preprepare.go
```diff
@@ -0,0 +1,128 @@
...
+	// If I'm the proposer and I have the same sequence with the proposal
+	if c.current.Sequence().Cmp(request.Proposal.Number()) == 0 && c.IsProposer() {
+		curView := c.currentView()
+		preprepare := message.NewPreprepare(curView.Sequence, curView.Round, request.Proposal)
+		preprepare.SetSource(c.Address())
+
+		// Sign payload
+		encodedPayload, err := preprepare.EncodePayloadForSigning()
+		if err != nil {
+			logger.Error("QBFT: Failed to encode payload of pre-prepare message", "msg", preprepare, "err", err)
```
nmvalera: for consistency between log message we should decide whether to always use a message name which is lower case or upper-case.

`pre-prepare` or `PRE-PREPARE`

consensus/qbft/core/preprepare.go
```diff
@@ -0,0 +1,128 @@
...
+	if preprepare.Round.Uint64() > 0 && !justify(preprepare.Proposal, preprepare.JustificationRoundChanges, preprepare.JustificationPrepares, c.QuorumSize()) {
+		logger.Error("QBFT: Unable to justify PRE-PREPARE message")
+		return errInvalidPreparedBlock
+	}
+
+	// Verify the proposal we received
+	if duration, err := c.backend.Verify(preprepare.Proposal); err != nil {
+		// if it's a future block, we will handle it again after the duration
+		if err == consensus.ErrFutureBlock {
+			logger.Info("Proposed block will be handled in the future", "err", err, "duration", duration)
```
nmvalera: Should message has a `QBFT: ` prefix?

consensus/qbft/core/roundchange.go
```diff
@@ -0,0 +1,291 @@
...
+package core
+
+import (
+	"math/big"
+	"sort"
+	"sync"
+
+	"github.com/ethereum/go-ethereum/consensus/qbft/message"
+	"github.com/ethereum/go-ethereum/log"
+
```
nmvalera: Unecessary skipped line

consensus/qbft/core/roundchange.go
```diff
@@ -0,0 +1,291 @@
...
+			rcs.prepareMessages[round] = prepareMessages
+		}
+	}
+
+	return nil
+}
+
+// higherRoundMessages returns the number of Round Change messages received for the round greater than the given round
+// and from different validators
+func (rcs *roundChangeSet) higherRoundMessages(round *big.Int) int {
```
nmvalera: I am not sure to completely understand what is this function expected to return. Could you please elaborate?

common/types.go
```diff
@@ -182,6 +182,10 @@ func (h *Hash) SetBytes(b []byte) {
 	copy(h[HashLength-len(b):], b)
 }
 
+func (h *Hash) IsEmpty() bool {
+	return *h == Hash{}
```
baptiste-b-pegasys: should we move `Hash{}` in a constant so we don't recreate everytime a ressource?
we could also call `EmptyHash(*h)` below

consensus/istanbul/backend/backend.go
baptiste-b-pegasys: a mistake? sb.config?
What if header is nil?

consensus/qbft/core/roundchange.go
jbhurat: `higherRoundMessages()` returns the number of round change messages which are higher than the given round.

core/types/block.go
```diff
@@ -293,6 +298,9 @@ func CopyHeader(h *Header) *Header {
 func (b *Block) DecodeRLP(s *rlp.Stream) error {
 	var eb extblock
 	_, size, _ := s.Kind()
+	if size == 0 {
```
nmvalera: @jbhurat It looks that this change is making some unit tests fail. What was the rationale for adding this check? Thanks!

jbhurat: This was added to support for nil blocks in Piggyback messages that were sent along with `RoundChange` and `Preprepare` messages. The logic for Piggyback messages was updated, but this was not updated. I have pushed a commit to remove this

raft/handler.go
jbhurat: This should no longer be needed after the `DecodeRLP()` changes are reverted. I have pushed a commit to remove this

consensus/qbft/core/roundchange.go
```diff
@@ -0,0 +1,291 @@
...
+	log.Info("QBFT: handleRoundChange count", "higherRoundMsgs", num, "currentRoundMsgs", currentRoundMessages)
+	if num == c.valSet.F()+1 {
+		newRound := c.roundChangeSet.getMinRoundChange(currentRound)
+		logger.Trace("Starting new Round", "round", newRound)
+		c.startNewRound(newRound)
+		c.broadcastRoundChange(newRound)
+	} else if currentRoundMessages >= c.QuorumSize() && c.IsProposer() && c.current.preprepareSent.Cmp(currentRound) < 0 {
+		_, proposal := c.highestPrepared(currentRound)
+		if proposal == nil {
+			proposal = c.current.pendingRequest.Proposal
```
nmvalera: I have seen some flaky panic here because `c.current.pendingRequest` is `nil`.

Should we extend if check? 

```go
if proposal == nil && c.current.pendingRequest != nil
```

consensus/qbft/core/core.go
```diff
@@ -0,0 +1,304 @@
...
+		return false
+	}
+	return v.IsProposer(c.backend.Address())
+}
+
+func (c *core) IsCurrentProposal(blockHash common.Hash) bool {
+	return c.current != nil && c.current.pendingRequest != nil && c.current.pendingRequest.Proposal.Hash() == blockHash
+}
+
+func (c *core) commitQBFT() {
```
nmvalera: Should `commitQBFT` be moved to `commit.go` ?

consensus/qbft/core/justification.go
jbhurat: Done

consensus/qbft/core/justification.go
jbhurat: Done

consensus/qbft/core/justification.go
jbhurat: Done

consensus/qbft/core/justification.go
jbhurat: Done

common/types.go
jbhurat: Done

consensus/qbft/core/core.go
jbhurat: Done

consensus/qbft/core/roundchange.go
jbhurat: Done

consensus/qbft/core/qbft_msg_set.go
jbhurat: Done

consensus/istanbul/backend/handler.go
jbhurat: Each node will move to QBFT when the `qbftBlock` fork block is reached, it may happen that when a node starts processing QBFT messages, it may receive a IBFT message, which will result in an error decoding the message and the message will be ignored.

consensus/istanbul/core/final_committed.go
jbhurat: During network transition, the legacy IBFT network is first stopped before starting QBFT consensus. As part of stopping legacy IBFT network, the engine is stopped as well and all the events are unsubscribed. Stopping of the engine is blocked until `core/handler.go::handleEvents()` is exited. The goroutine above makes sure `handleEvents()` exits due to unsubscribing of events and legacy IBFT engine is stopped

core/types/istanbul.go
jbhurat: Done

consensus/istanbul/backend/backend.go
jbhurat: Done

consensus/istanbul/backend/backend.go
jbhurat: Done

consensus/istanbul/backend/backend.go
jbhurat: Done

consensus/istanbul/validator.go
jbhurat: Done

consensus/qbft/message/prepare.go
jbhurat: Pushed a commit to remove `SignedPreparePayload`

consensus/qbft/core/backlog.go
jbhurat: Moved `c.backlogs[src] = backlog` to under `if` statement

consensus/istanbul/backend/backend.go
jbhurat: Currently we do not have a way to restart the legacy `IBFT` when the `QBFT` start fails. We can address that when we work on the updates to migration

consensus/istanbul/config.go
jbhurat: The intermediary struct `proposerPolicyToml` is still required, otherwise the Unmarshall goes into an infinite loop of calling itself

consensus/protocol.go
jbhurat: The idea here was to have function names in camelcase so they are more readable. I can update the names if you think `SendQBFTConsensus` would be more readable

nmvalera (dismissed):
LGTM

consensus/istanbul/backend/backend_test.go
```diff
@@ -184,6 +188,43 @@ func TestGetProposer(t *testing.T) {
...
+func TestIsQBFTConsensus(t *testing.T) {
+	chain, engine := newBlockChain(1, big.NewInt(2))
+	defer engine.Stop()
+	qbftConsensus := engine.IsQBFTConsensus()
+	if qbftConsensus {
+		t.Errorf("IsQBFTConsensus() should return false")
+	}
+
+	// Set the value of qbftBlock to 1
+	qbftConsensus = engine.IsQBFTConsensus()
```
ricardolyn: isn't duplicated this code block? as there is one exactly the same above and no state changes in the `engine`. Moreover, the comment above seems unrelated

jbhurat: Removed duplicate code block

ricardolyn (dismissed):

consensus/istanbul/backend/handler.go
```diff
@@ -109,12 +120,12 @@ func (sb *backend) HandleMsg(addr common.Address, msg p2p.Msg) (bool, error) {
 				TD    *big.Int
 			}
 			if err := msg.Decode(&request); err != nil {
-				log.Debug("Proposer was unable to decode the NewBlockMsg", "error", err)
+				log.Error("BFT: unable to decode the NewBlockMsg", "error", err)
```
baptiste-b-pegasys: line 110, `log.Debug` became `sb.logger.Debug`, why not here and at line 128?

consensus/istanbul/backend/snapshot.go
```diff
@@ -73,6 +72,7 @@ func newSnapshot(epoch uint64, number uint64, hash common.Hash, valSet istanbul.
 }
 
 // loadSnapshot loads an existing snapshot from the database.
+
```
baptiste-b-pegasys:
```suggestion
```

consensus/istanbul/ibft/core/handler.go
```diff
@@ -19,18 +19,20 @@ package core
...
-	// Start a new round from last sequence + 1
-	c.startNewRound(common.Big0)
-
 	// Tests will handle events itself, so we have to make subscribeEvents()
 	// be able to call in test.
 	c.subscribeEvents()
 	go c.handleEvents()
 
+	// Start a new round from last sequence + 1
+	c.startNewRound(common.Big0)
```
baptiste-b-pegasys, line 34: why not like the line in final_committed.go:24?
```suggestion
	go c.startNewRound(common.Big0)
```

jbhurat: `c.startNewRound(common.Big0)` in `consensus/istanbul/ibft/core/handler.go` is called when the Consensus Engine is first started and doesn't need to called in a goroutine, where as in `final_commited.go` it is called in a goroutine because during network transition, the legacy IBFT network is first stopped before starting QBFT consensus. As part of stopping legacy IBFT network, the engine is stopped as well and all the events are unsubscribed. Stopping of the engine is blocked until core/handler.go::handleEvents() is exited. The goroutine above makes sure handleEvents() exits due to unsubscribing of events and legacy IBFT engine is stopped

consensus/istanbul/backend/handler.go
jbhurat: Updated to use `sb.logger.Debug`

consensus/istanbul/backend/snapshot.go
jbhurat: Done

ricardolyn (approved):

consensus/istanbul/ibft/core/handler.go
nmvalera: Resolving conversation.

If updates are required we will cover them as part of the migration work.

PR #1215, opened 2021-06-16
Add Revert Reason to eth_getTransactionReceipt

ricardolyn:
## Summary

In smart contracts, the revert operation triggers an exception to flag an error and revert the current call. The EVM passes back to the client an optional string message containing information about the error.

Use the `--revertreason` command line option to include the revert reason in the transaction receipt response.

With revert reason enabled, the transaction receipt returned by eth_getTransactionReceipt includes the revert reason as an ABI-encoded string.

Example:
```
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
     "blockHash": "0xe7212a92cfb9b06addc80dec2a0dfae9ea94fd344efeb157c41e12994fcad60a",
     "blockNumber": "0x50",
     "contractAddress": null,
     "cumulativeGasUsed": "0x5208",
     "from": "0x627306090abab3a6e1400e9345bc60c78a8bef57",
     "gasUsed": "0x5208",
     "logs": [],
     "logsBloom": "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
     "status": "0x1",
     "to": "0xf17f52151ebef6c7334fad080c5704d77216b732",
     "transactionHash": "0xc00e97af59c6f88de163306935f7682af1a34c67245e414537d02e422815efc3",
     "transactionIndex": "0x0",    
"revertReason":"0x08c379a00000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000001a4e6f7420656e6f7567682045746865722070726f76696465642e000000000000"
   }
}
```

## Changes

* Save the revert reason into the Receipt when applying a transaction on the state db.
* Include the revert reason when saving to DB, supporting schemas without it (in both normal tx and MPS).
* Add new CLI argument to enable/disable the saving in DB.
* Note: upgrading to this version and enabling this functionality, saved receipts can't be parsed back by previous versions. full sync is required.


core/state_processor.go
```diff
@@ -305,5 +307,19 @@ func ApplyTransaction(config *params.ChainConfig, bc *BlockChain, author *common
 		privateReceipt.Bloom = types.CreateBloom(types.Receipts{privateReceipt})
 	}
 
+	// Save revert reason if feature enabled
+	if bc != nil && bc.saveRevertReason {
+		revertReasonBytes := result.Revert()
+		if revertReasonBytes != nil {
```
baptiste-b-pegasys: useful? as it is just declared the line before

baptiste-b-pegasys: sorry, I read {}, I'm tired

core/types/receipt.go
```diff
@@ -74,6 +74,11 @@ type Receipt struct {
 	//
 	// This nested structure would not have more than 2 levels.
 	PSReceipts map[PrivateStateIdentifier]*Receipt `json:"-"`
+
+	// TODO ricardolyn: do not include this as part of the receipt hash in the tried saved in the block
```
achraf17: is this TODO still required?

vdamle:
Thanks for this feature @ricardolyn , it's really useful and more straight forward than executing `eth_call` again! On Besu, back when this was introduced, we had seen that there was a warning stating this feature increases Memory consumption significantly: https://besu.hyperledger.org/en/1.2.1/Using-Pantheon/Transactions/Revert-Reason/ . Are you aware of any similar considerations for Quorum?

core/blockchain.go
```diff
@@ -2665,6 +2669,17 @@ func (bc *BlockChain) SubscribeBlockProcessingEvent(ch chan<- bool) event.Subscr
...
 	return nil, bc.isMultitenant
 }
+
+// Quorum
+func (bc *BlockChain) SaveRevertReason(enabled bool) {
+	bc.saveRevertReason = enabled
+}
+
+// Quorum
+func (bc *BlockChain) SaveRevertReasonEnabled() bool {
```
dolanor: I would follow the [effective Go for getters/setters](https://golang.org/doc/effective_go#Getters)
```suggestion
func (bc *BlockChain) SaveRevertReason() bool {
```

core/blockchain.go
```diff
@@ -2665,6 +2669,17 @@ func (bc *BlockChain) SubscribeBlockProcessingEvent(ch chan<- bool) event.Subscr
 	return bc.scope.Track(bc.blockProcFeed.Subscribe(ch))
 }
 
+// Quorum
 func (bc *BlockChain) SupportsMultitenancy(context.Context) (*proto.PreAuthenticatedAuthenticationToken, bool) {
 	return nil, bc.isMultitenant
 }
+
+// Quorum
+func (bc *BlockChain) SaveRevertReason(enabled bool) {
```
dolanor: I would follow the [effective Go for getters/setters](https://golang.org/doc/effective_go#Getters).
```suggestion
func (bc *BlockChain) SetSaveRevertReason(enabled bool) {
```

core/types/receipt.go
ricardolyn: removing, not required.

core/types/receipt.go
```diff
@@ -93,6 +97,23 @@ type receiptRLP struct {
 	Logs              []*Log
 }
 
+// Quorum
+
+// storedMPSReceiptRLPWithRevertReason is the storage encoding of a receipt which contains
+// receipts per PSI, including Revert Reason
+type storedMPSReceiptRLPWithRevertReason struct {
```
baptiste-b-pegasys: should we put all quorum additions in separate file(s)?
- receipt_revert_reasons.go?

or put all additionnal functions at the end of the file?

ricardolyn: I added in that place as we had already before the struct for MPS here. I do prefer to move to the end. what would you recommend to move?

about another file, I didn't do it to avoid making it public.

baptiste-b-pegasys: sure if we continue to add at the same place in order to have less diff, I have nothing to move.

It may be outside the scope of the PR then, if we do some reorganization of blocs.

I would have put all quorum types at the bottom of the original ones, after line 137.
All new GoQ's functions after the original ones, even if they are not grouped.

An idea would be to add some comments that link to the others functions (I will search if we can do some hypertext links in source code with the IDEs)

From my point of view, I would say from a Java dev perspective, I prefer to have the related interface, type, factories, methods in order. Here there are many private types from l.83 to l.168 that should be moved elsewhere, next from where they are used. Indeed I think it would improve the reading. Yet I understand that it's go-ethereum code that we should not modify, only when needed.

In some project we used [quilt](https://linux.die.net/man/1/quilt), that is a layer of diffs than you apply. Maybe those internals improvement, we could handle with a tool like that, on top of our current process.

ricardolyn: I refactored the code as we discussed on the call.

core/types/receipt.go
```diff
@@ -74,6 +74,9 @@ type Receipt struct {
 	//
 	// This nested structure would not have more than 2 levels.
 	PSReceipts map[PrivateStateIdentifier]*Receipt `json:"-"`
+	// support saving the revert reason into the receipt itself for later consultation.
+	RevertReason string `json:"revertReason,omitempty"`
```
nmvalera: Core types (`Transaction`, `Receipt`, `Log`, etc.) do not use `string` fields but `[]byte` (e.g the transaction Data, Logs Data, etc.). 

`string` is mainly used when returning `0x` prefixed values to users.

 So I think `RevertReason` should be `[]byte` as well and should be transformed into hex `string` by the api when returning revert reason to user

core/types/receipt.go
```diff
@@ -508,7 +506,44 @@ func (r Receipts) deriveFieldsOrig(config *params.ChainConfig, hash common.Hash,
...
-//
+
+// storedMPSReceiptRLPWithRevertReason is the storage encoding of a receipt which contains
+// receipts per PSI, including Revert Reason
+type storedMPSReceiptRLPWithRevertReason struct {
+	PostStateOrStatus []byte
+	CumulativeGasUsed uint64
+	Logs              []*LogForStorage
+	PSReceipts        []storedPSIToReceiptMapEntryWithRevertReason
+	RevertReason      string
```
nmvalera: To align with other internal types `RevertReason` should be  `[]byte` not `string`

core/types/receipt.go
```diff
@@ -508,7 +506,44 @@ func (r Receipts) deriveFieldsOrig(config *params.ChainConfig, hash common.Hash,
...
+	Key   PrivateStateIdentifier
+	Value storedReceiptRLP
+}
+
+// storedReceiptRLPWithRevertReason is the storage encoding of a receipt from geth upstream, with added revert reason
+type storedReceiptRLPWithRevertReason struct {
+	PostStateOrStatus []byte
+	CumulativeGasUsed uint64
+	Logs              []*LogForStorage
+	RevertReason      string
```
nmvalera: Same `string` -> `[]byte`

internal/ethapi/api.go
```diff
@@ -1584,6 +1584,12 @@ func (s *PublicTransactionPoolAPI) GetTransactionReceipt(ctx context.Context, ha
 		"logsBloom":         receipt.Bloom,
 	}
 
+	// Quorum
+	if receipt.RevertReason != "" {
+		fields["revertReason"] = receipt.RevertReason
```
nmvalera: Once `RevertReason` has been set to `[]byte` the hex encoding should happen here

eth/backend.go
```diff
@@ -219,6 +219,10 @@ func New(stack *node.Node, config *Config) (*Ethereum, error) {
 		newBlockChainFunc = core.NewMultitenantBlockChain
 	}
 	eth.blockchain, err = newBlockChainFunc(chainDb, cacheConfig, chainConfig, eth.engine, vmConfig, eth.shouldPreserve, &config.TxLookupLimit)
+
+	// Quorum
+	eth.blockchain.SetSaveRevertReason(config.SaveRevertReason)
```
nmvalera, line 224: Have you considered the option of updating `core.NewBlockChain` and `core.NewMultitenantBlockChain` to take `config` as input? (setting configuration after creating object may be error-prone)

ricardolyn: yes, I have and agree that is error-prone. the issue is that we need to change 100+ files, touching in many places of geth (because of `core.NewBlockChain`). one approach is to create a new object we call Quorum config or something so we can in the future pass other configuration needed (to avoid changing again 100+ files) @nmvalera wdyt?

core/types/receipt.go
ricardolyn: changing and I do the transform when returning on ethapi

dolanor: Is there some kind of type already in geth codebase that abstracts this?
Like some kind of
```go
type EncodedData []byte

func (ed EncodedData) Bytes() []byte {
    return []byte(ed)
}

func (ed EncodedData) String() string {
    //return whatever func that transforms to the correct encoded string format
}
```

ricardolyn: in the code, they mostly use `hexutil.Encode(receipt.RevertReason)`

eth/backend.go
ricardolyn: created https://github.com/ConsenSysQuorum/quorum/issues/253

ricardolyn:
> Thanks for this feature @ricardolyn , it's really useful and more straight forward than executing `eth_call` again! On Besu, back when this was introduced, we had seen that there was a warning stating this feature increases Memory consumption significantly: https://besu.hyperledger.org/en/1.2.1/Using-Pantheon/Transactions/Revert-Reason/ . Are you aware of any similar considerations for Quorum?

It adds more memory and storage consumption, but taking into consideration other data saved, it may not increase significantly or an exponential factor. It adds the revert reason byte array to the receipt of every failed transaction (on storage and memory cache). We didn't do any profiling at the moment on the memory consumption but if we do, we will provide more information.

nmvalera (dismissed):
LGTM

nmvalera (approved):
LGTM

PR #1227, opened 2021-07-19
Separate QuorumReceiptExtraData from receipt encoding

nicolae-leonte-go:
Introduce the QuorumReceiptExtraData struct to hold receipt fields introduced by quorum. Revert changes to ReceiptForStorage which was used to store MPS and RevertReason data.

If a receipt array to be stored contains at least one receipt that has QuorumReceiptExtraData fields populated, an additional byte array is produced containing the RLP encoded QuorumReceiptExtraData structures. The additional byte array is then concatenated to the vanilla receipt array for storage in the underlying database.

A procedure for adding new fields to the QuorumReceiptExtraData array has also been documented.

**Backward/Forward compatibility** - if a node is updated to the latest binary and makes use of the new receipt features (a few blocks are stored with new receipt encoding) rolling back to the previous binary is not possible. The older binary would fail to load the new receipts format.


ricardolyn (dismissed):
👍 
I would recommend another approval from another reviewer to confirm

core/types/receipt.go
```diff
@@ -66,19 +67,40 @@ type Receipt struct {
 	BlockHash        common.Hash `json:"blockHash,omitempty"`
 	BlockNumber      *big.Int    `json:"blockNumber,omitempty"`
 	TransactionIndex uint        `json:"transactionIndex"`
+}
 
-	// Quorum
-	//
+// Quorum
+/*
+The QuorumReceiptExtraData contains additional fields to be stored for receipts introduced by quourm.
```
baptiste-b-pegasys:
```suggestion
The QuorumReceiptExtraData contains additional fields to be stored for receipts introduced by Quorum.
```

core/types/receipt.go
```diff
@@ -66,19 +67,40 @@ type Receipt struct {
...
+
+Procedure for adding new fields to QuorumReceiptExtraData:
+
+1. Add the relevant field to the QuorumReceiptExtraData structure
+
+2. Introduce a new  version for the structures: storedQuorumReceiptExtraDataVxyzRLP, storedPSIToReceiptMapEntryVxyz and storedReceiptExtraDataVxyz with the new field
+
+3. Update the QuorumReceiptExtraData.IsEmpty
+
+3. Update the QuorumReceiptExtraData.SetReceiptExtraDataFromStorage
```
baptiste-b-pegasys:
```suggestion
4. Update the QuorumReceiptExtraData.SetReceiptExtraDataFromStorage
```

core/types/receipt.go
```diff
@@ -66,19 +67,40 @@ type Receipt struct {
...
+
+4. Update the QuorumReceiptExtraData.EncodeRLP - ensure that you increment the version and that you assign the version when you instantiate the new structure (you may also need to introduce a new version for any of the conversion helper methods)
```
baptiste-b-pegasys:
```suggestion
5. Update the QuorumReceiptExtraData.EncodeRLP - ensure that you increment the version and that you assign the version when you instantiate the new structure (you may also need to introduce a new version for any of the conversion helper methods)
```

core/types/receipt.go
```diff
@@ -66,19 +67,40 @@ type Receipt struct {
...
+
+5. Create a new method decodeStoredQuorumReceiptExtraDataVxyz that handles the decoding of the new structure
```
baptiste-b-pegasys:
```suggestion
6. Create a new method decodeStoredQuorumReceiptExtraDataVxyz that handles the decoding of the new structure
```

core/types/receipt.go
```diff
@@ -66,19 +67,40 @@ type Receipt struct {
...
+
+6. Update the QuorumReceiptExtraData.DecodeRLP - update the `switch version` statement and invoke the decode method for the newly added version
```
baptiste-b-pegasys:
```suggestion
7. Update the QuorumReceiptExtraData.DecodeRLP - update the `switch version` statement and invoke the decode method for the newly added version
```

core/types/receipt.go
```diff
@@ -508,37 +453,33 @@ func (r Receipts) deriveFieldsOrig(config *params.ChainConfig, hash common.Hash,
...
-	PSReceipts        []storedPSIToReceiptMapEntryWithRevertReason
-	RevertReason      []byte
-}
-
-type storedPSIToReceiptMapEntryWithRevertReason struct {
-	Key   PrivateStateIdentifier
-	Value storedReceiptRLPWithRevertReason
+func (r *Receipt) SetReceiptExtraDataFromStorage(data *QuorumReceiptExtraData) {
+	if data == nil {
+		return
```
baptiste-b-pegasys, line 458: no need to reset the fields?

core/types/receipt.go
```diff
@@ -609,6 +553,77 @@ func convertLogsForEncoding(logs []*Log) []*LogForStorage {
 	return result
 }
 
+func decodeStoredQuorumReceiptExtraDataV1(r *QuorumReceiptExtraData, blob []byte) error {
```
baptiste-b-pegasys, line 556: Just `decodeQuorumReceiptExtraDataV1` or
```suggestion
func decodeAndStoreQuorumReceiptExtraDataV1(r *QuorumReceiptExtraData, blob []byte) error {
```
?

baptiste-b-pegasys (commented):
How frequent do we need to add new fields, so to create a new version?
Is the multiplication of versions would be a problem in future?
Can we reuse the structure of older version or do we need to copy/paste again?
What will be the behavior of a transaction if a node doesn't support a version of the receipt? (it's not commited?, do we know the node(s) that doesn't support the feature?)

Thank you!

core/types/receipt.go
nicolae-leonte-go: It is derived from the name of the structure `storedQuorumReceiptExtraDataV1RLP` with the `decode` prefix... I'd prefer to keep it that way. 
There is no `QuorumReceiptExtraDataVxyz` - we don't have versions for that. We only have versions for the stored structure.

core/types/receipt.go
nicolae-leonte-go: fixed

core/types/receipt.go
nicolae-leonte-go: fixed

core/types/receipt.go
nicolae-leonte-go: fixed

core/types/receipt.go
nicolae-leonte-go: fixed

core/types/receipt.go
nicolae-leonte-go: fixed

core/types/receipt.go
nicolae-leonte-go: renamed to `FillReceiptExtraDataFromStorage`

core/types/receipt.go
```diff
@@ -66,19 +67,40 @@ type Receipt struct {
 	BlockHash        common.Hash `json:"blockHash,omitempty"`
 	BlockNumber      *big.Int    `json:"blockNumber,omitempty"`
 	TransactionIndex uint        `json:"transactionIndex"`
+}
 
-	// Quorum
-	//
+// Quorum
+/*
+The QuorumReceiptExtraData contains additional fields to be stored for receipts introduced by Quourm.
```
baptiste-b-pegasys:
```suggestion
The QuorumReceiptExtraData contains additional fields to be stored for receipts introduced by Quorum.
```
last one :)

trung (dismissed):
LGTM

core/types/receipt.go
nicolae-leonte-go: fixed

trung (approved):
LGTM

baptiste-b-pegasys (approved):

PR #1228, opened 2021-07-21
Refactorisation: use a global Quorum Chain Config to avoid mutability

baptiste-b-pegasys:


ricardolyn (dismissed):
Looks good. to be merged after 21.7.0 release

eth/config.go
```diff
@@ -202,10 +202,7 @@ type Config struct {
 	EVMCallTimeOut time.Duration
 
 	// Quorum
-	EnableMultitenancy bool `toml:"-"`
-
-	// Quorum
-	SaveRevertReason bool `toml:"-"`
+	QuorumChainConfig core.QuorumChainConfig `toml:"-"`
```
trung: I think with
```
type Config struct {
    *core.QuorumChainConfig // or core.QuorumChainConfig
}
```
We would shorten the references to the config. E.g.: `b.eth.config.QuorumChainConfig.MultiTenantEnabled()` would now become `b.eth.config.MultiTenantEnabled()`

core/blockchain.go
```diff
@@ -221,20 +220,22 @@ type BlockChain struct {
...
 	privateStateManager mps.PrivateStateManager
-	saveRevertReason    bool // if we should save the revert reasons in the Tx Receipts
 	// End Quorum
 }
 
 // NewBlockChain returns a fully initialised block chain using information
 // available in the database. It initialises the default Ethereum Validator and
 // Processor.
-func NewBlockChain(db ethdb.Database, cacheConfig *CacheConfig, chainConfig *params.ChainConfig, engine consensus.Engine, vmConfig vm.Config, shouldPreserve func(block *types.Block) bool, txLookupLimit *uint64) (*BlockChain, error) {
+func NewBlockChain(db ethdb.Database, cacheConfig *CacheConfig, chainConfig *params.ChainConfig, engine consensus.Engine, vmConfig vm.Config, shouldPreserve func(block *types.Block) bool, txLookupLimit *uint64, quorumChainConfig *QuorumChainConfig) (*BlockChain, error) {
```
trung, line 232: Hope there's a way not to change the original method signature so to ease the upstream merge effort.

core/blockchain.go
```diff
@@ -244,23 +245,24 @@ func NewBlockChain(db ethdb.Database, cacheConfig *CacheConfig, chainConfig *par
...
+		bodyRLPCache:   bodyRLPCache,
+		receiptsCache:  receiptsCache,
+		blockCache:     blockCache,
+		txLookupCache:  txLookupCache,
+		futureBlocks:   futureBlocks,
+		engine:         engine,
+		vmConfig:       vmConfig,
+		badBlocks:      badBlocks,
+		// Quorum
+		quorumConfig: *quorumChainConfig,
```
trung: How about passing by value to the `NewBlockChain()` so we can remove the pointer dereference here.
Or we use passing by reference instead so in the future any non-primitive type in QuorumChainConfig would be taken care automatically.

trung (commented):
Few nitpicks, otherwise it looks good to me.

core/blockchain.go
baptiste-b-pegasys: I did a variadic variable but got reviewed earlier
https://github.com/ConsenSysQuorum/quorum/pull/261#discussion_r672906158

ricardolyn: the goal was to in the future only have this config struct for any necessary quorum config, requiring less changes in future upstream merges.

ricardolyn: but without making it unclear with the variadic variable. @trung what do you think?

trung: LGTM

ricardolyn: keep as it is or changing to variadic variable?

ricardolyn (dismissed):
LGTM

ricardolyn (dismissed):
LGTM

ricardolyn (approved):

PR #1229, opened 2021-07-21
Check if sb.core is nil before processing NewBlockMsg

jbhurat:
A non-mining, non-validator node does not have istanbul backend core set. When a `p2p` message of code corresponding `NewBlockMsg` is received it checks if that node is a proposer. As backend core is nil, this results in a panic. This PR adds a check to see if core is nil and only processes the message is it is not nil, which should be the case for a mining node

### DIFF
--- a/consensus/istanbul/backend/handler.go
+++ b/consensus/istanbul/backend/handler.go
@@ -103,7 +103,7 @@ func (sb *Backend) HandleMsg(addr common.Address, msg p2p.Msg) (bool, error) {
 	}
 	//https://github.com/ConsenSys/quorum/pull/539
 	//https://github.com/ConsenSys/quorum/issues/389
-	if msg.Code == NewBlockMsg && sb.core.IsProposer() { // eth.NewBlockMsg: import cycle
+	if msg.Code == NewBlockMsg && sb.core != nil && sb.core.IsProposer() { // eth.NewBlockMsg: import cycle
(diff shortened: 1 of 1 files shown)


nmvalera (approved):
LGTM

consensus/istanbul/backend/handler.go
```diff
@@ -103,7 +103,7 @@ func (sb *Backend) HandleMsg(addr common.Address, msg p2p.Msg) (bool, error) {
 	}
 	//https://github.com/ConsenSys/quorum/pull/539
 	//https://github.com/ConsenSys/quorum/issues/389
-	if msg.Code == NewBlockMsg && sb.core.IsProposer() { // eth.NewBlockMsg: import cycle
+	if msg.Code == NewBlockMsg && sb.core != nil && sb.core.IsProposer() { // eth.NewBlockMsg: import cycle
```
ricardolyn, line 106: is there a unit test that can capture this?

nmvalera: Basically `sb.core` is nil for non-validator node.

So if we have ATs with non-validator nodes (with the flag `--mine` off) then this is well covered.

ricardolyn: it was only a way to capture faster, as AT take more time. 

non-blocking for me, so we can go ahead and merge.

ricardolyn (approved):

PR #1230, opened 2021-07-28
feature: more acceptance tests executed, and also on the master branch

baptiste-b-pegasys:


.github/workflows/pr.yml
```diff
@@ -87,6 +95,12 @@ jobs:
         tag:
           - 'basic || basic-raft || (advanced && raft) || networks/typical::raft'
           - 'basic || basic-istanbul || (advanced && istanbul && !block-heights) || networks/typical::istanbul'
+          - 'validator-management && networks/template::istanbul-3plus1'
```
nmvalera, line 91: This looks good for PRs.

For `master` we should be running all tests, I think we should add it to `build.yml` so binaries and docker image are not published in case test fail

ricardolyn: same as https://github.com/ConsenSys/quorum-acceptance-tests/blob/master/.github/workflows/run.yml#L97

b-2-83: solved

nmvalera (dismissed):
LGTM

.github/workflows/build.yml
```diff
@@ -49,10 +49,110 @@ jobs:
...
+          - 'privacy-enhancements-upgrade || networks/template::istanbul-4nodes-pe'
+          - 'multitenancy && networks/plugins::raft-multitenancy'
+          - 'basic || basic-raft || (advanced && raft) || networks/typical::raft-simple-mps'
+          - 'basic || basic-istanbul || (advanced && istanbul) || networks/typical::istanbul-simple-mps'
+          - 'basic || networks/typical::raftmps'
+          - 'basic || networks/typical::istanbulmps'
+          - 'mps-upgrade-txtrace || networks/template::raft-4nodes-mps'
+          - 'mps-upgrade-txtrace || networks/template::istanbul-4nodes-mps'
+          - '(basic && !nosupport && !mps && !(spam && !raw) && !eth-api-signed && !privacy-enhancements-disabled && !graphql && !async && !extension && !storage-root && !personal-api-signed) || networks/typical-besu::ibft2'
+        privacy-enhancements:
```
nmvalera, line 122: Open Q: how is this section used exactly? Thx

baptiste-b-pegasys: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-additional-values-into-combinations

you can read as all above is list of tuples of a tag and privacy-enhancements @ false, then we include two specific tuples

nmvalera (approved):
LGTM

PR #1231, opened 2021-07-29
Privacy Precompile: enables private transactions with extended obfuscated fields

SatpalSandhu61:
## Introduction

Add support for privacy precompiled contract to orchestrate the execution of private transactions. This new approach allows extending the obfuscated fields of a private transaction. 

So far `data` field was the only obfuscated field of a private transaction, other fields were visible from non-participants node. This new precompiled contract approach allows to obfuscate `data`, `to`, and `gasLimit`. In some future enhancement, we could use this same approach to obfuscate the `sender` address.

This introduces the concept of **Privacy Marker Transactions (PMT)**, which is a standard public transaction sent to the privacy precompiled contract to get private transactions executed
- `to` field is always be set to privacy precompiled at address `0x000000000000000000000000000000000000007a`
- `data` field is set to the Tessera key of the extended private transaction

When processing a PMT the privacy precompile of participant nodes executes the extended private transaction corresponding to the Tessera key in PMT `data`. The final execution of the private transaction is identical to the usual private transaction in particular it is the private state of participants that is mutated.

## New JSON-RPC methods

- `eth_distributePrivateTransaction` enables to push a private transaction payload to Tessera
- `eth_getPrivacyPrecompileAddress` that returns the precompile address (so far this is fixed to `0x000000000000000000000000000000000000007a`)

## Configuration

1. Add genesis field `"privacyPrecompileBlock": 0,` under `config` section which defines block number, after which privacy precompiled contract is enabled
2. Pass CLI `--privacymarker.enable` flag when starting node

#### Node signing

1. User sends the transaction to GoQuorum using `eth_sendTransaction` as normal in particular passing `privateFor` and `to` field set to the private contract address (or nil for contract deployment)
1. GoQuorum prepare and signs the private transaction containing all fields
1. GoQuorum sends signed private transaction to Tessera, which distributes the private transaction to other participants and returns the private hash
1. GoQuorum creates the ‘Privacy Marker Transaction’ (PMT), where:
    - `data` is the private hash of the transaction stored in Tessera.
    - `to` is the privacy precompile address `0x000000000000000000000000000000000000007a`
    - PMT is signed using the same account as the one used for the private transaction
1. GoQuorum submits the PMT to the tx pool as any standard public transaction

#### External signing

1. User creates and externally sign private tx as normal (with signature v modified value) particular setting `privateFor` and `to` field set to the private contract address (or nil for contract deployment)
1. User calls `eth_distributePrivateTransaction` with the externally signed private transaction. On this call, GoQuorum pushes the private transaction payload to Tessera that encrypts it and returns private tx manager hash
1. User creates and externally signs the PMT
   - `data` should be the private tx manager hash got on previous call to `eth_distributePrivateTransaction`
   - `to` should be the privacy precompile address `0x000000000000000000000000000000000000007a`
1. User sends the signed PMT with `eth_sendRawTransaction`

### Transaction Processing

When executing the PMT

1. Transaction is dispatch to the privacy precompiled contract for execution
2. Privacy precompile reads private transaction hash from PMT `data` field as private hash
3. Privacy precompile queries Tessera with private hash, if no data is found, then this node is not a participant so no further processing is needed
3. Privacy precompile triggers the EVM execution of the private transaction using the private state as per normal private transaction


internal/ethapi/api.go
```diff
@@ -2541,72 +2765,126 @@ func handlePrivateTransaction(ctx context.Context, b Backend, tx *types.Transact
...
+	if err != nil {
+		return nil, err
+	}
+
+	_, _, ptmHash, err := private.P.Send(data.Bytes(), privateTxArgs.PrivateFrom, privateTxArgs.PrivateFor, &engine.ExtraMetadata{})
+	if err != nil {
+		return nil, err
+	}
+
+	pmt := types.NewTransaction(privateTx.Nonce(), common.QuorumPrivacyPrecompileContractAddress(), privateTx.Value(), privateTx.Gas(), privateTx.GasPrice(), ptmHash.Bytes())
```
nmvalera, line 2882: @SatpalSandhu61 When we discussed we said that PMT gas limit should be equal to the `intrinsicGas`. From reading this it seems that the PMT gas limit is actually set to the private tx gas limit. Am I understanding correctly? If yes can we update?

A similar question applies to `GasPrice` 

Thanks a lot

SatpalSandhu61: I think there was a misunderstanding there. What I was trying to say is that since gas is free on GoQuorum, and we never show how much gas a private transaction actually used, users can set an arbitrarily high gasLimit on the private transaction to avoid revealing any information.

ricardolyn (dismissed):

nmvalera (approved):
LGTM

Krish1979 (approved):

PR #1232, opened 2021-08-03
Asynchronously call startNewRound() when transitioning from IBFT to QBFT

jbhurat:
When transitioning from Legacy IBFT to QBFT consensus `startNewRound()` needs to be called asynchronously, so that `stop()` on `core` can successfully unsubscribe from events. During normal IBFT consensus flow, `startNewRound()` should be called synchronously which was the case prior to QBFT changes

### DIFF
--- a/consensus/istanbul/ibft/core/final_committed.go
+++ b/consensus/istanbul/ibft/core/final_committed.go
@@ -16,11 +16,23 @@
 
 package core
 
-import "github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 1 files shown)


nmvalera (approved):
LGTM.

As a general comment, I think the switch from IBFT to QBFT is too complex with this type of unexpected asynchronous effects which are error prone and hard to debug. We should aim at solving this as part of future work on bft migrations.

PR #1234, opened 2021-08-11
Privacy Precompile: enables private transactions with extended obfuscated fields

SatpalSandhu61:
## Introduction

Add support for privacy precompiled contract to orchestrate the execution of private transactions. This new approach allows extending the obfuscated fields of a private transaction. 

So far `data` field was the only obfuscated field of a private transaction, other fields were visible from non-participants node. This new precompiled contract approach allows to obfuscate `data`, `to`, and `gasLimit`. In some future enhancement, we could use this same approach to obfuscate the `sender` address.

This introduces the concept of **Privacy Marker Transactions (PMT)**, which is a standard public transaction sent to the privacy precompiled contract to get private transactions executed
- `to` field is always be set to privacy precompiled at address `0x000000000000000000000000000000000000007a`
- `data` field is set to the Tessera key of the extended private transaction

When processing a PMT the privacy precompile of participant nodes executes the extended private transaction corresponding to the Tessera key in PMT `data`. The final execution of the private transaction is identical to the usual private transaction in particular it is the private state of participants that is mutated.

## New JSON-RPC methods

- `eth_distributePrivateTransaction` enables to push a private transaction payload to Tessera
- `eth_getPrivacyPrecompileAddress` that returns the precompile address (so far this is fixed to `0x000000000000000000000000000000000000007a`)

## Configuration

1. Add genesis field `"privacyPrecompileBlock": 0,` under `config` section which defines block number, after which privacy precompiled contract is enabled
2. Pass CLI `--privacymarker.enable` flag when starting node

#### Node signing

1. User sends the transaction to GoQuorum using `eth_sendTransaction` as normal in particular passing `privateFor` and `to` field set to the private contract address (or nil for contract deployment)
1. GoQuorum prepare and signs the private transaction containing all fields
1. GoQuorum sends signed private transaction to Tessera, which distributes the private transaction to other participants and returns the private hash
1. GoQuorum creates the ‘Privacy Marker Transaction’ (PMT), where:
    - `data` is the private hash of the transaction stored in Tessera.
    - `to` is the privacy precompile address `0x000000000000000000000000000000000000007a`
    - PMT is signed using the same account as the one used for the private transaction
1. GoQuorum submits the PMT to the tx pool as any standard public transaction

#### External signing

1. User creates and externally sign private tx as normal (with signature v modified value) particular setting `privateFor` and `to` field set to the private contract address (or nil for contract deployment)
1. User calls `eth_distributePrivateTransaction` with the externally signed private transaction. On this call, GoQuorum pushes the private transaction payload to Tessera that encrypts it and returns private tx manager hash
1. User creates and externally signs the PMT
   - `data` should be the private tx manager hash got on previous call to `eth_distributePrivateTransaction`
   - `to` should be the privacy precompile address `0x000000000000000000000000000000000000007a`
1. User sends the signed PMT with `eth_sendRawTransaction`

### Transaction Processing

When executing the PMT

1. Transaction is dispatch to the privacy precompiled contract for execution
2. Privacy precompile reads private transaction hash from PMT `data` field as private hash
3. Privacy precompile queries Tessera with private hash, if no data is found, then this node is not a participant so no further processing is needed
3. Privacy precompile triggers the EVM execution of the private transaction using the private state as per normal private transaction


accounts/abi/bind/bind_quorum_test.go
```diff
@@ -0,0 +1,231 @@
...
+	require.True(t, pvtTx.IsPrivate())
+
+	require.Equal(t, []string{"tm1"}, pvtTxArgs.PrivateFor)
+}
+
+func passthroughSigner(_ types.Signer, _ common.Address, tx *types.Transaction) (*types.Transaction, error) {
+	return tx, nil
+}
+
+type mockTransactor struct {
```
baptiste-b-pegasys, line 198: can we generate that mock with an interface?

core/types/transaction.go
```diff
@@ -581,6 +581,10 @@ func (tx *Transaction) SetPrivate() {
 	}
 }
 
+func (tx *Transaction) IsPrivacyMarker() bool {
```
baptiste-b-pegasys, line 584:
```suggestion
func (tx *Transaction) HasPrivacyMarker() bool {
```
many "is privacy marker" changes, I prefer with has, or use

accounts/abi/bind/bind_quorum_test.go
baptiste-b-pegasys, line 198: can it be generated.

signer/core/api.go
```diff
@@ -533,7 +533,7 @@ func (api *SignerAPI) SignTransaction(ctx context.Context, args SendTxArgs, meth
 		result SignTxResponse
 		msgs   *ValidationMessages
 	)
-	if args.IsPrivate {
+	if args.IsPrivate || args.isPrivacyMarker() {
```
baptiste-b-pegasys, line 536:
```suggestion
	if args.IsPrivate || args.hasPrivacyMarker() {
```

baptiste-b-pegasys (commented):
it may have an issue with `mps` ATs, checking that locally.
```
[ {
  "file" : "src/specs/01_basic/extend_private_contract.spec",
  "line" : 1,
  "col" : 1,
  "message" : "Scenario: Extend contract to a new party, creator cancels extension 1%0AStep: <fromNode> cancels <contractName>%0AConcept Execution Failure: 'org.opentest4j.AssertionFailedError: %0AExpecting:%0A <org.web3j.protocol.core.Response$Error@237a1875>%0Ato be equal to:%0A <null>%0Abut was not.'"
} ]
```

extension/api.go
```diff
@@ -313,6 +313,7 @@ func (api *PrivateExtensionAPI) ExtendContract(ctx context.Context, toExtend com
 		return "", err
 	}
 
+	//TODO(peter): revisit this, give back private tx hash or marker hash?
```
ricardolyn: anything to follow up on this?

internal/ethapi/api.go
```diff
@@ -1930,6 +2089,28 @@ func (s *PublicTransactionPoolAPI) SendTransaction(ctx context.Context, args Sen
 	if err != nil {
 		return common.Hash{}, err
 	}
+
+	// Quorum
+	if signed.IsPrivate() && s.b.IsPrivacyMarkerTransactionCreationEnabled() {
+		pmt, err := createPrivacyMarkerTransaction(s.b, signed, &args.PrivateTxArgs)
```
ricardolyn, line 2090: this block seems to be quite similar to the block on https://github.com/ConsenSys/quorum/pull/1234/files#diff-c426ecd2f7d247753b9ea8c1cc003f21fa412661c1f967d203d4edf8163da344R499, could we remove duplication?

accounts/abi/bind/base.go
```diff
@@ -424,6 +430,25 @@ func (c *BoundContract) createPrivateTransaction(tx *types.Transaction, payload
...
+	if err != nil {
+		return nil, err
+	}
+
+	hash, err := c.transactor.DistributeTransaction(ensureContext(opts.Context), signedTx, args)
+	if err != nil {
+		return nil, err
+	}
+
+	return types.NewTransaction(signedTx.Nonce(), common.QuorumPrivacyPrecompileContractAddress(), tx.Value(), tx.Gas(), tx.GasPrice(), common.FromHex(hash)), nil
```
ricardolyn: shouldn't be the gas limit here the same as https://github.com/ConsenSys/quorum/pull/1234/files#diff-c426ecd2f7d247753b9ea8c1cc003f21fa412661c1f967d203d4edf8163da344R2884?

accounts/abi/bind/bind_quorum_test.go
SatpalSandhu61: This hasn't changed since the PR was originally reviewed. We don't want to make changes at this late stage.

core/types/transaction.go
SatpalSandhu61: This hasn't changed since the PR was originally reviewed. We don't want to make further changes at this late stage now.

chris-j-h:
> it may have an issue with `mps` ATs, checking that locally.
> 
> ```
> [ {
>   "file" : "src/specs/01_basic/extend_private_contract.spec",
>   "line" : 1,
>   "col" : 1,
>   "message" : "Scenario: Extend contract to a new party, creator cancels extension 1%0AStep: <fromNode> cancels <contractName>%0AConcept Execution Failure: 'org.opentest4j.AssertionFailedError: %0AExpecting:%0A <org.web3j.protocol.core.Response$Error@237a1875>%0Ato be equal to:%0A <null>%0Abut was not.'"
> } ]
> ```

@baptiste-b-pegasys this PR requires a change to the atests - I am in the process of preparing that PR

signer/core/api.go
SatpalSandhu61: This hasn't changed since the PR was originally reviewed, so we'd prefer not to make changes now.

extension/api.go
SatpalSandhu61: No, I've removed the stale comment.

internal/ethapi/api.go
SatpalSandhu61: That link doesn't work for me - it just takes me to the head of the PR.
However, this code hasn't changed from the reviewed PR. I'd be reluctant to do any refactoring now as I believe you want to get the PR merged as soon as possible.

ricardolyn: `internal/ethapi/api.go` line 488

accounts/abi/bind/base.go
SatpalSandhu61: Good point, ideally it would also use the Intrinsic Gas value here too. I'll make that change.
It'll take a day or so, as I'll need to run acceptance tests against it to ensure it doesn't cause nay other issues. If necessary, it can be added after the PR is merged (to avoid holding up other merges to master).

ricardolyn: no need to add after, as we need to have this ready for release so lets put all in one PR that we can merge and then ready for release 👍 (other PRs won't go in in the same release)

SatpalSandhu61: Ok, I've just committed a change to use IntrinsicGas for the abi generated PMT.

accounts/abi/bind/base.go
```diff
@@ -424,6 +431,31 @@ func (c *BoundContract) createPrivateTransaction(tx *types.Transaction, payload
...
+		return nil, err
+	}
+
+	// Note: using isHomestead and isEIP2028 set to true, which may give a slightly higher gas value (but avoids making an API call to get the block number)
+	intrinsicGas, err := core.IntrinsicGas(common.Hex2Bytes(common.MaxPrivateIntrinsicDataHex), false, true, true)
+	if err != nil {
+		return nil, err
+	}
+
+	return types.NewTransaction(signedTx.Nonce(), common.QuorumPrivacyPrecompileContractAddress(), tx.Value(), intrinsicGas, tx.GasPrice(), common.FromHex(hash)), nil
```
ricardolyn, line 456: @SatpalSandhu61 does the hash has exactly 32 bytes?

SatpalSandhu61: It's 64 bytes.

core/state_processor.go
```diff
@@ -210,8 +210,7 @@ func handleMPS(ti int, tx *types.Transaction, gp *GasPool, usedGas *uint64, cfg
 func ApplyTransactionOnMPS(config *params.ChainConfig, bc *BlockChain, author *common.Address, originalGP *GasPool,
 	publicStateDBFactory func() *state.StateDB, privateStateDBFactory func(psi types.PrivateStateIdentifier) (*state.StateDB, error),
 	header *types.Header, tx *types.Transaction, usedGas *uint64, cfg vm.Config, privateStateRepo mps.PrivateStateRepository, applyOnPartiesOnly bool) (*types.Receipt, error) {
-	// clone the gas pool (as we don't want to keep consuming intrinsic gas multiple times for each MPS execution)
-	gp := new(GasPool).AddGas(originalGP.Gas())
```
ricardolyn, line 172: non-blocking: why before wasn’t needed a new gas pool for each PSI on MPS while now is needed? as PMT shouldn't impact this

accounts/abi/bind/base.go
ricardolyn: non-blocking: Besu uses 32 bytes I believe. as we are not doing the same implementation as Besu and we use a different pre-compile address, in the future we can handle a common approach. however, I ask if is useful to follow now the same size?

core/state_processor.go
SatpalSandhu61: Whilst investigating an issue with the intrinsic gas change, I realised this was an issue on MPS. The cloned gaspool is needed for each PSI, otherwise the execution of the txn on each PSI will decrement the gaspool, potentially leading to `Error: gas limit reached` on the MPS node.

accounts/abi/bind/base.go
SatpalSandhu61: Tessera returns a 64 byte hash, so we have to use that. If we wanted to use a 32 byte hash (either in just the PMT, or in general) then it would require major changes.

core/state_processor.go
ricardolyn: ok. thanks. I need to check this also for the private pre fetching as I might have the same issue.

accounts/abi/bind/base.go
ricardolyn: ok, thanks. understood

ricardolyn (approved):
LGTM

Krish1979 (approved):

accounts/abi/bind/base.go
nmvalera: @SatpalSandhu61 We could have computed the intrinsic gas of the actual PMT payload `common.FromHex(hash)` so we do not need to make any assumption on the format of the enclave key. Was there a reason for taking the intrinsic gas of `MaxPrivateIntrinsicDataHex` instead? Thanks.

SatpalSandhu61: @nmvalera Good point. It should really use the payload hash rather than `MaxPrivateIntrinsicDataHex`. I'll create a PR to amend that.

PR #1236, opened 2021-08-17
Amend IntrinsicGas calculation for Privacy Marker Transaction to use actual payload hash, instead of hardcoded value

SatpalSandhu61:
Minor amendment to the merged privacy precompile changes.


chris-j-h:
We can also undo the move of `MaxPrivateIntrinsicDataHex` to the `common` pkg to reduce the overall diff of the PMT changes

ricardolyn (approved):

PR #1237, opened 2021-08-18
change: define specific metrics per trie db (priv)

baptiste-b-pegasys:


trie/database.go
```diff
@@ -30,34 +30,9 @@ import (
 	"github.com/ethereum/go-ethereum/core/rawdb"
 	"github.com/ethereum/go-ethereum/ethdb"
 	"github.com/ethereum/go-ethereum/log"
-	"github.com/ethereum/go-ethereum/metrics"
 	"github.com/ethereum/go-ethereum/rlp"
 )
 
-var (
```
ricardolyn, line 37: this is mainly geth code. how can we make it easy to handle this on a geth upgrade?

baptiste-b-pegasys: I know this is why I isolated on a specific PR for discussion.

The only possibility I see is to keep one metric, but add some tag in it on each usage. (if feasible)

The way it's done is not valid for me because we use one metric for multiple instances, where geth has only one.
Pushing this PR to geth would certainly be rejected as there is only one trie db.

I can leave it as a comment block, so it will not block merge if one metric is added/modified/removed, then the conflicts/errors will be only on the usages.

PR #1238, opened 2021-08-18
Prefetch next block to warm up Private State 

ricardolyn:
### Summary

The goal of the PR is warm up the state cache when we are inserting blocks in batch into the system (like on a full sync). This is already applied on Public state but we are introducing here the same behaviour for dual state and MPS.

### Changes

* Modify the prefetched to handle private state changes
* Add support for MPS and PMT transactions
* Add unit tests to validate that the prefetching is updating the throwaway state (public, dual state and MPS, including PMT for those).
* Update some of the methods to generate chain data used in the tests to support private tx.


core/blockchain.go
```diff
@@ -2066,25 +2065,33 @@ func (bc *BlockChain) insertChain(chain types.Blocks, verifySeals bool) (int, er
...
+						bc.prefetcher.Prefetch(followup, throwaway, throwawayPrivateStateRepo, bc.vmConfig, &followupInterrupt)
+
+						blockPrefetchExecuteTimer.Update(time.Since(start))
+						if atomic.LoadUint32(interrupt) == 1 {
+							blockPrefetchInterruptMeter.Mark(1)
+						}
+					}(time.Now(), followup, throwaway, throwawayPrivateStateRepo, &followupInterrupt)
+				} else if stateRepoErr != nil {
+					log.Warn("Unable to load the private state repository for pre-fetching", "stateRepoErr", stateRepoErr)
+				}
```
baptiste-b-pegasys: I would put the standard go lang error handling, I think there is no case where err and private state repo are not nil same time.

if err != nil {
  log.Warn("Unable to load the private state repository for pre-fetching", "stateRepoErr", stateRepoErr)
} else {
...
}
but here is small change to handle the error with minimal change:
```suggestion
				if stateRepoErr == nil && privateStateRepo != nil {
					throwawayPrivateStateRepo := privateStateRepo.Copy()

					// Quorum: add privateStateThrowaway argument
					go func(start time.Time, followup *types.Block, throwaway *state.StateDB, privateStateThrowaway mps.PrivateStateRepository, interrupt *uint32) {
						bc.prefetcher.Prefetch(followup, throwaway, throwawayPrivateStateRepo, bc.vmConfig, &followupInterrupt)

						blockPrefetchExecuteTimer.Update(time.Since(start))
						if atomic.LoadUint32(interrupt) == 1 {
							blockPrefetchInterruptMeter.Mark(1)
						}
					}(time.Now(), followup, throwaway, throwawayPrivateStateRepo, &followupInterrupt)
				} else {
					log.Warn("Unable to load the private state repository for pre-fetching", "stateRepoErr", stateRepoErr)
				}
```

baptiste-b-pegasys (approved):

PR #1240, opened 2021-08-20
Standalone node MPS DB Upgrade

nicolae-leonte-go:
### Introduction
Since the release of the multiple private states feature the only way to switch from a standalone node to an MPS enabled configuration was to do a full node resync. This feature introduces a new `geth` command that allows a more expedient upgrade.

### MPS DB Upgrade Command

MPS can be enabled on a standalone node by executing the `mpsdbupgrade` command. It should be significantly faster than a normal sync from block 0 (especially if the network has reached a significant number of blocks).
The effect of the `mpsdbupgrade` command is to restructure/upgrade the existing database to present itself as an MPS enabled database with a single private state (and to set the chain `config.isMPS` to `true`).
Once the upgrade completes additional private states can be introduced but only from the current block height (you can't introduce private states at historic block heights).

**Command parameters:**

- `--datadir` - the directory containing the GoQuorum node database

**Example:**

```shell
$ geth mpsdbupgrade --datadir data
Processing block 1 with hash 0xf668e8b8320040a0cabd1b6ec963a79a08536409d82b9ccaa31f62b0a1a4dc10
...
Processing block 1232739 with hash 0x2423b5d0f4c2883172f657f08d7d359b81f144b8cb8393ee24ef285058d55ce8
MPS DB upgrade finished successfully.
```

### Related PRs
Docs: https://github.com/ConsenSys/doc.goquorum/pull/109
QAT: https://github.com/ConsenSys/quorum-acceptance-tests/pull/134


cmd/geth/chaincmd.go
```diff
@@ -400,6 +417,25 @@ func importChain(ctx *cli.Context) error {
 	return importErr
 }
 
+func mpsdbUpgrade(ctx *cli.Context) error {
+	stack, _ := makeConfigNode(ctx)
+	defer stack.Close()
+
+	// initialise the tx manager with the dummy tx mgr
+	private.P = &notinuse.DBUpgradePrivateTransactionManager{}
```
baptiste-b-pegasys, line 425: Why the `notinuse` package?

core/mps/multiple_psr.go
```diff
@@ -34,25 +36,43 @@ func NewMultiplePrivateStateRepository(db ethdb.Database, cache state.Database,
...
+	copy := &managedState{
 		stateDb:    ms.stateDb.Copy(),
 		stateCache: ms.stateCache,
 	}
+	copy.stateRootProviderFunc = copy.calPrivateStateRoot
+	return copy
+}
+
+// calPrivateStateRoot is to return state root hash from the commit of a managedState identified by psi
+func (ms *managedState) calPrivateStateRoot(isEIP158 bool) (common.Hash, error) {
```
baptiste-b-pegasys, line 66: What does mean `cal`?

core/mps/upgrade_db.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	emptyState, err := mpsRepo.DefaultState()
+	if err != nil {
+		return err
+	}
+	// pre-populate with dummy one as the state root is derived from block root hash
+	privateState := &managedState{}
+	mpsRepo.managedStates[types.DefaultPrivateStateIdentifier] = privateState
+	for idx := 1; idx <= int(currentBlockNumber); idx++ {
+		header := chain.GetHeaderByNumber(uint64(idx))
```
baptiste-b-pegasys: possible?
```suggestion
	for idx := uint64(1); idx <= currentBlockNumber; idx++ {
		header := chain.GetHeaderByNumber(idx)
```

core/mps/multiple_psr.go
nicolae-leonte-go: short for calculate

cmd/geth/chaincmd.go
nicolae-leonte-go: The `DBUpgradePrivateTransactionManager` extends the `PrivateTransactionManager` from the `notinuse` package and only overrides very few methods. When reviewing with Trung we decided that it was not worth putting it in it's own package.

Krish1979 (approved):

PR #1242, opened 2021-08-24
Protect ProposerPolicy Registry using Mutex

jbhurat:
`RegisterValidatorSet()` might throw a nil pointer, if the `validatorSet` registry is cleared in `ClearRegistry()` by another goroutine. To fix this access to `registry` needs to be protected.

### DIFF
--- a/consensus/istanbul/backend/engine_test.go
+++ b/consensus/istanbul/backend/engine_test.go
@@ -169,7 +169,6 @@ func TestSealStopChannel(t *testing.T) {
 		stop <- struct{}{}
 		eventSub.Unsubscribe()
 	}
-	go eventLoop()
 	resultCh := make(chan *types.Block, 10)
 	go func() {
 		err := engine.Seal(chain, block, resultCh, stop)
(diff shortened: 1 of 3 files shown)


consensus/istanbul/backend/engine_test.go
```diff
@@ -169,14 +169,14 @@ func TestSealStopChannel(t *testing.T) {
...
 	}
-	go eventLoop()
 	resultCh := make(chan *types.Block, 10)
 	go func() {
 		err := engine.Seal(chain, block, resultCh, stop)
 		if err != nil {
 			t.Errorf("error mismatch: have %v, want nil", err)
 		}
 	}()
+	go eventLoop()
```
ricardolyn, line 179: this change seems unrelated to title. what was it about?

ricardolyn (approved):
LGTM

consensus/istanbul/backend/engine_test.go
jbhurat: Yes, the change is unrelated to the PR. When I created the PR, this test failed. On looking closely, the test was failing like 1 in 250 - 300 runs. Moving the call to `eventLoop()` ensures that `stop` channel is not triggered before `Seal()` is called

PR #1243, opened 2021-08-25
Support mandatory recipients

namtruong:
The change is to allow users to add a `mandatoryFor` parameter when sending a transaction to Quorum. This feature is useful for use cases where there are requirements for regulatory/observing entities that will have access to the full state of a contract.

#### Changes

- A new privacy flag called `MandatoryRecipients(2)`. Transactions sent with this enhanced privacy mode will inherit the validations of `PartyProtection` but not `StateValidation`. It allows state divergence.
- A transaction sent with this privacy flag will need to include `mandatoryFor : []` - which are private TM public keys - otherwise it will be rejected
- Members that are part of `mandatoryFor` will need to be included in the `privateFor` otherwise transaction will be rejected.
- `/send` and `/sendsignedtx` requests to private TM will include mandatory recipients data 
- `mandatoryFor` members will also be checked against the contracts that the transaction impact. This is achieved by evm simulation that returns a list of ACOTHs and validations performed in Tessera
- Corresponding Tessera updates https://github.com/ConsenSys/tessera/pull/1325


nicolae-leonte-go (approved):
LGTM

chris-j-h (approved):

nmvalera (approved):
LGTM

Krish1979 (approved):
LGTM

ricardolyn (approved):
+1

PR #1247, opened 2021-09-02
fix: flacky unit test

baptiste-b-pegasys:
1. wait group add should be outside the go routine
2. it's not possible to have a event that the message has just been handled, so the test may happen too soon.

### DIFF
--- a/consensus/istanbul/backend/handler_test.go
+++ b/consensus/istanbul/backend/handler_test.go
@@ -21,6 +21,7 @@ import (
 	"io/ioutil"
 	"math/big"
 	"testing"
+	"time"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/consensus/istanbul"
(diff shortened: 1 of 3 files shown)


consensus/istanbul/backend/handler_test.go
```diff
@@ -84,8 +85,15 @@ func TestHandleNewBlockMessage_whenTypical(t *testing.T) {
 	arbitraryBlock, arbitraryP2PMessage := buildArbitraryP2PNewBlockMessage(t, false)
 	postAndWait(backend, arbitraryBlock, t)
 
-	handled, err := backend.HandleMsg(arbitraryAddress, arbitraryP2PMessage)
-
+	var handled bool
+	var err error
+	for i := 0; i < 5; i++ { // make 5 tries if a little wait
+		handled, err = backend.HandleMsg(arbitraryAddress, arbitraryP2PMessage)
```
ricardolyn, line 91: could we create a utility function in the same file to avoid the duplication across 3 tests?

ricardolyn (changes requested):

PR #1252, opened 2021-09-10
Contract extension with mandatory recipients

namtruong:
- Implement `/transaction/{hash}/mandatory`call to Tessera to retrieve mandatory recipients data to be included in contract extension deployment
- Update rpc call `eth.getContractPrivacyMetadata` to include mandatory recipients data in the json response - if transaction privacy type is `MandatoryRecipients`
- Note that contract extension will preserve the mandatory recipients list and is not designed to add a new mandatory recipient


nicolae-leonte-go (approved):
LGTM

ricardolyn (approved):

PR #1253, opened 2021-09-13
Fix private receipt of privacy marker transaction

namtruong:
The check on whether the transaction is a privacy marker transaction (PMT) so that the receipt for the inner private transaction need to be retrieved - was moved inside the `if tx.IsPrivate() {...}` check block (possibly during the geth upgrade) - which resulted in all receipts for PMTs being `null` as with PMT being treated as public transaction this block of code will never be reached.

This bug should be revealed by acceptance tests running with `privacy-precompile=true`


Krish1979 (approved):

ricardolyn (approved):
LGTM. @baptiste-b-pegasys for a final review

baptiste-b-pegasys (approved):

PR #1261, opened 2021-09-30
Send transactions to all peers instead of a sub-set

ricardolyn:
### Summary

The introduction of [EIP-2464](https://github.com/ethereum/EIPs/issues/2465) (`eth65` version) reduces the bandwidth used for transaction propagation from linear complexity in the number of peers to logarithmic one. While this approach removes the naive transfer of full transactions, it does create transaction pool consistency issues where some nodes are unable to promote transactions properly to be included in a block and thus creating empty blocks.

Moreover, it happens more specifically when there is a combination of the following: 
 * high throughput of transactions from the same EOA
 * lower value than throughput on [Transaction Pool Options](https://geth.ethereum.org/docs/interface/command-line-options) (like `--txpool.accountqueue`)
 * a low number of nodes in the network (the lower, the higher probability that the node is unable to promote remote transactions).

The following image illustrates the issue:
Untitled-6

Therefore, this PR addresses the issue by sending the transactions to all peers, while in the meantime we work on a logarithmic propagation solution for IBFT/QBFT.

### Changes

* Broadcast the full transaction data to all peers
* Add unit test to validate the logic


eth/handler_test.go
```diff
@@ -732,3 +732,68 @@ func TestBroadcastMalformedBlock(t *testing.T) {
...
+		case <-doneCh:
+			received++
+			if received == totalPeers {
+				// We found the right number
+				return
+			}
+		case <-time.After(2 * time.Second):
+			if received != totalPeers {
+				t.Errorf("broadcast count mismatch: have %d, want %d", received, totalPeers)
+			}
```
baptiste-b-pegasys: I think this case is not useful: either you have already exited just before or here you just saying the test is a failure
```suggestion
			t.Errorf("timeout: broadcast count mismatch: have %d, want %d", received, totalPeers)
```

eth/handler_test.go
```diff
@@ -732,3 +732,68 @@ func TestBroadcastMalformedBlock(t *testing.T) {
...
+	doneCh := make(chan struct{}, totalPeers)
+	for _, peer := range peers {
+		go func(p *testPeer) {
+			if err := p2p.ExpectMsg(p.app, TransactionMsg, transactions); err != nil {
+				errCh <- err
+			} else {
+				doneCh <- struct{}{}
+			}
+		}(peer)
+	}
```
baptiste-b-pegasys: 1/3 : merge the channels, clean close of the channel when all go routines are done
```suggestion
	doneCh := make(chan err, totalPeers)

	wg := sync.WaitGroup{}
	wg.Add(totalPeers)
	defer func() {
		wg.Wait()
		close(doneCh)
	}()

	for _, peer := range peers {
		go func(p *testPeer) {
			errCh <- p2p.ExpectMsg(p.app, TransactionMsg, transactions)
			wg.Done()
		}(peer)
	}
```

eth/handler_test.go
```diff
@@ -732,3 +732,68 @@ func TestBroadcastMalformedBlock(t *testing.T) {
...
+	var received int
+	for {
+		select {
+		case <-doneCh:
```
baptiste-b-pegasys: 2/3 : merge done & err cases
```suggestion
		case err := <-doneCh:
			if err != nil {
				t.Fatalf("broadcast failed: %v", err)
				return
			}
```

eth/handler_test.go
```diff
@@ -732,3 +732,68 @@ func TestBroadcastMalformedBlock(t *testing.T) {
...
+				// We found the right number
+				return
+			}
+		case <-time.After(2 * time.Second):
+			if received != totalPeers {
+				t.Errorf("broadcast count mismatch: have %d, want %d", received, totalPeers)
+			}
+			return
+		case err := <-errCh:
+			t.Fatalf("broadcast failed: %v", err)
```
baptiste-b-pegasys: 3/3: remove err case
```suggestion
```

eth/handler_test.go
```diff
@@ -732,3 +732,68 @@ func TestBroadcastMalformedBlock(t *testing.T) {
...
+	defer pm.Stop()
+
+	var peers []*testPeer
+	for i := 0; i < totalPeers; i++ {
+		peer, _ := newTestPeer(fmt.Sprintf("peer %d", i), eth65, pm, true)
+		defer peer.close()
+
+		peers = append(peers, peer)
+	}
+	time.Sleep(1 * time.Second) // wait until all peers are synced before pushing tx to the pool
```
baptiste-b-pegasys: Could cause issue when system is overloaded or too much wait for nothing. Is it possible to have a state to test?

eth/handler_test.go
```diff
@@ -732,3 +732,68 @@ func TestBroadcastMalformedBlock(t *testing.T) {
...
+	blockchain, _ := core.NewBlockChain(db, nil, config, pow, vm.Config{}, nil, nil, nil)
+	txPool := &testTxPool{pool: make(map[common.Hash]*types.Transaction)}
+
+	pm, _ := NewProtocolManager(config, nil, downloader.FullSync, DefaultConfig.NetworkId, evmux, txPool, pow, blockchain, db, 1, nil, false)
+	pm.Start(totalPeers)
+	defer pm.Stop()
+
+	var peers []*testPeer
+	for i := 0; i < totalPeers; i++ {
+		peer, _ := newTestPeer(fmt.Sprintf("peer %d", i), eth65, pm, true)
```
baptiste-b-pegasys, line 764: there is an error channel returned, should we check if there is any value in it?

baptiste-b-pegasys (commented):
I know it is inspired from code that exists priorly, but I just want to raise some concerns here about closing and emptying channels, checking all errors, and trying to be synchronous to avoid some flakiness.

eth/handler_test.go
ricardolyn: as this is a test, the test will automatically fail with a `nil` of `peer`. is it worth checking the error?

eth/handler_test.go
ricardolyn: agreed. changed this approach to use a channel from the peers.

ricardolyn:
@baptiste-b-pegasys changes done. please review again

eth/handler_test.go
```diff
@@ -732,3 +733,80 @@ func TestBroadcastMalformedBlock(t *testing.T) {
...
+// Tests that when broadcasting transactions, it sends the full transactions to all peers instead of Announcing (aka sending only hashes)
+func TestBroadcastTransactionsOnQuorum(t *testing.T) {
+	var (
+		evmux             = new(event.TypeMux)
+		pow               = ethash.NewFaker()
+		db                = rawdb.NewMemoryDatabase()
+		config            = &params.ChainConfig{}
+		gspec             = &core.Genesis{Config: config}
+		destinationKey, _ = crypto.GenerateKey()
+		totalPeers        = 100
```
achraf17, line 747: just a query: Any reason why the specific 100 peers ?

ricardolyn: a magic number on this case.

jbhurat (approved):
LGTM

achraf17 (approved):
LGTM

baptiste-b-pegasys (approved):

PR #1266, opened 2021-10-06
Receipt unit test was not checking QuorumReceiptExtraData in PSReceipt

SatpalSandhu61:
Small change to unit tests, so that failure to encode/decode any field in PSReceipts is correctly detected.

### DIFF
--- a/core/types/receipt_test.go
+++ b/core/types/receipt_test.go
@@ -138,6 +138,9 @@ func compareLogsConsensusFields(t *testing.T, encLogs []*Log, decLogs []*Log) {
 }
 
 func comparePSReceipts(t *testing.T, encPSReceipts map[PrivateStateIdentifier]*Receipt, decPSReceipts map[PrivateStateIdentifier]*Receipt, supportPTM bool) {
+	if encPSReceipts == nil {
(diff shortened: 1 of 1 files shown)


core/types/receipt_test.go
```diff
@@ -591,7 +594,7 @@ func TestQuorumReceiptExtraDataDecodingSuccess(t *testing.T) {
 	assert.Contains(t, decodedExtraData.PSReceipts, PrivateStateIdentifier("psi1"))
 	decodedReceipt := decodedExtraData.PSReceipts[PrivateStateIdentifier("psi1")]
 	assert.NotNil(t, decodedReceipt)
-	testConsensusFields(t, ReceiptForStorage(*decodedReceipt), receipt, true, true, true)
```
ricardolyn, line 594: no need anymore to do `testConsensusFields(...)`?

SatpalSandhu61: No, as testConsensusFields() expects a Receipt object which contains PSReceipts. I made a change to testConsensusFields() as it would previously pass if the PSReceipts object was actually missing, so that can no longer be used here.
In this case we just want to test the contents of PSReceipts itself, hence I use comparePSReceipts().

ricardolyn (approved):

PR #1272, opened 2021-10-28
Unit test for deadlock detection when testQBFTBlock is set to 1

jbhurat:
Commit _2a8310663ecafc0233758ca7883676bf568e926e_ added  a fix for a deadlock that occurred when _testQBFTBlock_ fork block was set to 1. This PR adds a unit test for that fix.

### DIFF
--- a/consensus/istanbul/backend/backend_test.go
+++ b/consensus/istanbul/backend/backend_test.go
@@ -188,6 +188,38 @@ func TestGetProposer(t *testing.T) {
 	}
 }
 
+// TestQBFTTransitionDeadlock test whether a deadlock occurs when testQBFTBlock is set to 1
+// This was fixed as part of commit 2a8310663ecafc0233758ca7883676bf568e926e
+func TestQBFTTransitionDeadlock(t *testing.T) {
+	timeout := time.After(1 * time.Minute)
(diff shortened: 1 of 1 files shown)


consensus/istanbul/backend/backend_test.go
```diff
@@ -188,6 +188,34 @@ func TestGetProposer(t *testing.T) {
...
+		block := makeBlock(chain, engine, chain.Genesis())
+		_, err := chain.InsertChain(types.Blocks{block})
+		if err != nil {
+			t.Errorf("Error inserting block: %v", err)
+		}
+
+		if err = engine.NewChainHead(); err != nil {
+			t.Errorf("Error posting NewChainHead Event: %v", err)
+		}
+		done <- true
```
ricardolyn, line 213: should we add an extra check like:

```suggestion
		qbftConsensus = engine.IsQBFTConsensus()
	        if !qbftConsensus {
		        t.Errorf("IsQBFTConsensus() should return true after block insertion")
	        }
	        done <- true
```

jbhurat: Added the suggested check

ricardolyn (approved):

PR #1273, opened 2021-10-28
Upstream merge fixes

chris-j-h:
Fixes relating to recent upstream geth merges.  

* https://github.com/ethereum/go-ethereum/pull/21509
    * The upstream merge introduced `ActivePrecompiles()` however there is a typo causing tests added in this PR to fail.  This bug was fixed in upstream 1.10.2 - the content of that fix has been included in this change.  To simplify future merges the content of the commit has been copied instead of cherry-picking the commit.
    * The Quorum Privacy Precompile address has been added to `ActivePrecompiles()` list. This is currently only used for the experimental YoloV2 EIP 2929 (where all active precompiles for the current block height must be known) so not critical but worth fixing ahead of the concrete release that will be pulled in from a future upstream geth merge.

### DIFF
--- a/core/vm/contracts.go
+++ b/core/vm/contracts.go
@@ -113,7 +113,7 @@ func init() {
 		PrecompiledAddressesHomestead = append(PrecompiledAddressesHomestead, k)
 	}
 	for k := range PrecompiledContractsByzantium {
-		PrecompiledAddressesHomestead = append(PrecompiledAddressesByzantium, k)
+		PrecompiledAddressesByzantium = append(PrecompiledAddressesByzantium, k)
 	}
 	for k := range PrecompiledContractsIstanbul {
 		PrecompiledAddressesIstanbul = append(PrecompiledAddressesIstanbul, k)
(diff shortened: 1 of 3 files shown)


Krish1979 (dismissed):
LGTM

baptiste-b-pegasys (dismissed):

antonydenyer (changes requested):
I don't think cherry-picking upstream changes makes sense. We should endeavour to stay up to date with geth changes. Pulling in an additional changes that aren't part of geth v1.9.25 makes it difficult to track and maintain what is going on with quorum.

chris-j-h:
@antonydenyer would you rather getting rid of the upstream cherry-pick and making the bug fix as part of commit a8338d8?

antonydenyer:
:+1: Yeah let's do that for now.

antonydenyer (approved):

Krish1979 (approved):
LGTM

baptiste-b-pegasys (approved):

PR #1276, opened 2021-11-09
fix: flacky ATs: limit number of concurrent ATs running and the size of the matrix

baptiste-b-pegasys:


ricardolyn:
what is the goal of this?

baptiste-b-pegasys:
It's for controlling the ressources consumed by the ATs in github action. I run many flacky ATs locally without any problem on my machine.

I made some tests with GitHub Action, and notice also that there is some issue where the matrix is big, even when I limit the number of concurrent ATs running, and splitting the matrix seems to help.

GitHub Action is like Azure pipeline, it's running in a VM of 2 VCPU. Not controlling the number of processes running can produce starvation effect in CPU / Memory.

Matrix seems to becleaned at the end, therefore large matrix may consume too much ressources. The clue was that Tessera often has problem to start because some ressource is already closed.

In the end, for a PR that takes 18min, it will take 30 minutes with this setup. (1,5x)

https://github.com/baptiste-b-pegasys/quorum-acceptance-tests/actions/runs/1435263505

antonydenyer (approved):

PR #1283, opened 2021-12-16
Fix issue with inconsistent private state root when using MPS with PMT.

SatpalSandhu61:
When using Privacy Marker Transactions with MPS, the private state (hence, the privateStateRoot) is different between the MPS and non-MPS nodes.
This is due to the following:

- An entry is being created in the private state for the precompile account ("0x000000000000000000000000000000000000007a")
- Because there is no state change, this entry is normally removed from the statedb by `statedb.Finalise()`. This only removes objects which have an entry in ` s.journal.dirties` and the value of `obj.empty()` is true.
- However, on nodes where MPS is enabled, the state is copied using `statedb.Copy()`, which does NOT copy the journal. Hence the precompile account isn't removed and gets included in the state root.

The proposed fix here, is to NOT add the precompile account to the private state (this makes sense as there should never be any state changes to the precompile account).
Note that any users who are already using PMT should resync their nodes to ensure that the privateStateRoot is corrected. Otherwise they may have issues in the future if any GoQuorum feature expects the privateStateRoot to be in sync between nodes.

Also extended the `debug.dumpBlock()` API to accept an optional second parameter. This allows private state to be viewed using:
```
debug.dumpBlock(eth.blockNumber, "private")

 ```


nicolae-leonte-go (approved):
LGTM

Krish1979 (approved):

antonydenyer (approved):
Agreed, this is in line with what happens in besu, precompile contracts should not affect the state root (public or private). 

Would be good to have a test to that effect either in here or in [quorum acceptance tests.](github.com/consenSys/quorum-acceptance-tests).

Issue #1287, opened 2021-12-30
GoQuorum Geth Crash with fatal error: concurrent map iteration and map write


frankie-lim-partior:
#### System information

Geth version: `geth version`
Geth
Version: 1.9.25-stable
Git Commit: 919800f019cc5d2b931b5cd81600640a8e7cd444
Quorum Version: 21.10.0
Architecture: amd64
Protocol Versions: [65 64 63]
Network Id: 1337
Go Version: go1.15.6
Operating System: linux
GOPATH=
GOROOT=go

OS & Version: Windows/Linux/OSX
Running on Google CLoud Platform GKE Kubernetes
/ # cat /etc/os-release
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.14.2
PRETTY_NAME="Alpine Linux v3.14"
HOME_URL="https://alpinelinux.org/"
BUG_REPORT_URL="https://bugs.alpinelinux.org/"

#### Expected behaviour
When transactions are send to the Geth RPC, it continue to process the txn and return receipt without crashing.

#### Actual behaviour
Geth (GoQuorum) crash with "fatal error: concurrent map iteration and map write" randomly, especially when we are sending > ~10 send transaction call per seconds to Geth.

#### Steps to reproduce the behaviour
Keep sending private smart contract transaction to Geth, at more than ~10 Txn per seconds, it would randomly crash. With public smart contract transaction, it does not crash even with 100 TPS.

#### Backtrace
Full log attached

````
fatal error: concurrent map iteration and map write

goroutine 82381 [running]:
runtime.throw(0x17480af, 0x26)
	runtime/panic.go:1116 +0x72 fp=0xc00db61980 sp=0xc00db61950 pc=0x450c52
runtime.mapiternext(0xc00db61aa8)
	runtime/map.go:853 +0x554 fp=0xc00db61a00 sp=0xc00db61980 pc=0x429f14
github.com/ethereum/go-ethereum/core/state.Storage.Copy(0xc00b8561e0, 0xc00bc9cc00)
	github.com/ethereum/go-ethereum/core/state/state_object.go:54 +0x156 fp=0xc00db61b18 sp=0xc00db61a00 pc=0x9541b6
github.com/ethereum/go-ethereum/core/state.(*stateObject).deepCopy(0xc00da03400, 0xc00db86000, 0xc00db61c98)
[...]

PR #1302, opened 2022-01-12
Fix errors when there is high load

baptiste-b-pegasys:
- concurrent map modification
- seg fault


antonydenyer (dismissed):

core/blockchain.go
```diff
@@ -427,20 +427,21 @@ func NewBlockChain(db ethdb.Database, cacheConfig *CacheConfig, chainConfig *par
...
 			defer bc.wg.Done()
-			privatetriedb.TrieDB().SaveCachePeriodically(bc.cacheConfig.PrivateTrieCleanJournal, bc.cacheConfig.TrieCleanRejournal, bc.quit)
+			bc.saveCachePeriodically(triedb, privatetriedb)
 		}()
 	}
 	return bc, nil
 }
 
+func (bc *BlockChain) saveCachePeriodically(triedb *trie.Database, privatetriedb mps.PrivateStateManager) {
+	triedb.SaveCachePeriodically(bc.cacheConfig.TrieCleanJournal, bc.cacheConfig.TrieCleanRejournal, bc.quit)
```
nicolae-leonte-go: Doesn't this mean that `triedb.SaveCachePeriodically` is executed and does not return until the `bc.quit` channel is closed and of course `privatetriedb.TrieDB().SaveCachePeriodically` is only invoked once the `bc.quit` channel has already been closed (in which case it is unlikely to do much)?

core/mps/default_psr.go
```diff
@@ -39,6 +39,9 @@ func NewDefaultPrivateStateRepository(db ethdb.Database, cache state.Database, p
 }
 
 func (dpsr *DefaultPrivateStateRepository) DefaultState() (*state.StateDB, error) {
+	if dpsr == nil {
```
nicolae-leonte-go, line 42: I've tried to reproduce a situation where one is able to invoke `dpsr.DefaultState()` with `dpsr` being nil but couldn't figure out how (I get the panic before getting inside the `DefaultState()` method).

nicolae-leonte-go:
Adding the mutex to StateObject appears to solve the copy problem (which is obviously a good thing). 
What I don't understand is why the copy issue only happens for the private state (as the same problem should be reproducible for the public state).
Another concern is also whether we should consider adding the mutex further up (at StateDB level).

core/mps/default_psr.go
baptiste-b-pegasys: I don't know, it seems impossible.
Yet it is reproducable, and I asked to have the same env in order to reproduce the error.
https://github.com/ConsenSys/quorum/issues/1287?id=issuecomment-1009834186#issuecomment-1007232132

core/blockchain.go
baptiste-b-pegasys: It's a guess, and it's been a while I wanted to refactor this part.

Because in the issue, there is a complain that the data is not here for the private state.

baptiste-b-pegasys: a part I made.
What version do you prefer?

nicolae-leonte-go (approved):
LGTM

antonydenyer (approved):

PR #1330, opened 2022-02-21
fix: ibft1 version for support of besu

antonydenyer:


eth/protocols/eth/handshake.go
```diff
@@ -94,7 +95,7 @@ func (p *Peer) readStatus(network uint64, status *StatusPacket, genesis common.H
 	if status.NetworkID != network {
 		return fmt.Errorf("%w: %d (!= %d)", errNetworkIDMismatch, status.NetworkID, network)
 	}
-	if uint(status.ProtocolVersion) != p.version {
+	if uint(status.ProtocolVersion) != p.version && !(strings.HasPrefix(p.Name(), "besu") && status.ProtocolVersion == 99) {
```
antonydenyer, line 98: Do we really need `strings.HasPrefix(p.Name(), "besu")`

achraf17: yes, at this level we dont have the into about the peer caps as we already changed it to what we want. The only availbale info is the name

PR #1353, opened 2022-03-07
Feature/public and private states shared cache

nicolae-leonte-go:
This change allows the private state (or states in case of MPS) to behave similarly to the public state when it comes to what is written to the disk and how often.
Before this PR the private state is being flushed to the disk for each block (even though the corresponding public state is not - depending on the `--gcmode` parameter value).
The behaviour is enabled by passing a command line parameter: `--privatetriecache.enable`

When enabled, the public and the private state(s) share the same `state.Database` instance. The single cache management (previously applying only to the public state) now also applies to the private state(s). The `TrieDB.Reference` mechanism is used to link the various tries with each-other so that when a `TrieDB.Commit` on the public state root is issued it cascades to all other referenced tries (including the private states).

When disabled (the parameter is missing from the command line) the private state(s) is still committed for each block with **ONE NOTABLE DIFFERENCE**: the order of commits **is reversed**  - public state is committed first followed by the private state(s). The reason for the reversal is that when we commit the private state the public state root is already in the cache (which then allows us to add the reference from the private state root to the public state root).


Krish1979 (dismissed):

antonydenyer (dismissed):
An elegant solution that will mean we can benefit more easily from upstream changes. :pray:

Krish1979 (approved):

achraf17 (approved):

PR #1371, opened 2022-04-05
feat: Add support for transitions

achraf17:
Added support for transitions to switch from IBFT to QBFT dynamically  and for other fields like:
- RequestTimeout
- Epoch
- BlockPeriod

### DIFF
--- a/consensus/istanbul/config.go
+++ b/consensus/istanbul/config.go
@@ -20,6 +20,7 @@ import (
 	"math/big"
 	"sync"
 
+	"github.com/ethereum/go-ethereum/params"
 	"github.com/naoina/toml"
 )
 
@@ -125,6 +126,7 @@ type Config struct {
 	Ceil2Nby3Block         *big.Int        `toml:",omitempty"` // Number of confirmations required to move from one state to next [2F + 1 to Ceil(2N/3)]
 	AllowedFutureBlockTime uint64          `toml:",omitempty"` // Max time (in seconds) from current time allowed for blocks, before they're considered future blocks
 	TestQBFTBlock          *big.Int        `toml:",omitempty"` // Fork block at which block confirmations are done using qbft consensus instead of ibft
+	Transitions            []params.Transition
 }
 
 var DefaultConfig = &Config{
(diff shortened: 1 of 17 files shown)


antonydenyer (dismissed):
LGTM

consensus/istanbul/config.go
```diff
@@ -147,17 +149,35 @@ func (c Config) QBFTBlockNumber() int64 {
 
 // IsQBFTConsensusAt checks if qbft consensus is enabled for the block height identified by the given header
 func (c *Config) IsQBFTConsensusAt(blockNumber *big.Int) bool {
-	// If qbftBlock is not defined in genesis qbft consensus is not used
-	if c.TestQBFTBlock == nil {
-		return false
+	if c.TestQBFTBlock != nil {
```
baptiste-b-pegasys, line 152: should we add a deprecation comment?

consensus/istanbul/config.go
```diff
@@ -147,17 +149,35 @@ func (c Config) QBFTBlockNumber() int64 {
...
+		}
+
+		if blockNumber.Cmp(c.TestQBFTBlock) >= 0 {
+			return true
+		}
 	}
-
-	if c.TestQBFTBlock.Uint64() == 0 {
-		return true
+	for i := 0; c.Transitions != nil && i < len(c.Transitions) && c.Transitions[i].Block.Cmp(blockNumber) <= 0; i++ {
```
baptiste-b-pegasys, line 161: should we put this block at the beginning of the func because it's the new and right version?
if there is both testqbft and transitions, should we take transitions in priority?

consensus/istanbul/config.go
```diff
@@ -147,17 +149,35 @@ func (c Config) QBFTBlockNumber() int64 {
...
+		}
 	}
+	return false
+}
 
-	if blockNumber.Cmp(c.TestQBFTBlock) >= 0 {
-		return true
+func (c Config) GetConfig(blockNumber *big.Int) Config {
+	newConfig := c
+	for i := 0; c.Transitions != nil && i < len(c.Transitions) && c.Transitions[i].Block.Cmp(blockNumber) <= 0; i++ {
```
baptiste-b-pegasys, line 171: is c.Transitions ordered for sure?
https://go.dev/play/p/PX4bREJAiEg

params/config.go
```diff
@@ -336,11 +336,12 @@ type ChainConfig struct {
...
-	Clique   *CliqueConfig   `json:"clique,omitempty"`
-	Istanbul *IstanbulConfig `json:"istanbul,omitempty"` // Quorum
-	IBFT     *IBFTConfig     `json:"ibft,omitempty"`     // Quorum
-	QBFT     *QBFTConfig     `json:"qbft,omitempty"`     // Quorum
+	Ethash      *EthashConfig   `json:"ethash,omitempty"`
+	Clique      *CliqueConfig   `json:"clique,omitempty"`
+	Istanbul    *IstanbulConfig `json:"istanbul,omitempty"`    // Quorum
+	IBFT        *IBFTConfig     `json:"ibft,omitempty"`        // Quorum
+	QBFT        *QBFTConfig     `json:"qbft,omitempty"`        // Quorum
+	Transitions []Transition    `json:"transitions,omitempty"` // Quorum
```
baptiste-b-pegasys: as it is not a consensus engine, should we move it in quorum specific settings?

consensus/istanbul/config.go
achraf17: we added [one](https://github.com/ConsenSys/quorum/blob/master/cmd/utils/flags.go#L2217) for the `config.istanbul` which `TestQBFTBlock` is under it.

consensus/istanbul/config.go
achraf17: currently we are supporting both transitions and testqbft but not at the same time when we have testqbft and transition of Algorithm. so in a way it does not matter which block is first.

consensus/istanbul/config.go
achraf17: That is being verified at the init command by calling `CheckTransitionsData()` [code](https://github.com/ConsenSys/quorum/pull/1371/files#diff-10ff00a15ef58da7485fedeca3906c73cb236fde4b143a9d61601c63cbf1ffe8R613)

params/config.go
achraf17: updated

baptiste-b-pegasys (approved):

PR #1373, opened 2022-04-07
Fix: update transitions for old config 

achraf17:
When updating the genesis it is possible to have a situation where the chain will continue to work but will fail when you re-sync your node.

For example:

Initialise `testQBFTBlock=100` you reach block 101
Change `testQBFTBlock=99` restart you node and it continues from 102 ...

Something happens and you need to re-sync, at block 99/100 you will have a failure/fork.


antonydenyer (approved):

params/config.go
```diff
@@ -695,6 +695,70 @@ func isMaxCodeSizeConfigCompatible(c1, c2 *ChainConfig, head *big.Int) (error, *
...
+	}
+
+	c2RecsBelowHead := 0
+	for _, data := range c2.Transitions {
+		if data.Block.Cmp(head) <= 0 {
+			c2RecsBelowHead++
+		} else {
+			break
+		}
+	}
```
baptiste-b-pegasys, line 735: we should factorize these to a single function to avoid duplicate

params/config.go
```diff
@@ -695,6 +695,70 @@ func isMaxCodeSizeConfigCompatible(c1, c2 *ChainConfig, head *big.Int) (error, *
...
+
+	// if the count of past records is not matching return error
+	if c1RecsBelowHead != c2RecsBelowHead {
+		return errors.New("transitions data incompatible. updating transitions for past"), head, head
+	}
+
+	// validate that each past record is matching exactly. if not return error
+	for i := 0; i < c1RecsBelowHead; i++ {
+		isSameBlock := c1.Transitions[i].Block.Cmp(c2.Transitions[i].Block) != 0
```
baptiste-b-pegasys, line 744: if it is not the same block, that means some block(s) has(ve) disappear?
should we raise an error?

params/config.go
```diff
@@ -724,6 +788,11 @@ func (c *ChainConfig) CheckCompatible(newcfg *ChainConfig, height uint64, isQuor
 	if err != nil {
 		return newCompatError(err.Error(), cBlock, newCfgBlock)
 	}
+	// compare the transitions data between the old and new config
+	err, cBlock, newCfgBlock = isTransitionsConfigCompatible(c, newcfg, bhead)
```
baptiste-b-pegasys, line 792: so cBlock and newCfgBlock are used for creating a compat error, why not returning this error directly from the isTransiti...Compatible()?

also isTransi...Compatible(), I thought it would return a bool, could it be named checkCompatibilityOfConfigs()?

PR #1376, opened 2022-04-08
Feat: transition of contract size limit

achraf17:


params/config.go
```diff
@@ -430,6 +431,7 @@ type Transition struct {
 	EpochLength           uint64   `json:"epochlength,omitempty"`           // Number of blocks that should pass before pending validator votes are reset
 	BlockPeriodSeconds    uint64   `json:"blockperiodseconds,omitempty"`    // Minimum time between two consecutive IBFT or QBFT blocks’ timestamps in seconds
 	RequestTimeoutSeconds uint64   `json:"requesttimeoutseconds,omitempty"` // Minimum request timeout for each IBFT or QBFT round in milliseconds
+	ContractSizeLimit     uint64   `json:"contractSizeLimit,omitempty"`     // Maximum smart contract code size
```
baptiste-b-pegasys: it's not camel case for the others

params/config.go
```diff
@@ -580,6 +583,13 @@ func (c *ChainConfig) GetMaxCodeSize(num *big.Int) int {
 			maxCodeSize = int(c.MaxCodeSize) * 1024
 		}
 	}
+	if c.Transitions != nil && len(c.Transitions) > 0 {
```
baptiste-b-pegasys: the nil check is not useful, len(nil) will give 0
```suggestion
	if len(c.Transitions) > 0 {
```

params/config.go
achraf17: updated

params/config.go
achraf17: updated

antonydenyer (approved):

baptiste-b-pegasys (approved):

PR #1380, opened 2022-04-13
qlight token manager plugin

baptiste-b-pegasys:
- [x] Move this plugin to ConsenSys area: https://github.com/baptiste-b-pegasys/quorum-plugin-qlight-token-manager


achraf17 (approved):
LGTM

eth/backend.go
```diff
@@ -284,12 +286,15 @@ func New(stack *node.Node, config *ethconfig.Config) (*Ethereum, error) {
...
+			if err != nil {
+				return nil, fmt.Errorf("new token holder: %w", err)
+			}
 			switch eth.config.QuorumLightClient.TokenManagement {
 			case "client-security-plugin":
-				// TODO QLight - when hte client-security-plugin is implemented this may be a good place to initialize it
-				return nil, fmt.Errorf("The client-scurity-plugin token management is not implemented")
+				log.Info("Starting qlight client with auth token enabled without external API and token from argument, plugin has to be provided")
 			case "none":
+				eth.qlightTokenHolder.SetCurrentToken(eth.config.QuorumLightClient.TokenValue)
```
baptiste-b-pegasys, line 296: no plugin manager

qlight/token_holder.go
```diff
@@ -1,11 +1,96 @@
...
+		log.Warn("token holder nil, returning empty token")
+		return ""
+	}
+	if h.plugin != nil {
+		returnedToken, err := h.plugin.TokenRefresh(context.Background(), h.token, h.psi)
+		if err != nil {
+			log.Error("get token from plugin", "err", err)
+		} else {
+			if h.token != returnedToken {
+				log.Debug("new token from plugin")
```
baptiste-b-pegasys: notify p2p qlight server

qlight/token_holder.go
```diff
@@ -1,11 +1,96 @@
...
+	return h.token, nil
+}
+
+func (h *TokenHolder) GetCurrentToken() string {
+	if h == nil {
+		log.Warn("token holder nil, returning empty token")
+		return ""
+	}
+	if h.plugin != nil {
+		returnedToken, err := h.plugin.TokenRefresh(context.Background(), h.token, h.psi)
```
baptiste-b-pegasys: reduce calls here with JWT expiration test

qlight/token_holder.go
```diff
@@ -1,11 +1,96 @@
...
+
+func NewTokenHolderWithPlugin(psi string, plugin qlightplugin.PluginTokenManager) *TokenHolder {
+	return &TokenHolder{
+		psi:    psi,
+		plugin: plugin,
+	}
+}
+
+func (h *TokenHolder) RefreshPlugin(pluginManager *plugin.PluginManager) (err error) {
+	h.plugin, err = getPlugin(pluginManager)
```
baptiste-b-pegasys: configurable expiration delay check

go.mod
```diff
@@ -12,6 +12,7 @@ require (
 	github.com/BurntSushi/toml v0.3.1
 	github.com/VictoriaMetrics/fastcache v1.5.7
 	github.com/aws/aws-sdk-go v1.25.48
+	github.com/baptiste-b-pegasys/quorum-plugin-qlight-token-manager v0.0.0-20220426133706-3b9390db12f7
```
baptiste-b-pegasys: must move the plugin before any merge

antonydenyer (changes requested):
LGTM

qlight/api.go
```diff
@@ -17,17 +18,24 @@ func NewPrivateQLightAPI(peerUpdater RunningPeerAuthUpdater, rpcClient *rpc.Clie
 }
 
 func (p *PrivateQLightAPI) SetCurrentToken(token string) {
-	SetCurrentToken(token)
+	if token == p.GetCurrentToken() {
+		return
+	}
```
baptiste-b-pegasys:
```suggestion
```

qlight/token_holder.go
```diff
@@ -1,11 +1,177 @@
...
+	if !expired {
+		return h.token
+	}
+	returnedToken, err := h.plugin.TokenRefresh(context.Background(), h.token, h.psi)
+	if err != nil {
+		log.Error("get token from plugin", "err", err)
+	} else {
+		if h.token != returnedToken {
+			log.Debug("new token from plugin")
+			err = h.peerUpdater.UpdateTokenForRunningQPeers(returnedToken)
```
nicolae-leonte-go, line 139: From what I understand if you keep calling `CurrentToken()` (most likely due to RPC requests that must be forwarded to the qlight server node) the corresponding token will also be updated in the P2P connection (whenever it is refreshed).
What happens if the qlight client node DOES NOT make any RPC requests (sendTransaction*) that are forwarded to the qlight server? 
I suspect the qlight server will close the P2P connection because the token expires and when a new connection is established the qlight client will get a new token. Is that the expected behaviour?
I was kind of expecting some timer logic to prevent the P2P connection from being dropped.

plugin/qlight/gateway.go
```diff
@@ -0,0 +1,33 @@
...
+		Psi:          psi,
+	})
+	if err != nil {
+		return "", fmt.Errorf("refresh token: %w", err)
+	}
+	return resp.Token, nil
+}
+
+func (p *PluginGateway) PluginTokenManager(ctx context.Context) (int32, error) {
+	resp, err := p.client.PluginQLightTokenManager(ctx, &proto.PluginQLightTokenManager_Request{})
```
chris-j-h, line 28: One potential concern with having the plugin return a statically configured refresh interval is that it couples the plugin configuration to the oauth2 server configuration.  For example, if the the oauth2 server admin decides to change the expiration period of generated tokens (say from 1hr to 30mins, perhaps due to a change in policy) then this will require action from the quorum admin to update the plugin config with a new `RefreshAnticipationInMillisecond` and reload the plugin, otherwise requests will start failing with expired token errors.

If we're happy with that then fine, but if we want to address that then an alternative could be for the plugin to use the expiration time returned in the JWT token (if there is one) instead of having the quorum admin provide the expected refresh period in config.

baptiste-b-pegasys: it's an anticipation, that is the expiry time/delay - anticipation time/delay.
If anticipation is greater than expiry, it will end with a refresh at every request.
In which case no refresh would happen?

qlight/token_holder.go
baptiste-b-pegasys: I will add a timer that will do a default refresh. Thank you for raising this.

plugin/qlight/gateway.go
chris-j-h: sure - and if the anticipation is longer than the expiry the request would fail because the token would have expired correct? in that case does it trigger an refresh?

baptiste-b-pegasys: the refresh is made before the request

qlight/token_holder.go
baptiste-b-pegasys: I've updated the code

antonydenyer (approved):

qlight/token_holder.go
nicolae-leonte-go: LGTM

nicolae-leonte-go (approved):

chris-j-h (approved):

baptiste-b-pegasys:
plugin has been pushed to cloudsmith, quorum-examples work with it : https://github.com/baptiste-b-pegasys/quorum-examples/pull/1

PR #1381, opened 2022-04-14
Feat: Smart contract validators

achraf17:
Allow users to specify a smart contract address at a given block height that will enable the validator list to be managed by that contract from that height.

The only interface that `GoQuorum` cares about is `function getValidators() external view returns (address[] memory)` how the internal list of validators is updated and agreed upon is up to the contract implementation. There is a reference implementation available https://github.com/ConsenSys/validator-smart-contracts


consensus/istanbul/backend/engine.go
```diff
@@ -360,18 +363,43 @@ func (sb *Backend) snapshot(chain consensus.ChainHeaderReader, number uint64, ha
...
+				log.Info("Initialising snap with contract validators", "address", sb.config.ValidatorContract, "client", sb.config.Client)
 
-			snap = newSnapshot(sb.config.GetConfig(new(big.Int).SetUint64(number)).Epoch, 0, genesis.Hash(), validator.NewSet(validators, sb.config.ProposerPolicy))
-			if err := sb.storeSnap(snap); err != nil {
-				return nil, err
+				validatorContractCaller, err := contract.NewValidatorContractInterfaceCaller(sb.config.ValidatorContract, sb.config.Client)
+
+				if err != nil {
+					log.Error("BFT: invalid smart contract in genesis alloc", "err", err)
+					return nil, err
```
baptiste-b-pegasys: it would be better to wrap the error and not log it (double log?)
```suggestion
					return nil, fmt.Errorf("invalid smart contract in genesis alloc: %w", err)
```

consensus/istanbul/config.go
```diff
@@ -181,3 +185,13 @@ func (c Config) GetConfig(blockNumber *big.Int) Config {
 	}
 	return newConfig
 }
+
+func (c Config) GetValidatorContractAddress(blockNumber *big.Int) common.Address {
+	validatorContractAddress := c.ValidatorContract
+	for i := 0; c.Transitions != nil && i < len(c.Transitions) && c.Transitions[i].Block.Cmp(blockNumber) <= 0; i++ {
```
baptiste-b-pegasys, line 191: should we start from the end and break the loop at first occurence found?

consensus/istanbul/config_test.go
```diff
@@ -51,8 +52,8 @@ func TestGetConfig(t *testing.T) {
 		t.Errorf("error default config:\nexpected: %v\n", DefaultConfig)
 	}
 
-	config := DefaultConfig
-	config.Transitions = []params.Transition{{
+	config := &DefaultConfig
```
baptiste-b-pegasys: is there a new issue here as we store the pointer and then we use the same pointer to config3 and config5?

consensus/istanbul/config.go
achraf17: its kind of the same starting from the end or the start, it goes either way depending on the blockNumber

consensus/istanbul/config_test.go
achraf17: yes, as we are changing the values while i want a copy

antonydenyer (approved):

consensus/istanbul/backend/engine.go
```diff
@@ -406,6 +432,31 @@ func (sb *Backend) snapshot(chain consensus.ChainHeaderReader, number uint64, ha
...
 
+	validatorContract := sb.config.GetValidatorContractAddress(new(big.Int).SetUint64(targetBlockHeight))
+	if validatorContract != (common.Address{}) {
+		log.Trace("Applying snap with smart contract validators", "address", validatorContract, "client", sb.config.Client)
+
+		validatorContractCaller, err := contract.NewValidatorContractInterfaceCaller(validatorContract, sb.config.Client)
+
+		if err != nil {
+			log.Error("BFT: invalid smart contract in genesis alloc", "err", err)
+			return nil, err
```
baptiste-b-pegasys: avoid possible double logging. can wrap error like this:
```suggestion
			return nil, fmt.Errorf("BFT: invalid smart contract in genesis alloc: %w", err)
```

consensus/istanbul/config_test.go
```diff
@@ -51,8 +52,8 @@ func TestGetConfig(t *testing.T) {
...
+	(*config).Transitions = []params.Transition{{
```
baptiste-b-pegasys: why? DefaultConfig is already a pointer.
Also, The problem with pointer is that Go is copying the address not the object.

1. having the pointer doesn't copy the value, so DefaultConfig is modified for me.
2. you need the (*...) because you have a pointer of the pointer

consensus/istanbul/config_test.go
```diff
@@ -126,3 +127,41 @@ func TestIsQBFTConsensusAt(t *testing.T) {
...
+		{3, address1},
+		{4, address2},
+		{5, address2},
+		{10, address3},
+		{100, address3},
+	}
+
+	for _, test := range tests {
+		c := config.GetValidatorContractAddress(big.NewInt(test.blockNumber))
+		if !reflect.DeepEqual(c, test.expectedAddress) {
```
baptiste-b-pegasys, line 163: don't we have an equal function for big int type? reflect deep equal is the least performant solution

https://pkg.go.dev/math/big?utm_source=gopls#Int.Cmp

consensus/istanbul/config_test.go
```diff
@@ -126,3 +127,41 @@ func TestIsQBFTConsensusAt(t *testing.T) {
 		}
 	}
 }
+
+func TestGetValidatorContractAddress(t *testing.T) {
+	address, address1, address2, address3 := common.Address{}, common.Address{0x2}, common.Address{0x4}, common.Address{0x6}
+
+	config := *DefaultConfig
```
baptiste-b-pegasys, line 134: here it's *, so it will copy the value, not the pointer, so DefaultConfig is not modified.
Proper fix would be to make DefaultConfig not a pointer but the value directly.

baptiste-b-pegasys (changes requested):

consensus/istanbul/config_test.go
antonydenyer: The return type is `common.Address` which is a `[]byte`

consensus/istanbul/backend/engine.go
```diff
@@ -360,11 +364,33 @@ func (sb *Backend) snapshot(chain consensus.ChainHeaderReader, number uint64, ha
...
+				}
+
+				opts := bind.CallOpts{
+					Pending:     false,
+					BlockNumber: big.NewInt(0),
+				}
+				validators, err = validatorContractCaller.GetValidators(&opts)
+				if err != nil {
+					log.Error("BFT: invalid smart contract in genesis alloc", "err", err)
+					return nil, err
```
baptiste-b-pegasys, line 384:
```suggestion
					return nil, fmt.Errorf("BFT: invalid smart contract in genesis alloc: %w", err)
```
```suggestion
					log.Error("BFT: invalid smart contract in genesis alloc", "err", err)
					return nil, err
```

consensus/istanbul/backend/engine.go
```diff
@@ -360,11 +364,33 @@ func (sb *Backend) snapshot(chain consensus.ChainHeaderReader, number uint64, ha
...
 			}
 
-			// Get the validators from genesis to create a snapshot
-			validators, err := sb.EngineForBlockNumber(big.NewInt(0)).Validators(genesis)
-			if err != nil {
-				sb.logger.Error("BFT: invalid genesis block", "err", err)
-				return nil, err
+			var validators []common.Address
+			if sb.config.ValidatorContract != (common.Address{}) {
+				log.Info("Initialising snap with contract validators", "address", sb.config.ValidatorContract, "client", sb.config.Client)
```
baptiste-b-pegasys:
```suggestion
				sb.logger.Info("Initialising snap with contract validators", "address", sb.config.ValidatorContract, "client", sb.config.Client)
```

baptiste-b-pegasys, line 369:
```suggestion
				sb.logger.Info("Initialising snap with contract validators", "address", sb.config.ValidatorContract, "client", sb.config.Client)
```
```suggestion
				log.Info("Initialising snap with contract validators", "address", sb.config.ValidatorContract, "client", sb.config.Client)
```

consensus/istanbul/backend/engine.go
```diff
@@ -406,6 +432,30 @@ func (sb *Backend) snapshot(chain consensus.ChainHeaderReader, number uint64, ha
 	}
 	sb.recents.Add(snap.Hash, snap)
 
+	validatorContract := sb.config.GetValidatorContractAddress(new(big.Int).SetUint64(targetBlockHeight))
+	if validatorContract != (common.Address{}) {
+		log.Trace("Applying snap with smart contract validators", "address", validatorContract, "client", sb.config.Client)
```
baptiste-b-pegasys:
```suggestion
		sb.logger.Trace("Applying snap with smart contract validators", "address", validatorContract, "client", sb.config.Client)
```

consensus/istanbul/backend/engine.go
```diff
@@ -406,6 +432,30 @@ func (sb *Backend) snapshot(chain consensus.ChainHeaderReader, number uint64, ha
...
+			Pending:     false,
+			BlockNumber: new(big.Int).SetUint64(targetBlockHeight),
+		}
+		validators, err := validatorContractCaller.GetValidators(&opts)
+
+		if err != nil {
+			log.Error("BFT: invalid validator smart contract", "err", err)
+			return nil, err
+		}
+		log.Trace("Fetched validators from smart contract", "validators", validators)
```
baptiste-b-pegasys:
```suggestion
		sb.logger.Trace("Fetched validators from smart contract", "validators", validators)
```

consensus/istanbul/backend/engine.go
```diff
@@ -406,6 +432,30 @@ func (sb *Backend) snapshot(chain consensus.ChainHeaderReader, number uint64, ha
...
+		}
+		opts := bind.CallOpts{
+			Pending:     false,
+			BlockNumber: new(big.Int).SetUint64(targetBlockHeight),
+		}
+		validators, err := validatorContractCaller.GetValidators(&opts)
+
+		if err != nil {
+			log.Error("BFT: invalid validator smart contract", "err", err)
+			return nil, err
```
baptiste-b-pegasys, line 452:
```suggestion
			return nil, fmt.Errorf("BFT: invalid validator smart contract: %w", err)
```

consensus/istanbul/config_test.go
achraf17: Yes, you are right, thanks for spotting that will update. my goal is to copy the value not the pointer

consensus/istanbul/ibft/engine/engine.go
```diff
@@ -276,19 +277,29 @@ func (e *Engine) Prepare(chain consensus.ChainHeaderReader, header *types.Header
...
+		header.Extra = extra
+	} else {
+
+		// add validators in snapshot to extraData's validators section
+		extra, err := prepareExtra(header, validator.SortedAddresses(validators.List()))
+		if err != nil {
+			return err
+		}
+		header.Extra = extra
+	}
```
baptiste-b-pegasys: factorize code
```suggestion
	num := big.NewInt(0).SetUint64(number - 1)
	validatorContract := e.cfg.GetValidatorContractAddress(num)
	var addresses []common.Address
	if validatorContract != (common.Address{}) && e.cfg.GetValidatorSelectionMode(num) == params.ContractMode {
		addresses = []common.Address{}
	} else {
		// add validators in snapshot to extraData's validators section
		addresses = validator.SortedAddresses(validators.List())
	}
	extra, err := prepareExtra(header, addresses)
	if err != nil {
		return err
	}
	header.Extra = extra
```

baptiste-b-pegasys (approved):

PR #1390, opened 2022-04-27
Feature: no empty block in QBFT protocol

baptiste-b-pegasys:


saltiniroberto:
Do I understand correctly that whether a block is empty is checked only at the time of sealing the block, and if the block is empty and the no-empty-block configuration parameter is on, then the sealing fails?

Is there any check performed when proposing a block?

PR #1399, opened 2022-05-10
prevent quorum from starting if privacy is enabled and syncmode is not full

nicolae-leonte-go:
prevent quorum from starting if privacy is enabled and syncmode is not full

### DIFF
--- a/eth/ethconfig/config.go
+++ b/eth/ethconfig/config.go
@@ -57,7 +57,9 @@ var LightClientGPO = gasprice.Config{
 
 // Defaults contains default settings for use on the Ethereum main net.
 var Defaults = Config{
-	SyncMode: downloader.FastSync,
+	// Quorum - make full sync the default sync mode in quorum (as opposed to upstream geth)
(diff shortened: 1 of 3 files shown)


eth/ethconfig/config.go
```diff
@@ -57,7 +57,9 @@ var LightClientGPO = gasprice.Config{
 
 // Defaults contains default settings for use on the Ethereum main net.
 var Defaults = Config{
-	SyncMode: downloader.FastSync,
+	// Quorum - make full sync the default sync mode in quorum (as opposed to upstream geth)
+	SyncMode: downloader.FullSync,
```
nicolae-leonte-go, line 61: I think `FullSync` should be the default syncmode in quorum (which is arguably chosen over vanilla geth due to its privacy features). 
**Please comment here if you think otherwise.**

antonydenyer (approved):
Agreed, nice catch.

Krish1979 (approved):

PR #1413, opened 2022-05-22
Fix potential goroutine leak in istanbul backend_test

BurtonQin:
1. In func `TestQBFTTransitionDeadlock`,
Goroutine may leak because sending to chan doneis blocked forever on [L195](https://github.com/ConsenSys/quorum/blob/c11973341c37354845d61729eb7165ad8a241593/consensus/istanbul/backend/backend_test.go#L195) when select selects `timeout` on [L217](https://github.com/ConsenSys/quorum/blob/c11973341c37354845d61729eb7165ad8a241593/consensus/istanbul/backend/backend_test.go#L217). Although `t.Fatal()` is called in this case, it won't stop the leaking, because  because "[Calling FailNow does not stop those other goroutines.](https://golang.org/pkg/testing/#T.FailNow)".The fix is to replace the unbuffered channel with a buffered channel (buffer size 1), and the semantic is not changed.  
2. Similarly, in func `TestCommit`, 
sending to chan commitChis blocked  forever on [L122](https://github.com/ConsenSys/quorum/blob/c11973341c37354845d61729eb7165ad8a241593/consensus/istanbul/backend/backend_test.go#L122) when selectselects time.Afteron [L172](https://github.com/ConsenSys/quorum/blob/c11973341c37354845d61729eb7165ad8a241593/consensus/istanbul/backend/backend_test.go#L172). The buffer size equals to the num of testCases.

### DIFF
--- a/consensus/istanbul/backend/backend_test.go
+++ b/consensus/istanbul/backend/backend_test.go
@@ -119,7 +119,6 @@ func TestCommit(t *testing.T) {
 	backend := newBackend()
 	defer backend.Stop()
 
-	commitCh := make(chan *types.Block)
 	// Case: it's a proposer, so the backend.commit will receive channel result from backend.Commit function
 	testCases := []struct {
 		expectedErr       error
@@ -147,6 +146,7 @@ func TestCommit(t *testing.T) {
 			},
 		},
 	}
+	commitCh := make(chan *types.Block, len(testCases))
 
 	for _, test := range testCases {
 		expBlock := test.expectedBlock()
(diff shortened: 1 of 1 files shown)


baptiste-b-pegasys:
sorry but make(chan) is not about making an array, default value is 1 anyway by the go language, therefore you PR doesn't bring any additionnal value
https://gobyexample.com/channels

for the other one with the size of tests, only one test is executed consequently, there you are preparing a channel with a buffer of X messages when we expect only one.

songlh:
@baptiste-b-pegasys `make(chan int, 1)` is not to create an array. Actually, it is to create a channel whose buffer size is 1. If you do not provide the extra 1, the channel buffer size is 0. 

The pull request is to avoid leaked goroutines when the buffer size is 0.

PR #1414, opened 2022-05-22
Fix potential goroutine leak in rawdb/freezer_test

BurtonQin:
In func `Test_freezer_Close`,
Goroutine may leak because sending to chan errChis blocked forever on [L21](https://github.com/ConsenSys/quorum/blob/c11973341c37354845d61729eb7165ad8a241593/core/rawdb/freezer_test.go#L21) when select selects `timeout` on [L28](https://github.com/ConsenSys/quorum/blob/c11973341c37354845d61729eb7165ad8a241593/core/rawdb/freezer_test.go#L28). Although `t.Fatal()` is called in this case, it won't stop the leaking, because  because "[Calling FailNow does not stop those other goroutines.](https://golang.org/pkg/testing/#T.FailNow)". The fix is to replace the unbuffered channel with a buffered channel (buffer size 1), and the semantic is not changed.

### DIFF
--- a/core/rawdb/freezer_test.go
+++ b/core/rawdb/freezer_test.go
@@ -18,7 +18,7 @@ func Test_freezer_Close(t *testing.T) {
 
 	// Close second time should return error but do not hang sending data to freeze.quit channel
 	timeout := time.After(1 * time.Second)
-	errCh := make(chan error)
+	errCh := make(chan error, 1)
 
 	go func() {
 		errCh <- mockFreezer.Close()


baptiste-b-pegasys:
sorry but make(chan) is not about making an array, default value is 1 anyway by the go language, therefore you PR doesn't bring any additionnal value
https://gobyexample.com/channels

songlh:
@baptiste-b-pegasys the pull request is not to make an array of channels. It is to create a channel with buffer size 1 not 0. You should check what are buffered channels in Go: https://go.dev/tour/concurrency/3

PR #1439, opened 2022-06-22
core/freezer - retry sync write to disk

antonydenyer:


core/rawdb/freezer.go
```diff
@@ -259,17 +259,28 @@ func (f *freezer) TruncateAncients(items uint64) error {
...
-	if errs != nil {
+	if errs != nil && retry < 10 {
 		return fmt.Errorf("%v", errs)
+	} else if errs != nil {
+		log.Info("RETRY %i %v", retry, errs)
+		return f.SyncRetry(retry+1)
 	}
 	return nil
 }
+// Quorum
```
baptiste-b-pegasys:
```suggestion
// End Quorum
```

core/rawdb/freezer.go
```diff
@@ -259,17 +259,28 @@ func (f *freezer) TruncateAncients(items uint64) error {
...
 		if err := table.Sync(); err != nil {
 			errs = append(errs, err)
 		}
 	}
-	if errs != nil {
+	if errs != nil && retry < 10 {
 		return fmt.Errorf("%v", errs)
+	} else if errs != nil {
+		log.Info("RETRY %i %v", retry, errs)
+		return f.SyncRetry(retry+1)
```
baptiste-b-pegasys:
```suggestion
	hasError := len(errs) > 0
	if hasError && retry < 10 {
		log.Info("sync", "retry", retry, "errors", errs)
		return f.SyncRetry(retry+1)
	} else if hasError {
		return fmt.Errorf("%v", errs)
```

baptiste-b-pegasys (changes requested):

core/rawdb/freezer.go
```diff
@@ -259,17 +259,29 @@ func (f *freezer) TruncateAncients(items uint64) error {
...
 	for _, table := range f.tables {
 		if err := table.Sync(); err != nil {
 			errs = append(errs, err)
 		}
 	}
-	if errs != nil {
+	hasError := len(errs) > 0
+	if hasError && retry < 10 {
+		log.Info("sync", "retry", retry, "errors", errs)
+		return f.SyncRetry(retry+1)
```
baptiste-b-pegasys: should we add some (exponential backoff) delay?

antonydenyer: yeah probably not a bad idea - or even just a little :)

baptiste-b-pegasys (approved):

PR #1457, opened 2022-07-20
Feat transitions transaction size limit

antonydenyer:
All transactionSizeLimit to be specified in a transition:

```
"transitions" : [{
  "block": 200,
  "transactionSizeLimit": 128
}]
```

### DIFF
--- a/cmd/geth/config.go
+++ b/cmd/geth/config.go
@@ -368,12 +368,10 @@ func quorumValidateConsensus(ethereum *eth.Ethereum, isRaft bool) {
 func quorumValidatePrivacyEnhancements(ethereum *eth.Ethereum) {
 	privacyEnhancementsBlock := ethereum.BlockChain().Config().PrivacyEnhancementsBlock
 
-	if !privacyEnhancementsBlock {
-		for _, transition := range ethereum.BlockChain().Config().Transitions {
-			if transition.PrivacyPrecompileEnabled {
-				privacyEnhancementsBlock = transition.Block
-				break
-			}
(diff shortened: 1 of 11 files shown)


cmd/geth/config.go
```diff
@@ -362,6 +367,13 @@ func quorumValidateConsensus(ethereum *eth.Ethereum, isRaft bool) {
 // the PrivacyEnhancements feature
 func quorumValidatePrivacyEnhancements(ethereum *eth.Ethereum) {
 	privacyEnhancementsBlock := ethereum.BlockChain().Config().PrivacyEnhancementsBlock
+
+	for _, transition := range ethereum.BlockChain().Config().Transitions {
+		if transition.PrivacyPrecompileEnabled {
+			privacyEnhancementsBlock = transition.Block
+		}
+	}
```
baptiste-b-pegasys: I think we can avoid many ops as it is one way switch off to on
```suggestion
	if !privacyEnhancementsBlock {
		for _, transition := range ethereum.BlockChain().Config().Transitions {
			if transition.PrivacyPrecompileEnabled {
				privacyEnhancementsBlock = transition.Block
				break
			}
		}
	}
```

antonydenyer: In theory it's not one way, you could decide to turn it off. Why you would do this I have node idea 😆

baptiste-b-pegasys: but here, currently in the code you can't disable
not a pointer

antonydenyer: 🤦 Good point!

PR #1466, opened 2022-08-02
Fix acceptance tests after changes for gas price enabled

SatpalSandhu61:
Prior to the changes to enable gas price (PR#1446), Quorum was (incorrectly) rewarding account 0x0 with the static block reward. This behaviour was not carried forward, and if the gas price was not enabled then no reward was given. However, this would cause `BAD BLOCK` errors when upgrading the network and there was a mixture of code versions running. Additionally it was causing the mixed network acceptance tests to fail.

This PR corrects the behaviour and ensures that the reward is still given to account 0x0 if gas price has not been enabled.


antonydenyer (approved):
LGTM!

PR #1578, opened 2022-12-01
perf: improved func ValidateNodeForTxn node id parsing

denis-gudim:
Enhanced permission ValidateNodeForTxn function contains repeating parsing of enode identifier from NodeInfo.Url field. This fix stores parsed enode id value to cached NodeInfo object and reuse stored value on each ValidateNodeForTxn function call.

In our EOA transactions performance test with enabled enhanced permission, this fix decrease CPU usage on ~10% in collected pprof report.

Benchmark source:
```
goos: linux
goarch: amd64
pkg: github.com/ethereum/go-ethereum/permission/core
cpu: Intel(R) Xeon(R) CPU E5-2696 v3 @ 2.30GHz
BenchmarkValidateNodeForTxn-8             100000             17177 ns/op           11911 B/op        181 allocs/op
BenchmarkValidateNodeForTxn-8             100000             16033 ns/op           11914 B/op        181 allocs/op
BenchmarkValidateNodeForTxn-8             100000             16371 ns/op           11886 B/op        181 allocs/op
BenchmarkValidateNodeForTxn-8             100000             16227 ns/op           11915 B/op        181 allocs/op
BenchmarkValidateNodeForTxn-8             100000             15923 ns/op           11870 B/op        181 allocs/op
PASS
ok      github.com/ethereum/go-ethereum/permission/core 8.204s
```

Benchmark fix:
```
goos: linux
goarch: amd64
pkg: github.com/ethereum/go-ethereum/permission/core
cpu: Intel(R) Xeon(R) CPU E5-2696 v3 @ 2.30GHz
BenchmarkValidateNodeForTxn-8             100000              1104 ns/op              40 B/op          3 allocs/op
BenchmarkValidateNodeForTxn-8             100000              1750 ns/op              40 B/op          3 allocs/op
BenchmarkValidateNodeForTxn-8             100000              1584 ns/op              41 B/op          3 allocs/op
BenchmarkValidateNodeForTxn-8             100000              1467 ns/op              41 B/op          3 allocs/op
BenchmarkValidateNodeForTxn-8             100000              1631 ns/op              42 B/op          3 allocs/op
PASS
ok      github.com/ethereum/go-ethereum/permission/core 0.789s
```

### DIFF
--- a/eth/api_backend.go
+++ b/eth/api_backend.go
@@ -344,7 +344,7 @@ func (b *EthAPIBackend) SendTx(ctx context.Context, signedTx *types.Transaction)
 	if !pcore.ValidateNodeForTxn(b.nodeId, signedTx.From()) {
 		return errors.New("cannot send transaction from this node")
 	}
-	// End Quourum
+	// End Quorum
 	return b.eth.txPool.AddLocal(signedTx)
 }
 
--- a/eth/api_backend.go
+++ b/eth/api_backend.go
@@ -40,6 +40,7 @@ import (
 	"github.com/ethereum/go-ethereum/event"
 	"github.com/ethereum/go-ethereum/internal/ethapi"
 	"github.com/ethereum/go-ethereum/miner"
+	"github.com/ethereum/go-ethereum/p2p/enode"
 	"github.com/ethereum/go-ethereum/params"
 	pcore "github.com/ethereum/go-ethereum/permission/core"
 	"github.com/ethereum/go-ethereum/rpc"
@@ -55,8 +56,8 @@ type EthAPIBackend struct {
 
 	// Quorum
 	//
-	// hex node id from node public key
-	hexNodeId string
+	// node id from node public key
+	nodeId enode.ID
 
 	// timeout value for call
 	evmCallTimeOut time.Duration
@@ -337,11 +338,13 @@ func (b *EthAPIBackend) SubscribeLogsEvent(ch chan<- []*types.Log) event.Subscri
 }
 
 func (b *EthAPIBackend) SendTx(ctx context.Context, signedTx *types.Transaction) error {
+	// Quourum
 	// validation for node need to happen here and cannot be done as a part of
 	// validateTx in tx_pool.go as tx_pool validation will happen in every node
-	if b.hexNodeId != "" && !pcore.ValidateNodeForTxn(b.hexNodeId, signedTx.From()) {
+	if !pcore.ValidateNodeForTxn(b.nodeId, signedTx.From()) {
 		return errors.New("cannot send transaction from this node")
 	}
+	// End Quourum
 	return b.eth.txPool.AddLocal(signedTx)
 }
 
--- a/permission/core/cache.go
+++ b/permission/core/cache.go
@@ -132,7 +132,6 @@ type OrgDetailInfo struct {
 }
 
 var syncStarted = false
-var emptyNodeId = enode.ID{}
 var defaultAccess = FullAccess
 var qip714BlockReached = false
 var networkBootUpCompleted = false
@@ -601,7 +600,7 @@ func CheckIfAdminAccount(acctId common.Address) bool {
 
 // validates if the account can transact from the current node
 func ValidateNodeForTxn(nodeId enode.ID, from common.Address) bool {
-	if !PermissionsEnabled() || nodeId == emptyNodeId {
+	if !PermissionsEnabled() || nodeId == (enode.ID{}) {
 		return true
 	}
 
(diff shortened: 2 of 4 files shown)


eth/api_backend.go
```diff
@@ -337,11 +338,13 @@ func (b *EthAPIBackend) SubscribeLogsEvent(ch chan<- []*types.Log) event.Subscri
...
 
 func (b *EthAPIBackend) SendTx(ctx context.Context, signedTx *types.Transaction) error {
+	// Quourum
 	// validation for node need to happen here and cannot be done as a part of
 	// validateTx in tx_pool.go as tx_pool validation will happen in every node
-	if b.hexNodeId != "" && !pcore.ValidateNodeForTxn(b.hexNodeId, signedTx.From()) {
+	if !pcore.ValidateNodeForTxn(b.nodeId, signedTx.From()) {
 		return errors.New("cannot send transaction from this node")
 	}
+	// End Quourum
```
baptiste-b-pegasys:
```suggestion
	// End Quorum
```

permission/core/cache.go
```diff
@@ -91,6 +92,19 @@ type NodeInfo struct {
...
 	Url    string     `json:"url"`
 	Status NodeStatus `json:"status"`
+	id     atomic.Value
+}
+
+func (n *NodeInfo) ID() enode.ID {
+	if id := n.id.Load(); id != nil {
+		return id.(enode.ID)
+	}
+	if node, err := enode.ParseV4(n.Url); err == nil {
```
baptiste-b-pegasys, line 102: here is either a duplicate of https://github.com/ConsenSys/quorum/pull/1578/files#diff-9cc7046826ce92d9cc5432cf41ed4fc14bb6907da44f771e3932302c40e8dbc2R394 or an issue with the error handling.
I am not confident.

permission/core/cache.go
```diff
@@ -118,7 +132,7 @@ type OrgDetailInfo struct {
 }
 
 var syncStarted = false
-
+var emptyNodeId = enode.ID{}
```
baptiste-b-pegasys: I don't like global variables, but it's just me.

permission/core/cache.go
```diff
@@ -579,16 +600,11 @@ func CheckIfAdminAccount(acctId common.Address) bool {
 }
 
 // validates if the account can transact from the current node
-func ValidateNodeForTxn(hexnodeId string, from common.Address) bool {
-	if !PermissionsEnabled() || hexnodeId == "" {
+func ValidateNodeForTxn(nodeId enode.ID, from common.Address) bool {
+	if !PermissionsEnabled() || nodeId == emptyNodeId {
```
baptiste-b-pegasys: so back with what I don't like, maybe an `IsEmpty()` function might work or a `len(nodeId) == 0`?

baptiste-b-pegasys (commented):
LGTM, I raise few small remarks as I don't entirely master the subject with errors handling and go types

permission/core/cache.go
denis-gudim: It's not duplicate, in this block we reuse cached value in atomic field or parse specified node id from node URL in loop. But in https://github.com/ConsenSys/quorum/pull/1578/files#diff-9cc7046826ce92d9cc5432cf41ed4fc14bb6907da44f771e3932302c40e8dbc2R394 block, we parse current node URL and store node id to ETH API Backend object. So it's different nodes.

The main idea of my fix is replaced string node id value to parse binary value

permission/core/cache.go
denis-gudim: Agree, me too. Replaced to empty enode id value in if block. https://github.com/ConsenSys/quorum/pull/1578/commits/39fb51961fa51af64d0768e0b05c681a1a3b6d42#diff-a39234da6bbd13a5c02519dd2dff8aac2ea8a8d948ac58ed60c4ce5bd84f34b5R603

permission/core/cache.go
denis-gudim: From my point of view function is too complex for this action and I didn't want to change enode package logic for this case. 

`len(nodeId) == 0` will not work for this case because enode.ID is fixed size bytes array and `len(enode.ID{})` is 32 bytes

My solution is to compare node id with empty enode.ID value . Please check this fix: https://github.com/ConsenSys/quorum/pull/1578/commits/39fb51961fa51af64d0768e0b05c681a1a3b6d42#diff-a39234da6bbd13a5c02519dd2dff8aac2ea8a8d948ac58ed60c4ce5bd84f34b5R603

baptiste-b-pegasys (approved):

Issue #1582, opened 2022-12-09
Wrong `msg.sender` during pre-transaction with `privacyFlag=1`


frank-lim-partior:
#### System information

Geth version: Geth/32e486ea6b3f-istanbul/v1.10.2-stable-e5b930e8(quorum-v22.7.0)/linux-amd64/go1.16.15
OS & Version: Linux

#### Expected behaviour
When a transaction is being sent with `privacyFlag=1`, `msg.sender` returns a value with the public address of the transaction signer.

#### Actual behaviour
A different public address (observed to be that of the original contract deployer) is being returned and reverts in a contract that verifies the address. The txn reverts on call with error message containing the value of the `msg.sender` (we see a different address than expected), but the txn hash cannot be found in the geth node, indicating that the txn did not hit the blockchain.

#### Steps to reproduce the behaviour
(EDIT: Please see first comment for file with sample node script, README.md, and contracts for replicating this issue)
1. Start quorum test network with some nodes (genesis file attached with enhancedPrivacy enabled)
2. Choose a node to use. Take note of the accounts in the geth signer list.
3. Deploy SimpleSignerCheck.sol (from wrongsigner.zip) into the network with privacyFlag=1. Use the first account as the deploying account. Use the second account's address in the constructor of the contract so we expect the caller to be the 2nd account.
4. Call `testCall(10)` on deployed SimpleSignerCheck with second account, it should revert with the error. The error data will contain 'msg.sender' which reflects the public address of the first account.

#### Backtrace

````
Error: Returned error: execution reverted
...
../node_modules/xhr2-cookies/dist/xml-http-request.js:289:61)
    at IncomingMessage.emit (events.js:412:35)
    at endReadableNT (internal/streams/readable.js:1333:12)
    at processTicksAndRejections (internal/process/task_queues.js:82:21) {
  data: '0xa150e3da000000000000000000000000ca843569e3427144cead5e4d5999a3d0ccf92b8e'
}
````

frank-lim-partior:
[wrongsigner.tar.gz](https://github.com/ConsenSys/quorum/files/10191780/wrongsigner.tar.gz)
Please see attached for a script, test contract, genesis file, and instructions for replicating the issue.

frank-lim-partior:
The script works when all `privacyFlag` in the txns are set to 0.

frank-lim-partior:
Here is the `package.json` used:

`{
  "name": "@consensys-software/smart_contracts",
[...]

PR #1584, opened 2022-12-14
fix: use txn's from instead of hardcoding the address in simulations

frank-lim-partior:
Fix for https://github.com/ConsenSys/quorum/issues/1582.

For some reason, the 'from' address parameter is hardcoded to `common.Address{}` instead of using the signed transaction's `from` value. This causes the wrong `msg.sender` to be picked up during simulations for `privacyFlag=1`.


antonydenyer (approved):

antonydenyer:
Thanks @frank-lim-partior nice catch

Krish1979 (approved):

PR #1609, opened 2023-02-21
all: more linters (#24783)

antonydenyer:
all: more linters (#24783)
This enables the following linters

- typecheck
- unused
- staticcheck
- bidichk
- durationcheck
- exportloopref
- gosec

With a few exceptions.

- We use a deprecated protobuf in trezor. I didn't want to mess with that, since I cannot meaningfully test any changes there.
- The deprecated TypeMux is used in a few places still, so the warning for it is silenced for now.
- Using string type in context.WithValue is apparently wrong, one should use a custom type, to prevent collisions between different places in the hierarchy of callers. That should be fixed at some point, but may require some attention.
- The warnings for using weak random generator are squashed, since we use a lot of random without need for cryptographic guarantees.

Some amendments from https://github.com/ethereum/go-ethereum/commit/a907d7e81aaeea15d80b2d3209ad8e08e3bf49e0


Krish1979 (approved):
LGTM

PR #1666, opened 2023-08-14
feat: Fix generating uuid requiring admin account and json rpc number of parameters

rodion-lim-partior:
There are 2 issues fixed in this PR:

1. Querying through JSON rpc was throwing an error due to invalid number of parameters specified in `web3ext.go`
2. Generating an extension uuid requires admin account, even though approving the contract state extension does not

Validation on admin accounts was implicitly introduced in the previous PR when calling `GenerateTransactOptions` function, which is not required for generating uuid. This is now removed.


rodion-lim-partior:
@antonydenyer, could you kindly help me to take a look at this PR pls? This is a fix to patch some of the remaining issues in the previous PR.

antonydenyer (approved):

PR #1670, opened 2023-08-28
Fix abigen not supporting enhanced privacy for go

rodion-lim-partior:
abigen creates wrapper contract structs that do not have the capability to specify enhanced privacy options. This PR allows users to pass in privacy flag, private from and mandatory for fields to a generated wrapper.


rodion-lim-partior:
@antonydenyer, could you kindly help to take a look at this PR please? Existing abigen only supports `privateFor` on standard privacy.

PR #1677, opened 2023-09-19
Add function to toggle on/off IP Validation in Enhanced Permissioing NodeManager.

john-sobrepena-partior:
**Use case** 
We are running a 1st Quorum node in kubernetes pod with Enhanced Permissioning v2 enabled. A 2nd quorum node is hosted in another kubernetes cluster / namespace also running Enhanced Permissioning v2. The 1st and 2nd quorum node's enodeIds, ports and IP addresses were added to the list of nodes that are allowed to connect. The **geth** API `quorumPermission_addNode()` was used.
However, the 1st and 2nd node are not able to connect to each other because, for example - the incoming connection from 2nd node's IP has changed. Based on discussion with our infra team, and due to securities in place, the IP address of incoming connection may change or re-written before the connection request finally reach the pod where the other Quorum node is running. Due to this, the  `connectionAllowed()` of `v2/contract/NodeManager.sol` rejects the connection.

**Proposed Solution / Changes**
I am proposing in this change to introduce a `setIpValidation()` function in NodeManager.sol to toggle IP validation on or off. This is enabled by default. The property can only be set before network boot status is finalized.

For your review / comments. Let me know if you require further information. I will be more than happy to provide them and to help contribute further. I look forward working with you.

Thank you in advance.

Regards, John

### DIFF
--- a/permission/v2/contract/NodeManager.sol
+++ b/permission/v2/contract/NodeManager.sol
@@ -38,7 +38,8 @@ contract NodeManager {
     mapping(bytes32 => uint256) private enodeIdToIndex;
     // tracking total number of nodes in network
     uint256 private numberOfNodes;
-
+    // whether to do IP validation during checks if connection is allowed. This is enabled by default.
+    bool private isIpValidationEnabled = true;
 
     // node permission events for new node propose
     event NodeProposed(string _enodeId, string _ip, uint16 _port, uint16 _raftport, string _orgId);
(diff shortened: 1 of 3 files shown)


john-sobrepena-partior:
Hi @antonydenyer, kindly help look into and review my proposed changes for Enhanced Permissioning v2. Let me know any further clarifications required. Thank you in advance. 

Regards, John

Issue #1660, opened 2023-07-21
Memory usage of non-validator nodes grows indefinitely, leading to OOM and unclean shutdown


hhsel:
#### Expected behaviour
When running a QBFT cluster, memory usage should stay within a moderate value range as long as the cluster is not busy.

#### Actual behaviour
Memory usage of QBFT non-validator nodes grows over time at a rate of approx. 50MB/day, if the cluster keeps producing empty blocks every 1 second, for example.
Non-validator nodes will be killed by OOM as a result. I have experienced this with 2GB and 4GB nodes, and it took about 1 and 2 months for the nodes to be killed by OOM.

OOM causes an unclean shutdown, which means the node loses its intermediate states that are not persisted to its disk.
The memory usage grows indefinitely, even the cluster is producing just empty blocks and does almost nothing on the chain.
In my case 8 out of 8 non-validator nodes in the cluster have the same results.

Validator nodes, on the other hand, have similar tendencies but several sudden memory usage drops have been observed (its frequency is not regular nor expectable but about once in 1-2 weeks).

As a result, for non-validator nodes, I must watch its memory usage closely and take nodes out of a load balancer and restart them when memory usage gets high, to avoid OOM.

#### Steps to reproduce the behaviour
Start a QBFT cluster with an arbitrary number of non-validator nodes, and let the cluster produce empty blocks.
Memory usage of non-validator nodes grows indefinitely, causing OOM after some months.

hhsel:
Executing `debug.writeMemProfile()` every 15 minutes for 2 months showed that, among `go tool pprof` entries, the following entries are growing constantly, finally causing OOM:

- `github.com/ethereum/go-ethereum/consensus/istanbul/validator.newDefaultSet`
- `github.com/ethereum/go-ethereum/consensus/istanbul/validator.New (inline)`

`validator.newDefaultSet` calls `policy.RegisterValidatorSet()` and add a validator set to validator set list called `registry` inside `ProposerPolicy`. It seems that this function is called at least once in the block production process.
It seems that the only way to clear this array is calling `ClearRegistry()`, but this is only called in `Commit()`, so non-validator nodes does not call this function, accumulating validator sets indefinitely.

I don't know why validator nodes also accumulating these validator sets in the same manner and sudden memory usage drops occur, because validator nodes call `Commit()` every block so `registry` is expected to keep relarively small.

naoki4:
Hello
[...]

PR #1687, opened 2023-12-11
fix: memory leak from policy registration

rodion-lim-partior:
Start a QBFT/IBFT cluster with an arbitrary number of non-validator nodes. Memory usage of non-validator nodes grows indefinitely, causing OOM eventually (Reference issue #1660). Issue can easily be reproduced by logging out the length of validator sets within the policy's registry at runtime.

Non validator nodes calls `RegisterValidatorSet` method for every single block, adding on to the in memory ProposerPolicy.registry. GC is unable to free the memory since ClearRegistry is never called. There are cases when a restart of the quorum node causes even validator nodes to indefinitely add new validator sets to the registry.

Registry is currently not used in any of the other implementation code, this PR sets a cap to the maximum number of validator sets allowed in memory at any one point in time (in the event when registry is planned for future use cases). It caps the registry up to the latest 128 blocks, aligned to the number of in-memory state transitions. Older registrations are sequentially booted out of the registry.

### DIFF
--- a/consensus/istanbul/config.go
+++ b/consensus/istanbul/config.go
@@ -24,6 +24,7 @@ import (
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/math"
+	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/params"
 	"github.com/naoina/toml"
 )
@@ -34,6 +35,8 @@ const (
 	RoundRobin ProposerPolicyId = iota
 	Sticky
 )
+
+const MaxValidatorSetInRegistry = 128 // Max number of ValidatorSet in the registry
 
 // ProposerPolicy represents the Validator Proposer Policy
 type ProposerPolicy struct {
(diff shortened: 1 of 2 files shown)


frank-lim-partior:
@YoshihitoAso, this is possibly related to the issue you raised in https://github.com/BoostryJP/quorum/issues/56

cheeweng-tan-partior:
After implementing the fix, we conducted a thorough profiling analysis on the non-validator nodes. The observed memory behavior now indicates a significant improvement, with memory consumption stabilizing instead of exhibiting indefinite growth, as previously observed. 

Please refer to the attached memory consumption data for the past 3 days.
image

We used pprof snapshots to compare the memory usage before and after the fix. Here are the key observation:
- Before the Fix: The pprof snapshot distinctly reveals that the predominant portion of memory consumption stemmed from 'validator.newDefaultSet'.
image

- After the Fix: Upon re-profiling, 'validator.newDefaultSet' no longer appears in the top 20 memory consumers. Instead, only 'leveldb' remains as one of the main top 5 memory consumers, which is expected during block minting.
image

**In summary:**
The fix has effectively addressed memory leakage that was observed, notably, 'validator.newDefaultSet' no longer contributes to excessive memory consumption post-fix, aligning with expected behavior. These observations highlight the efficacy of the fix to show a positive results in mitigating memory leakage and enhancing overall system stability / reliability.

purplesmoke05 (approved):

PR #1688, opened 2023-12-13
Logging

ashp7:
This PR does the following : 
1) Add extra logging to the IBFT workflow through the following phases 
  i) proposal ii)  pre-prepare iii) prepare iv) commit v) round change

2) add a new contract file NFTCollateral.sol which handles the following scenario
Preconditions:
○ A borrower owns a number of NFTs
○ A lender owns certain amount of fungible tokens
● Financing life cycle:
○ The borrower submits a proposal to put up one or more of his NFTs as collateral
to borrow certain amount of “money”, and proposed an interest rate and payment
terms (in how many days will the debt be paid off)
○ At this point the NFTs promised in the proposal are deposited and locked in a
contract
  
○ A lender sees the proposal on the blockchain, reviews the terms and picks up the proposal by submitting a transaction which transfers the borrowed amount to the borrower
○ When the borrower pays off the debt plus interest to the lender, the NFTs are released by the contract and transferred back to the borrower
○ If the borrower fails to pay off before the deadline that was set in the proposal, the NFTs are released by the contract and transferred to the lender
○ If no lenders are interested in the proposal, the borrower can retract it and get back the NFTs from the contract

### DIFF
--- a/consensus/istanbul/backend/backend.go
+++ b/consensus/istanbul/backend/backend.go
@@ -225,6 +225,7 @@ func (sb *Backend) Commit(proposal istanbul.Proposal, seals [][]byte, round *big
 	// Remove ValidatorSet added to ProposerPolicy registry, if not done, the registry keeps increasing size with each block height
 	sb.config.ProposerPolicy.ClearRegistry()
 
+	sb.logger.Info("Update block's header")
 	// update block's header
 	block = block.WithSeal(h)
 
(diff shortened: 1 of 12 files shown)


SatpalSandhu61:
A lot of that logging appears to be debug related and should really be 'trace' level.
Also, the NFTContract is dapp-specific and doesn't belong in Quorum (especially since it adds a dependency on openzeppelin).

ashp7:
Sorry. This was intended to be private code to try out some stuff. will close the PR

PR #1734, opened 2024-09-27
fix: fix slice init length

cuishuang:
The intention here should be to initialize a slice with a capacity of types.IstanbulExtraVanity rather than initializing the length of this slice.


cuishuang:
@Krish1979 Hi, could you please review this PR? Thank you!

